PR #144, opened 2017-06-06
add decision state machine

yiminc:
Verified that: 1) all unit test passes; 2) all integration tests passes; 3) all samples worked.
Here is the structure of the code:
1) define decisionStateMachine interface that contains methods to drive the state machine. The methods could easily be mapped to the history events. State machine also knows when to generate decisions based on its current state.
2) implement a base class decisionStateMachineBase that defines the common logic of states transition.
3) implement classes that embed decisionStateMachineBase and added specific logic for each entity (activity/timer/childworkflow/marker etc.)
4) added a helper class decisionsHelper, that has methods defined directly for each history event. It is also used to keep track of all decisions.
5) updated workflowEnvironmentImpl to make use of the decisionsHelper. When decider code call into workflowEnvironmentImpl's methods (ExecuteActivity/ExecuteChildworkflow/StartTimer etc), it uses the decisionsHelper to put new decision state machine into its internal map.
6) update workflowExecutionEventHandlerImpl to make use of the decisionsHelper. When it process history events, the event handler will use the decisionsHelper to drive the states of the decision state machines to make state transitions.

Some things worth mention:
1) workflowEnvironmentImpl has maps of scheduled activities/timers/childworkflows. Previously, we would delete entity from those maps once they are handled. Now we won't delete them, instead we mark them as handled. The reason is so we could always get the entity with a valid key (activityID/timerID/childWorkflowID), and if we fail to get the entity, then we know we have an invalid key which indicate bugs so we could just panic. Previously, we also keep a map  waitForCancelRequestActivities for the waitForCancelRequest flag, and we would need another one for the child workflow. Now, as we don't delete the entity, those flag are part of the entity.

2) In the state machine's state transition, if there is illegal move, the state machine would panic instead of return error. This make the code much easier because normal code path don't have to check errors everywhere, and it should just work. If we use the error pattern, we would need to check error everywhere, and if some code forget to check the returned error, the state transition would silently fail, and would cause trouble down the road. So panic is a better choice in my opinion. The should only happen when there is either ill history events sequences due to server bug or it is due to decider's code been non-deterministic.

3) for cancel external workflow decision, it would be child workflow or it could be non-child external workflow. For child workflow, I want this to be part of the child workflow state machine's life cycle. This is how swf handled it. For external non-child workflow, I would use a separate state machine to model it. For non-child workflow, the state is simple, it can only go one way from Created->Sent->Initiated->Completed. For child workflow, the states is more sophisticated.

### DIFF
--- a/internal_decision_state_machine.go
+++ b/internal_decision_state_machine.go
@@ -1,0 +1,820 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cadence
+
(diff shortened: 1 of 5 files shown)


yiminc:
ok, i cannot update this PR since the repo changed to public and I have to refork. Will land this as is and send a small diff to address the comments.

internal_decision_state_machine.go
```diff
@@ -0,0 +1,820 @@
...
+		handleDecisionSent()
+	}
+
+	decisionStateMachineBase struct {
+		id      decisionID
+		state   decisionState
+		history []string
+	}
+
+	activityDecisionStateMachine struct {
```
venkat1109: Just adding to maxim's comment, a table driven state machine implementation could also make the code totally self-documenting here. With a table driven approach, you create a state table for every kind of decisionStateMachine (i.e. activity/timer/childWF etc).. The table itself will look like the following:

```
stateTable struct {
      [NUM_EVENTS] eventToNextState
} [NUM_STATES]

var activityStateTable = [NUM_STATES] {
   { },
   { },
   ...
}

```

i.e. you declare every possible state transition in a static table and the machine just follows the table.

mfateev: It is actually a cool idea. The functions that generate decision based on state can be part of the table as well. It is also self documenting.

PR #148, opened 2017-06-09
add child workflow support to test suite

yiminc:
Adding child workflow to test suite.

Known issue: need to rethink about how we auto forward the mock clock. The clock probably does not work in some cases with child workflow. I plan to address that in a separate diff.

### DIFF
--- a/internal_workflow_testsuite.go
+++ b/internal_workflow_testsuite.go
@@ -240,6 +240,9 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(optio
 	childEnv.workflowInfo.ExecutionStartToCloseTimeoutSeconds = *options.executionStartToCloseTimeoutSeconds
 	childEnv.workflowInfo.TaskStartToCloseTimeoutSeconds = *options.taskStartToCloseTimeoutSeconds
 
+	workflowID := childEnv.workflowInfo.WorkflowExecution.ID
+	env.childWorkflows[workflowID] = &testChildWorkflowHandle{env: childEnv}
+
 	return childEnv
 }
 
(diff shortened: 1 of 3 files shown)


internal_workflow_testsuite.go
```diff
@@ -296,6 +364,26 @@ func (env *testWorkflowEnvironmentImpl) overrideActivity(activityFn, fakeActivit
 	env.overrodeActivities[fnName] = fakeActivityFn
 }
 
+func (env *testWorkflowEnvironmentImpl) overrideWorkflow(workflowFn, fakeWorkflowFn interface{}) {
+	// verify both functions are valid workflow func
```
venkat1109: this is the same code for overrideActivity() - consider refactoring the validation into its own helper - something like validateFunctionOverride(actual, fake) string

internal_workflow_testsuite.go
```diff
@@ -827,13 +936,56 @@ func (env *testWorkflowEnvironmentImpl) RegisterSignalHandler(handler func(name
...
+		// check if current workflow is a child workflow
+		parentEnv := env.parentEnv
+		if parentEnv != nil && parentEnv.onChildWorkflowCanceledListener != nil {
+			parentEnv.postCallback(func() {
+				parentEnv.onChildWorkflowCanceledListener(env.workflowInfo)
+			}, false)
+		}
+	} else if childHandle, ok := env.childWorkflows[workflowID]; ok {
+		// current workflow is a parent workflow, and we are canceling a child workflow
+		delete(env.childWorkflows, workflowID)
```
venkat1109: shouldn't this be deleted inside the if{} instead of the else{} ?

internal_workflow_testsuite.go
```diff
@@ -186,6 +211,38 @@ func newTestWorkflowEnvironmentImpl(s *WorkflowTestSuite) *testWorkflowEnvironme
 	return env
 }
 
+func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(options *workflowOptions) *testWorkflowEnvironmentImpl {
+	// create a new test env
+	childEnv := newTestWorkflowEnvironmentImpl(env.testSuite)
+	childEnv.parentEnv = env
```
venkat1109: add this workflow to the parent's map[childWorkflows]

internal_workflow_testsuite.go
yiminc: no, it should be in the else case, as we are deleting child from parent, so we do it when current workflow is parent workflow.

internal_workflow_testsuite.go
yiminc: will update

internal_workflow_testsuite.go
yiminc: child was added in ExecuteChildWorkflow() method. Put it here also works, will update.

PR #156, opened 2017-06-12
fix noisy logging for non-deterministic unit test

yiminc:
use no-op logger for this unit test. also update the unit test to cover both cases of matched and non-matched case.


internal_task_handlers_test.go
```diff
@@ -177,25 +177,31 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_ActivityTaskScheduled() {
...
 	response, _, err := taskHandler.ProcessWorkflowTask(task, false)
 
+	// there should be no error as the history events matched the decisions.
+	t.NoError(err)
+	t.NotNil(response)
+
+	// now change the history event so it does not match to decision produced via replay
+	testEvents[1].ActivityTaskScheduledEventAttributes.ActivityType.Name = common.StringPtr("some-other-activity")
+	response, _, err = taskHandler.ProcessWorkflowTask(task, false)
 	t.NotNil(err)
```
madhuravi: nit: you can use t.Error for this instead of NotNil.

PR #158, opened 2017-06-14
Workflow panic to fail decision task

sivakku:
- undo activity panic .
- panic to timeout decision task.


internal_task_handlers.go
```diff
@@ -432,6 +432,13 @@ ProcessEvents:
 		return nil, "", err
 	}
 
+	if _, ok := failure.(PanicError); ok {
+		// Timeout the Decision instead of failing workflow.
+		// TODO: Pump this stack trace on to workflow history for debuggability by exposing decision type fail to client.
+		wth.logger.Error("Workflow failed with a panic error.", zap.Error(failure))
+		return nil, "", failure
```
mfateev: Wouldn't returning the failure cause it to be logged?

sivakku: yes, removed this log.

PR #161, opened 2017-06-14
Start workflow Options rename

sivakku:
- rename workflow options timeout to time.duration
- add comments about options.
- On worker rename group name to task list.


internal_workflow_client.go
```diff
@@ -70,6 +74,23 @@ func (wc *workflowClient) StartWorkflow(
...
+	executionTimeout := int32(options.ExecutionStartToCloseTimeout.Seconds())
+	if executionTimeout <= 0 {
+		return nil, errors.New("missing or invalid ExecutionStartToCloseTimeout")
+	}
+
+	decisionTaskTimeout := int32(options.DecisionTaskStartToCloseTimeout.Seconds())
+	if decisionTaskTimeout < 0 {
+		return nil, errors.New("negative DecisionTaskStartToCloseTimeout provided")
+	}
+	if decisionTaskTimeout == 0 {
```
yiminc: use <= 0

sivakku: less than is one level above that throws error.

yiminc: 👍

PR #162, opened 2017-06-15
remove our private modified mock copy

yiminc:


internal_workflow_testsuite.go
```diff
@@ -756,12 +756,15 @@ func (m *mockWrapper) getMockReturn(ctx interface{}, input []byte) mock.Argument
...
-	found, _ := m.mock.FindExpectedCall(m.name, realArgs...)
 
-	if found < 0 {
-		// mock call not found
-		return nil
-	}
+	// There is no way to check if a mock call is expected or not. A pull request to add it was rejected. See PR:
+	// https://github.com/stretchr/testify/pull/453. We could try to call the mock method, and if the call is not
+	// expected, the mock.MethodCalled() will panic, which we would catch and just return nil from this method.
+	defer func() {
```
mfateev: Are any other situations that cause panic? My concern that some other panic is swallowed and troubleshooting is practically impossible.

yiminc: The only other possible is that user would supply some function to be run before mock execute, and that user function could panic because of some bug. This is a valid concern. I will address that in my next diff which I will need to wrap the mock call to add the support for delay mock call on workflow clock.

PR #164, opened 2017-06-16
allow mock to wait on workflow clock before it returns

yiminc:
We originally thought that we could make use of the mock.Call.WaitUntil() method. And what we need was just to define a new method on our test environment to return a channel that wait on our internal workflow's mock clock. That approach doesn't work. Here is the reason: when the mock is waiting on the workflow clock, we need to mark the activity as not running, otherwise we won't be able to move forward clock and the mock will wait forever. The mock could be blocked within mock.MethodCalled(), but we don't know if it will be blocked or not before we call it because we don't have insight into the mock call object. There is no place where we could detect that mock call is about to be blocked and we need to reduce the runningCount.

In order to make this work, we need a hook method that will be called before mock call is waiting on our workflow clock. This is achieved by setup a run function for the mock call via *mock.Call.Run(fn), and define our MockCallWrapper struct and intercept the .After()  and .Run() method on the mock call object.

### DIFF
--- a/internal_workflow_testsuite.go
+++ b/internal_workflow_testsuite.go
@@ -63,7 +63,8 @@ type (
 	}
 
 	testChildWorkflowHandle struct {
-		env *testWorkflowEnvironmentImpl
+		env      *testWorkflowEnvironmentImpl
+		callback resultHandler
 	}
 
 	testCallbackHandle struct {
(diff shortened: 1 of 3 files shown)


internal_workflow_testsuite.go
```diff
@@ -566,6 +570,21 @@ func (env *testWorkflowEnvironmentImpl) Complete(result []byte, err error) {
 	}
 
 	close(env.doneChannel)
+
+	if env.isChildWorkflow() {
+		// this is completion of child workflow
+		childWorkflowID := env.workflowInfo.WorkflowExecution.ID
+		if childWorkflowHandle, ok := env.childWorkflows[childWorkflowID]; ok {
```
yiminc: In the case of child workflow completed after cancellation. The childWorkflowHandle would already have been removed from the map when it was request to cancel.

PR #167, opened 2017-06-21
Decouple HeartBeat invoke from client to server.

sivakku:
- Optimize frequent call if it is with in certain time window(60% of timeout configured).

### DIFF
--- a/internal_task_handlers.go
+++ b/internal_task_handlers.go
@@ -780,16 +780,16 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 			}
 
 			// We close the batch and report the progress.
-			var progressToReport *[]byte
+			var detailsToReport *[]byte
 
 			i.Lock()
-			progressToReport = i.lastDetailsToReport
+			detailsToReport = i.lastDetailsToReport
 			i.hbBatchEndTimer.Stop()
 			i.hbBatchEndTimer = nil
 			i.Unlock()
 
-			if progressToReport != nil {
-				i.Heartbeat(*progressToReport)
+			if detailsToReport != nil {
+				i.Heartbeat(*detailsToReport)
 			}
 		}()
 	}
--- a/internal_task_handlers.go
+++ b/internal_task_handlers.go
@@ -735,10 +735,9 @@ type cadenceInvoker struct {
 	taskToken             []byte
 	cancelHandler         func()
 	retryPolicy           backoff.RetryPolicy
-	heartBeatTimeoutInSec int32 // The heart beat interval configured for this activity.
-	hbBatchEndTimer       *time.Timer
-	lastProgressToReport  *[]byte
-	isInBatchingWindow    bool // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
+	heartBeatTimeoutInSec int32       // The heart beat interval configured for this activity.
+	hbBatchEndTimer       *time.Timer // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
+	lastDetailsToReport   *[]byte
 	closeCh               chan struct{}
 }
 
@@ -746,9 +745,9 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 	i.Lock()
 	defer i.Unlock()
 
-	if i.isInBatchingWindow {
+	if i.hbBatchEndTimer != nil {
 		// If we have started batching window, keep track of last reported progress.
-		i.lastProgressToReport = &details
+		i.lastDetailsToReport = &details
 		return nil
 	}
 
@@ -758,13 +757,9 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 	// and complete. Our cancellation is co-operative, so we will try to heartbeat.
 	if err == nil || isActivityCancelled {
 		// We have successfully sent heartbeat, start next batching window.
-		i.lastProgressToReport = nil
-		i.isInBatchingWindow = true
+		i.lastDetailsToReport = nil
 
-		// reset timer to fire before the threshold to report.
-		if i.hbBatchEndTimer != nil {
-			i.hbBatchEndTimer.Stop()
-		}
+		// Create timer to fire before the threshold to report.
 		deadlineToTrigger := i.heartBeatTimeoutInSec
 		if deadlineToTrigger <= 0 {
 			// If we don't have any heartbeat timeout configured.
@@ -786,9 +781,11 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 
 			// We close the batch and report the progress.
 			var progressToReport *[]byte
+
 			i.Lock()
-			i.isInBatchingWindow = false
-			progressToReport = i.lastProgressToReport
+			progressToReport = i.lastDetailsToReport
+			i.hbBatchEndTimer.Stop()
+			i.hbBatchEndTimer = nil
 			i.Unlock()
 
 			if progressToReport != nil {
@@ -806,7 +803,6 @@ func (i *cadenceInvoker) internalHeartBeat(details []byte) (bool, error) {
 
 	switch err.(type) {
 	case *CanceledError:
-		// We are asked to cancel. inform the activity about cancellation through context.
 		// We are asked to cancel. inform the activity about cancellation through context.
 		i.cancelHandler()
 		isActivityCancelled = true
--- a/internal_task_handlers.go
+++ b/internal_task_handlers.go
@@ -752,9 +752,11 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 		return nil
 	}
 
-	nonTransientError, err := i.internalHeartBeat(details)
+	isActivityCancelled, err := i.internalHeartBeat(details)
 
-	if err == nil || nonTransientError {
+	// If the activity is cancelled, the activity can ignore the cancellation and do its work
+	// and complete. Our cancellation is co-operative, so we will try to heartbeat.
+	if err == nil || isActivityCancelled {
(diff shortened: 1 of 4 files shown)


internal_task_handlers.go
```diff
@@ -724,45 +728,82 @@ func newActivityTaskHandler(activities []activity,
...
+		}
+		return false
+	}
+
+	// If we have heartbeat timeout configured.
+	// The heartbeat time indicates that on any two successive pings the distance should not be more than the
+	// the configured timeout.
+	// We want to suppress reporting HB too frequently to server, until we get auto heartbeating.
+	// we will suppress reporting HB if it is with-in a certain window(60% of HB timeout)
+	duration := time.Duration(0.6*float32(i.heartBeatTimeoutInSec)) * time.Second
```
yiminc: this won't work. assume we have activity with heartbeat timeout configured to 100s, and the activity starts and immediately heartbeat, lets say it heartbeat at time 0, then at time 59s, it heart beat again, by the proposed rule, this heartbeat will be suppressed, then the activity continues its work, and at time 150s, it heartbeat again, which is supposed to be 91s after its previous heartbeat and should not timeout, but since we suppressed the previous one, it will timeout.

sivakku: After todays discussion, i moved to another approach of inline batching and expiring at the end. updating CR.

internal_task_handlers.go
```diff
@@ -724,45 +729,123 @@ func newActivityTaskHandler(activities []activity,
...
+
+	if i.isInBatchingWindow {
+		// If we have started batching window, keep track of last reported progress.
+		i.lastProgressToReport = &details
+		return nil
+	}
+
+	nonTransientError, err := i.internalHeartBeat(details)
+
+	if err == nil || nonTransientError {
```
molteanu: I'm not sure that "transient" is the right word to use here. "Transient" error to me means that it is a temporary error that should go away on its own. So "nonTransientError" would mean it is an error that's not going to recover on its own. Why would we want to start a new "heartbeat window" when we have an error that will not recover? Unless "transient" was meant to mean something else.

internal_task_handlers.go
```diff
@@ -724,45 +729,123 @@ func newActivityTaskHandler(activities []activity,
...
+	i.Lock()
+	defer i.Unlock()
+
+	if i.isInBatchingWindow {
+		// If we have started batching window, keep track of last reported progress.
+		i.lastProgressToReport = &details
+		return nil
+	}
+
+	nonTransientError, err := i.internalHeartBeat(details)
```
molteanu: Can this call happen on the background go routine? Can we simplify this method to just start the background go routine?

internal_task_handlers.go
```diff
@@ -724,45 +729,123 @@ func newActivityTaskHandler(activities []activity,
...
+			}
+
+			// We close the batch and report the progress.
+			var progressToReport *[]byte
+			i.Lock()
+			i.isInBatchingWindow = false
+			progressToReport = i.lastProgressToReport
+			i.Unlock()
+
+			if progressToReport != nil {
```
sivakku: This is a pointer to the encoded progress, which can point to result which can be nil if user provided nil. Added a unit test as well.

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
-	retryPolicy   backoff.RetryPolicy
+	sync.Mutex
+	identity              string
+	service               m.TChanWorkflowService
+	taskToken             []byte
+	cancelHandler         func()
+	retryPolicy           backoff.RetryPolicy
+	heartBeatTimeoutInSec int32 // The heart beat interval configured for this activity.
+	hbBatchEndTimer       *time.Timer
+	lastProgressToReport  *[]byte
```
yiminc: rename to lastDetailsToReport to keep consistent with the API's parameter

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
-	taskToken     []byte
-	cancelHandler func()
-	retryPolicy   backoff.RetryPolicy
+	sync.Mutex
+	identity              string
+	service               m.TChanWorkflowService
+	taskToken             []byte
+	cancelHandler         func()
+	retryPolicy           backoff.RetryPolicy
+	heartBeatTimeoutInSec int32 // The heart beat interval configured for this activity.
```
yiminc: consider using time.Duration?

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
+	hbBatchEndTimer       *time.Timer
+	lastProgressToReport  *[]byte
+	isInBatchingWindow    bool // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
```
yiminc: can we infer this by checking hbBatchEndTimer != nil?

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
+}
+
+func (i *cadenceInvoker) internalHeartBeat(details []byte) (bool, error) {
+	isActivityCancelled := false
 	err := recordActivityHeartbeat(i.service, i.identity, i.taskToken, details, i.retryPolicy)
 
 	switch err.(type) {
 	case *CanceledError:
 		// We are asked to cancel. inform the activity about cancellation through context.
 		// We are asked to cancel. inform the activity about cancellation through context.
```
yiminc: duplicate comments

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
+
+	// If the activity is cancelled, the activity can ignore the cancellation and do its work
+	// and complete. Our cancellation is co-operative, so we will try to heartbeat.
+	if err == nil || isActivityCancelled {
+		// We have successfully sent heartbeat, start next batching window.
+		i.lastProgressToReport = nil
+		i.isInBatchingWindow = true
+
+		// reset timer to fire before the threshold to report.
+		if i.hbBatchEndTimer != nil {
```
yiminc: this should never be true. We should stop the timer and set hbBatchEndTimer to nil for previous batch when that batch is done.

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
+				// We are close to deadline.
+			case <-i.closeCh:
+				// We got closed.
+				return
+			}
+
+			// We close the batch and report the progress.
+			var progressToReport *[]byte
+			i.Lock()
+			i.isInBatchingWindow = false
```
yiminc: set hbBatchEndTimer to nil here

internal_task_handlers.go
```diff
@@ -724,45 +729,125 @@ func newActivityTaskHandler(activities []activity,
...
+	lastProgressToReport  *[]byte
+	isInBatchingWindow    bool // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
+	closeCh               chan struct{}
 }
 
 func (i *cadenceInvoker) Heartbeat(details []byte) error {
+	i.Lock()
+	defer i.Unlock()
+
+	if i.isInBatchingWindow {
```
yiminc: what happen if we are already closed? should we check for i.closeCh here?

internal_task_handlers.go
yiminc: have same question as Mihnea. If user provide no details, the progressToReport will be nil, and you are skipping the heartbeat here.

internal_task_handlers.go
```diff
@@ -724,45 +729,121 @@ func newActivityTaskHandler(activities []activity,
...
+			select {
+			case <-i.hbBatchEndTimer.C:
+				// We are close to deadline.
+			case <-i.closeCh:
+				// We got closed.
+				return
+			}
+
+			// We close the batch and report the progress.
+			var progressToReport *[]byte
```
yiminc: rename to detailsToReport

internal_task_handlers.go
```diff
@@ -724,45 +729,121 @@ func newActivityTaskHandler(activities []activity,
...
+
+			i.Lock()
+			progressToReport = i.lastDetailsToReport
+			i.hbBatchEndTimer.Stop()
+			i.hbBatchEndTimer = nil
+			i.Unlock()
+
+			if progressToReport != nil {
+				i.Heartbeat(*progressToReport)
```
yiminc: after this heartbeat, should we start a new timer? Otherwise, the next heartbeat call will report immediately.

internal_task_handlers.go
```diff
@@ -724,45 +729,121 @@ func newActivityTaskHandler(activities []activity,
...
+	return isActivityCancelled, err
+}
+
+func (i *cadenceInvoker) Close() {
+	i.Lock()
+	defer i.Unlock()
+
+	close(i.closeCh)
+	if i.hbBatchEndTimer != nil {
+		i.hbBatchEndTimer.Stop()
```
yiminc: set the timer to nil

internal_task_handlers.go
sivakku: updated.

internal_task_handlers.go
sivakku: We don't have to.  one of them can happen, the user call HeratBeat or this one. Since we don't make promise on what progress we report.

internal_task_handlers.go
sivakku: updated.

sivakku: didn't add re-setting this to nil, i ran in to a race, where the background GO select is waiting on this and that throws up.

PR #170, opened 2017-06-23
Added cadence.GetVersion

mfateev:
GetVersion is used to safely perform backwards incompatible changes to workflow definitions.
It is not allowed to update workflow code while there are workflows running as it is going to break
determinism. The solution is to have both old code that is used to replay existing workflows
as well as the new one that is used when it is executed for the first time.
GetVersion returns maxSupported version when is executed for the first time. This version is recorded into the
workflow history as a marker event. Even if maxSupported version is changed the version that was recorded is
returned on replay. DefaultVersion constant contains version of code that wasn't versioned before.

Currently there is no supported way to completely remove GetVersion call after it was introduced.


internal_task_handlers.go
```diff
@@ -472,17 +472,58 @@ ProcessEvents:
...
+	hi := 0
+	hSize := len(historyEvents)
+	dSize := len(replayDecisions)
+matchLoop:
+	for hi < hSize || di < dSize {
+		var e *s.HistoryEvent
+		if hi < hSize {
+			e = historyEvents[hi]
+		}
+		if isVersionMarkerEvent(e) {
```
yiminc: not fully follow the intention of the change here, but looks like you are trying to skip comparing versionMarkerEvent and versionMarkerDecision. If that is the case, I think it is clearer to avoid adding these 2 types into the replayDecisions and historyEvents at first place, then you don't need any change to matchReplayWithHistory.

yiminc: Also, why do we need to skip comparing version marker decision and event during replay?

internal_task_handlers.go
```diff
@@ -472,17 +472,58 @@ ProcessEvents:
...
+			hi++
+			continue matchLoop
+		}
+
+		var d *s.Decision
+		if di < dSize {
+			d = replayDecisions[di]
+		}
+		if isVersionMarkerDecision(d) {
```
yiminc: if di == dSize, then d will be nil, isVersionMarkerDecision will panic?

internal_event_handlers.go
```diff
@@ -633,23 +667,37 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
...
+		weh.sideEffectResult[sideEffectID] = result
+		return nil
+	case versionMarkerName:
+		dec := gob.NewDecoder(bytes.NewBuffer(attributes.GetDetails()))
+		var changeID string
+
+		if err := dec.Decode(&changeID); err != nil {
+			return fmt.Errorf("failure decoding version changeID: %v", err)
+		}
+		var version int
```
yiminc: can you use var version Version, then you don't need to cast type after extracted.

internal_workflow.go
```diff
@@ -189,12 +189,18 @@ type (
 
 		Set(value interface{}, err error)
 	}
+
+	changeVersion struct {
+		minSupported Version
+		version      *Version
```
yiminc: why use pointer for version but not for minSupported?

internal_workflow_testsuite.go
```diff
@@ -1124,5 +1124,10 @@ func (env *testWorkflowEnvironmentImpl) signalWorkflow(name string, input interf
 	}, true)
 }
 
+func (env *testWorkflowEnvironmentImpl) GetVersion(changeID string, minSupported, maxSupported Version) Version {
+	// TODO: Add code that returns the same version for all calls for the given changeID.
+	return maxSupported
```
yiminc: always return maxSupported should be fine unless you expect workflow code to have multiple calls to GetVersion with same changeID but different minSupported/maxSupported. Is that even legit? What does it mean if user do that?

internal_workflow_testsuite_test.go
```diff
@@ -461,6 +461,61 @@ func (s *WorkflowTestSuiteUnitTest) Test_SideEffect() {
...
+	RegisterActivity(oldActivity)
+	RegisterActivity(newActivity)
+	var name string
+	ctx := context.WithValue(context.Background(), nameKey, &name)
+	env.SetWorkerOption(WorkerOptions{BackgroundActivityContext: ctx})
+	env.ExecuteWorkflow(workflowFn)
+
+	s.True(env.IsWorkflowCompleted())
+	s.Nil(env.GetWorkflowError())
+	s.EqualValues("newActivity", name)
```
yiminc: instead of using this custom solution to verify that newActivity is called, you should use the env.SetOnActivityStartedListener() to verify the called activity.

internal_workflow_testsuite_test.go
```diff
@@ -461,6 +461,61 @@ func (s *WorkflowTestSuiteUnitTest) Test_SideEffect() {
...
+		} else {
+			f = ExecuteActivity(ctx, newActivity, "anotherMsg")
+		}
+		err := f.Get(ctx, nil) // wait for result
+		return err
+	}
+
+	env := s.NewTestWorkflowEnvironment()
+	RegisterActivity(oldActivity)
+	RegisterActivity(newActivity)
```
yiminc: oldActivity and newActivity can be defined as func variable as well, just like workflowFn.

internal_workflow.go
```diff
@@ -327,6 +333,7 @@ func (d *syncWorkflowDefinition) Execute(env workflowEnvironment, input []byte)
 	d.rootCtx = WithValue(background, workflowEnvironmentContextKey, env)
 	var resultPtr *workflowResult
 	d.rootCtx = WithValue(d.rootCtx, workflowResultContextKey, &resultPtr)
+	d.rootCtx = WithValue(d.rootCtx, changeVersionsContextKey, make(map[string]*changeVersion))
```
yiminc: why this needs to be on rootCtx? can the versions map be part of workflowEnvironment, just like the sideEffectResult is part of workflowEnvironmentImpl.

internal_workflow.go
```diff
@@ -189,12 +189,18 @@ type (
 
 		Set(value interface{}, err error)
 	}
+
+	changeVersion struct {
+		minSupported Version
```
yiminc: why do we need to remember the minSupported when a version is recorded?

workflow.go
```diff
@@ -540,3 +543,73 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) EncodedValue {
...
+//
+// Currently there is no supported way to completely remove GetVersion call after it was introduced.
+// Keep it even if single branch is left:
+//
+// GetVersion(ctx, "fooChange", 2, 2)
+// err = cadence.ExecuteActivity(ctx, baz).Get(ctx, nil)
+//
+// It is necessary as GetVersion performs validation of a version against a workflow history and fails decisions if
+// a workflow code is not compatible with it.
+func GetVersion(ctx Context, changeID string, minSupported, maxSupported Version) Version {
```
yiminc: can we make this method simply return getWorkflowEnvironment(ctx).GetVersion(changeID, minSupported, maxSupported) and move the logic into workflowEnvironmentImpl?

internal_event_handlers.go
```diff
@@ -633,23 +667,37 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
...
-	var sideEffectID int32
-
-	if err := dec.Decode(&sideEffectID); err != nil {
-		return fmt.Errorf("failure decodeing sideEffectID: %v", err)
-	}
-	var result []byte
-	if err := dec.Decode(&result); err != nil {
-		return fmt.Errorf("failure decoding side effect result: %v", err)
-	}
-	weh.decisionsHelper.handleSideEffectMarkerRecorded(sideEffectID)
```
yiminc: this call is removed, is that intentionally? why?

workflow.go
```diff
@@ -540,3 +543,73 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) EncodedValue {
...
+	versions := ctx.Value(changeVersionsContextKey).(map[string]*changeVersion)
+	cv, ok := versions[changeID]
+	if ok {
+		if cv.minSupported < minSupported {
+			minSupported = cv.minSupported
+		} else {
+			cv.minSupported = minSupported
```
yiminc: where is this used after this set?

workflow.go
```diff
@@ -540,3 +543,73 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) EncodedValue {
...
+		}
+
+		if cv.version != nil {
```
yiminc: when will cv.version == nil?

yiminc:
This work has been landed by https://github.com/uber-go/cadence-client/pull/182

PR #176, opened 2017-06-29
Added public method to get thread dump of a workflow

mfateev:


internal_workflow_client.go
```diff
@@ -241,6 +246,62 @@ GetHistoryLoop:
...
+	workerParams := workerExecutionParameters{
+		TaskList:                  startWorkflowEvent.GetTaskList().GetName(),
+		ConcurrentPollRoutineSize: defaultConcurrentPollRoutineSize,
+		Identity:                  startWorkflowEvent.GetIdentity(),
+		MetricsScope:              wc.metricsScope,
+		Logger:                    logger,
+		EnableLoggingInReplay:     false,
+		UserContext:               context.Background(),
+	}
+	var maxInt64 int64 = math.MaxInt64
```
sivakku: minor: maxInt64 := math.MaxInt64?

internal_workflow_client.go
```diff
@@ -241,6 +246,62 @@ GetHistoryLoop:
...
+		PreviousStartedEventId: &maxInt64,
+		StartedEventId:         &maxInt64,
+		WorkflowExecution:      &s.WorkflowExecution{WorkflowId: &workflowID, RunId: &runID},
+		WorkflowType:           startWorkflowEvent.WorkflowType,
+	}
+	response, stackTrace, err := taskHandler.ProcessWorkflowTask(task, true)
+	if err == nil && response != nil && len(response.GetDecisions()) > 0 && len(stackTrace) == 0 {
+		lastDecision := response.GetDecisions()[0]
+		if lastDecision.GetDecisionType() == s.DecisionType_FailWorkflowExecution {
+			return "", fmt.Errorf("Cannot get stack trace dump for failed worklfow: %s",
```
sivakku: worklfow -> workflow.

client.go
```diff
@@ -78,6 +78,16 @@ type (
 		//	- InternalServiceError
 		GetWorkflowHistory(workflowID string, runID string) (*s.History, error)
 
+		// GetWorkflowThreadDump gets thread dump of a particular workflow.
+		// The errors it can return:
+		//	- EntityNotExistsError
+		//	- BadRequestError
+		//	- InternalServiceError
+		GetWorkflowThreadDump(workflowID string, runID string) (string, error)
```
venkat1109: nit: How about renaming this to GetWorkflowStackTrace() ?  ThreadDump can be misleading in Golang.

client.go
```diff
@@ -78,6 +78,16 @@ type (
...
+
+		// GetWorkflowThreadDumpForHistory gets thread dump of a particular workflow history.
+		GetWorkflowThreadDumpForHistory(*s.History) (string, error)
```
venkat1109: same comment as above.

internal_workflow_client.go
mfateev: It doesn't compile. It looks like := assignment defines maxInt64 as int.

internal_task_pollers.go
```diff
@@ -239,11 +251,21 @@ func (wtp *workflowTaskPoller) poll() (*workflowTask, error) {
...
+		if size > 0 && atDecisionTaskCompletedEventID > 0 {
+			if h.Events[size-1].GetEventId() > atDecisionTaskCompletedEventID {
+				first := h.Events[0].GetEventId() // eventIds start from 1
+				h.Events = h.Events[:atDecisionTaskCompletedEventID-first+1]
+				if h.Events[len(h.Events)-1].GetEventType() != s.EventType_DecisionTaskCompleted {
+					return nil, nil, fmt.Errorf("newGetHistoryPageFunc: atDecisionTaskCompletedEventID(%v) "+
+						"points to event that is not DecisionTaskCompleted", atDecisionTaskCompletedEventID)
+				}
+			}
+			return h, nil, nil
```
yiminc: the nextPageToken is nil only if h.Events[size-1].GetEventId() > atDecisionTaskCompletedEventID is true.

internal_workflow_client.go
```diff
@@ -245,6 +250,65 @@ GetHistoryLoop:
...
+		History:                firstPage,
+		PreviousStartedEventId: &maxInt64,
+		StartedEventId:         &maxInt64,
+		WorkflowExecution:      &s.WorkflowExecution{WorkflowId: &workflowID, RunId: &runID},
+		WorkflowType:           startWorkflowEvent.WorkflowType,
+		NextPageToken:          taskToken,
+	}
+	response, stackTrace, err := taskHandler.ProcessWorkflowTask(task, getHistoryPage, true)
+	if err == nil && response != nil && len(response.GetDecisions()) > 0 && len(stackTrace) == 0 {
+		lastDecision := response.GetDecisions()[0]
```
yiminc: i think you want the last decision, not the first?

internal_public.go
```diff
@@ -36,15 +36,15 @@ import (
 )
 
 type (
-	// WorkflowHistoryIterator - To fetch history given a continuation token.
-	WorkflowHistoryIterator func(nextToken []byte) (*s.History, []byte, error)
+	// GetHistoryPage - Fetches a history page given a continuation token.
+	GetHistoryPage func(nextToken []byte) (*s.History, []byte, error)
```
venkat1109: [suggestion - not a blocker] rename this to NextHistoryPageFunc

internal_task_handlers.go
```diff
@@ -59,8 +59,8 @@ type (
 
 	// workflowTask wraps a decision task.
 	workflowTask struct {
-		task     *s.PollForDecisionTaskResponse
-		iterator WorkflowHistoryIterator
+		task               *s.PollForDecisionTaskResponse
+		getHistoryPageFunc GetHistoryPage
```
venkat1109: same comment as above

internal_public.go
mfateev: I believe GetHistoryPage is a better name as it is going to return the same page given the same token.

internal_workflow_client.go
```diff
@@ -245,6 +250,65 @@ GetHistoryLoop:
...
+		if lastDecision.GetDecisionType() == s.DecisionType_FailWorkflowExecution {
```
yiminc: why do we special handle this fail execution decision? Do we need to special handle all workflow termination decisions the same way?

PR #179, opened 2017-07-06
Logging cleanup

sivakku:


venkat1109:
> My suggestion is to keep current solution, and to add those very noisy logs (like poller logs) behind the enableVerboseLogging check.

How about an alternative idea, API and usage examples below:
```
API:
func Trace(logFunc func()) {
      if tracingEnabled {
             logFunc()
      }
}

Usage:

Trace(func() { log.Debug("this is a test") })
Trace(func() { log.Debug("this is an expensive log")})

```
With this approach, we don't have the annoying if (verbose) checks all over the code, while at the same time, we don't incur a perf penalty unless we actually turn on tracing.

sivakku:
@yiminc 
 for (1) i do agree that we may not have to go that route of single knob as long as we use error log appropriately like dump the errors that usually care. As a first option not introduce that now but clean up logs under this flag as much as possible.
for (3) That just seems like a hack that depends on logging flag to optimize computing the decisions, which by the way that kind of perf doesn't going to impact IMO on client library.

i think with this CR i am getting to a point of every where trying to avoid logging check. i do like what  Venkat suggested but every log has this extra prefix Trace(func()...) which probably is not that bad.

internal_worker_base.go
```diff
@@ -169,10 +175,14 @@ func (bw *baseWorker) execute(routineID int) {
 				if err == nil {
 					bw.retrier.Succeeded()
 				} else if isClientSideError(err) {
-					bw.logger.Info("Poll and processing failed with client side error", zap.Int(tagRoutineID, routineID), zap.Error(err))
+					traceLog(func() {
```
yiminc: this should not be trace level, it indicate something is wrong that probably needs to be looked at.

internal_worker_base.go
```diff
@@ -169,10 +175,14 @@ func (bw *baseWorker) execute(routineID int) {
...
+						bw.logger.Info("Poll and processing failed with client side error", zap.Int(tagRoutineID, routineID), zap.Error(err))
+					})
 					// This doesn't count against server failures.
 				} else {
-					bw.logger.Info("Poll and processing task failed with error", zap.Int(tagRoutineID, routineID), zap.Error(err))
+					traceLog(func() {
```
yiminc: same as this one.

internal_worker_base.go
```diff
@@ -128,13 +128,17 @@ func (bw *baseWorker) Start() {
 	}
 
 	bw.isWorkerStarted = true
-	bw.logger.Info("Started Worker", zap.Int("RoutineCount", bw.options.routineCount))
+	traceLog(func() {
```
yiminc: this is worker life cycle event, which should not be trace level.

internal_worker_base.go
```diff
@@ -128,13 +128,17 @@ func (bw *baseWorker) Start() {
...
+		bw.logger.Info("Started Worker", zap.Int("RoutineCount", bw.options.routineCount))
+	})
 }
 
 func (bw *baseWorker) Run() {
 	bw.Start()
 	d := <-getKillSignal()
-	bw.logger.Info("Worker has been killed", zap.String("Signal", d.String()))
+	traceLog(func() {
+		bw.logger.Info("Worker has been killed", zap.String("Signal", d.String()))
```
yiminc: should not be trace level.

internal_worker_base.go
```diff
@@ -149,7 +153,9 @@ func (bw *baseWorker) Stop() {
 	// poll routines.
 
 	if success := awaitWaitGroup(&bw.shutdownWG, 2*time.Second); !success {
-		bw.logger.Info("Worker timed out on waiting for shutdown.")
+		traceLog(func() {
+			bw.logger.Info("Worker timed out on waiting for shutdown.")
```
yiminc: should not be trace level.

internal_worker_base.go
```diff
@@ -181,7 +191,9 @@ func (bw *baseWorker) execute(routineID int) {
 		select {
 		// Shutdown the Routine.
 		case <-bw.shutdownCh:
-			bw.logger.Info("Worker shutting down.", zap.Int(tagRoutineID, routineID))
+			traceLog(func() {
+				bw.logger.Info("Worker shutting down.", zap.Int(tagRoutineID, routineID))
```
yiminc: should not be trace level.

internal_worker_base.go
sivakku: Tried hiding these logs at worker_base, this is internal layer of worker. Added life cycle logs on the AggregatedWorker in internal_worker.go.  same for below comments as well.

internal_worker_base.go
sivakku: This is too verbose, All poll errors are temporary issues, they are not real errors, since we have enough retires. This is one of the main reasons to hide these logs under verbose level.
if something at poll is important it should be logged as bw.logger.Error(...)

internal_worker_base.go
sivakku: We have one at the top level at Aggregated worker.

sivakku:
@yiminc - yep, its not improvement in terms of writing logs. the main purpose was to hide logs of worker. The only improvement in the way it is presented now, we can add complex logic inside the common function with out repeating, like some more levels or flags can be just added there.

PR #181, opened 2017-07-10
Update README.md

molteanu:


README.md
```diff
@@ -23,9 +23,9 @@ See [samples](https://github.com/samarabbas/cadence-samples) to get started
...
+Activities are implemented as functions. Data can be passed directly to an activity via function parameters. The parameters can be either basic types or structs, with the only requirement being that the parameters need to be serializable. Even though it is not required, we recommand that the first parameter of an activity function is of type `context.Context`, in order to allow the activity to interact with other framework methods. The function must return an `error` value, and can optionally return a result value. The result value can be either a basic type or a struct with the only requirement being that it is serializable.
```
sivakku: "The function must return an `error` value, and can optionally return a result value."

For activities we allow them to have three possible return types.
   - Void (No return values, like quick loggers, where the caller doesn't care)
   - error (Just error)
   - (result, error) - result can be any serializable value.

PR #182, opened 2017-07-11
GetVersion implementation

yiminc:
This is based on https://github.com/uber-go/cadence-client/pull/170, the only change is changing the map that attached to rootCtx from type map[string]*changeVersion to map[string]Version, as a result, also change the cadence.GetVersion() code a little bit.

### DIFF
--- a/internal_decision_state_machine.go
+++ b/internal_decision_state_machine.go
@@ -23,9 +23,6 @@ package cadence
 import (
 	"fmt"
 
-	"bytes"
-	"encoding/gob"
-
 	s "go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/common"
 	"go.uber.org/cadence/common/util"
@@ -686,15 +683,10 @@ func (h *decisionsHelper) getActivityID(event *s.HistoryEvent) string {
 
 func (h *decisionsHelper) recordVersionMarker(changeID string, version Version) decisionStateMachine {
 	markerID := fmt.Sprintf("%v_%v", versionMarkerName, changeID)
-	var buf bytes.Buffer
-	enc := gob.NewEncoder(&buf)
-	if err := enc.Encode(changeID); err != nil {
-		panic(fmt.Sprintf("Failure encoding changeID name: %v", err))
+	details, err := getHostEnvironment().encodeArgs([]interface{}{changeID, version})
+	if err != nil {
+		panic(err)
 	}
-	if err := enc.Encode(version); err != nil {
-		panic(fmt.Sprintf("Failure encoding version value: %v", err))
-	}
-	details := buf.Bytes()
 
 	recordMarker := &s.RecordMarkerDecisionAttributes{
 		MarkerName: common.StringPtr(versionMarkerName),
--- a/internal_decision_state_machine.go
+++ b/internal_decision_state_machine.go
@@ -23,6 +23,9 @@ package cadence
 import (
 	"fmt"
 
+	"bytes"
+	"encoding/gob"
+
 	s "go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/common"
 	"go.uber.org/cadence/common/util"
@@ -136,6 +139,11 @@ const (
 	eventCancelInitiated  string = "handleCancelInitiatedEvent"
 	eventCancelFailed     string = "handleCancelFailedEvent"
 	eventCanceled         string = "handleCanceledEvent"
+)
+
+const (
+	sideEffectMarkerName = "SideEffect"
+	versionMarkerName    = "Version"
 )
 
 func (d decisionState) String() string {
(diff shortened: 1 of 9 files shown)


internal_event_handlers.go
```diff
@@ -628,28 +662,41 @@ func (weh *workflowExecutionEventHandlerImpl) handleWorkflowExecutionCancelReque
...
 
-// Currently handles only side effect markers
 func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
 	eventID int64,
 	attributes *m.MarkerRecordedEventAttributes,
 ) error {
-	if attributes.GetMarkerName() != sideEffectMarkerName {
+	switch attributes.GetMarkerName() {
+	case sideEffectMarkerName:
+		dec := gob.NewDecoder(bytes.NewBuffer(attributes.GetDetails()))
```
molteanu: NIT: Could would be more readable if you extract the decoding into a helper method. (Same comment for decoding the "version marker"

internal_decision_state_machine.go
```diff
@@ -676,6 +684,28 @@ func (h *decisionsHelper) getActivityID(event *s.HistoryEvent) string {
 	return activityID
 }
 
+func (h *decisionsHelper) recordVersionMarker(changeID string, version Version) decisionStateMachine {
+	markerID := fmt.Sprintf("%v_%v", versionMarkerName, changeID)
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	if err := enc.Encode(changeID); err != nil {
```
sivakku: Why are you not going through encoding interface exposed through hostEnv.Encoder()? that would give us single place to switch different encoders.

internal_event_handlers.go
```diff
@@ -335,6 +335,40 @@ func (wc *workflowEnvironmentImpl) RequestCancelTimer(timerID string) {
...
+func validateVersion(changeID string, version, minSupported, maxSupported Version) {
+	if version < minSupported {
+		panic(fmt.Sprintf("Workflow code removed support of version %v. "+
+			"for \"%v\" changeID. The oldest supported version is %v",
+			version, changeID, minSupported))
+	}
+	if version > maxSupported {
+		panic(fmt.Sprintf("Workflow code is too old to support version %v "+
+			"for \"%v\" changeID. The maximum supported version is %v",
+			version, changeID, minSupported))
```
sivakku: minSupported -> maxSupported

internal_event_handlers.go
```diff
@@ -335,6 +335,40 @@ func (wc *workflowEnvironmentImpl) RequestCancelTimer(timerID string) {
 	wc.logger.Debug("RequestCancelTimer", zap.String(tagTimerID, timerID))
 }
 
+func validateVersion(changeID string, version, minSupported, maxSupported Version) {
```
sivakku: Can these methods return error message and panic at GetVersion() level? sometimes seeing too many panics in library code gives a feeling that panic is accepted pattern but we only do it in certain method calls, where error is not obvious choice.

internal_event_handlers.go
```diff
@@ -335,6 +335,40 @@ func (wc *workflowEnvironmentImpl) RequestCancelTimer(timerID string) {
...
+func (wc *workflowEnvironmentImpl) GetVersion(changeID string, minSupported, maxSupported Version) Version {
+	var version Version
+	if wc.isReplay {
+		var ok bool
+		version, ok = wc.changeVersions[changeID]
+		if !ok {
+			version = DefaultVersion
+		}
+		validateVersion(changeID, version, minSupported, maxSupported)
+		wc.logger.Debug("GetVersion return version from a marker",
```
sivakku: Can you hide this logger under the verbose check?

internal_event_handlers.go
```diff
@@ -335,6 +335,40 @@ func (wc *workflowEnvironmentImpl) RequestCancelTimer(timerID string) {
...
+			zap.String(tagChangeID, changeID), zap.Int(tagVersion, int(version)))
+		return version
+	}
+	version = maxSupported
+	wc.decisionsHelper.recordVersionMarker(changeID, version)
+
+	wc.logger.Debug("GetVersion return",
```
sivakku: same here? the more it looks this flag of checking to log looks hacky all over place.

internal_event_handlers.go
sivakku: Why are you not using encoding interface on hostEnv to decode, so we can switch easily with a different encoder.

internal_worker_base.go
```diff
@@ -53,6 +53,7 @@ type (
 		asyncActivityClient
 		workflowTimerClient
 		SideEffect(f func() ([]byte, error), callback resultHandler)
+		GetVersion(changeID string, maxSupported, minSupported Version) Version
```
sivakku: The order of variables is inversed. minSupported, maxSupported. Can you add unit test for this layer as well?

internal_workflow.go
```diff
@@ -327,6 +328,7 @@ func (d *syncWorkflowDefinition) Execute(env workflowEnvironment, input []byte)
 	d.rootCtx = WithValue(background, workflowEnvironmentContextKey, env)
 	var resultPtr *workflowResult
 	d.rootCtx = WithValue(d.rootCtx, workflowResultContextKey, &resultPtr)
+	d.rootCtx = WithValue(d.rootCtx, changeVersionsContextKey, make(map[string]Version))
```
sivakku: What is the value if having this look up in this layer when you have the lookup already in task handler GetVersion()?

yiminc: This layer server as a cache so we always return same version for same changeID, only if the changeID is known ID. The lookup at workflowEnvironment layer serve lookup from marker result. If a changeID is not present at workflowEnvironment layer, we would return DefaultVersion, but if a changeID is not present at here, we won't return DefaultVersion.

yiminc: the 2 layer lookup can be combined, just updated.

internal_event_handlers.go
yiminc: This is a helper method that has no much of logic inside. If we would return flag here and panic in GetVersion() level, then this method does not need to be exists.

sivakku: you can return error instead of flag. it is still good to have validation as a separate method.

yiminc: Use error when invalid version is expected in legit case, and call site can handle that error accordingly. In this case, invalid version is not expected, and it is not recoverable. So I think panic is appropriate here.

sivakku: instead of panic'ing here, panic inside GetVersion() if validateVersion() returns error. so it is very clear that method panics.

internal_workflow_testsuite_test.go
```diff
@@ -910,3 +910,34 @@ func (s *WorkflowTestSuiteUnitTest) Test_ChildWithChild() {
...
+	oldActivity := func(ctx context.Context, msg string) (string, error) {
+		return "hello" + "_" + msg, nil
+	}
+	newActivity := func(ctx context.Context, msg string) (string, error) {
+		return "hello" + "_" + msg, nil
+	}
+	workflowFn := func(ctx Context) error {
+		ctx = WithActivityOptions(ctx, s.activityOptions)
+		var f Future
+		v := GetVersion(ctx, "test_change_id", DefaultVersion, 2)
```
mfateev: Looking at this we probably want to change DefaultVersion value to 1.

PR #183, opened 2017-07-11
Add worker method for blocking start to enable clean exit

madhuravi:
Fixes https://github.com/uber/cadence/issues/254. Today client is killed, defer functions do not get called. The context does not get cancelled and in turn the server is not notified.

This adds a blocking Run method that cadence users can use that takes care of clean exit. Users that don't want a blocking call can use Start, wait on Done channel for exit and call Stop explicitly.

Also fixes a waitgroup bug in Stop.

### DIFF
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -739,10 +739,10 @@ func (aw *aggregatedWorker) Start() error {
 }
 
 func (aw *aggregatedWorker) Run() {
-	Start()
-	d := <-Done()
+	aw.Start()
+	d := <-aw.Done()
 	bw.logger.Info("Worker has been killed: %s", d.String())
-	Stop()
+	aw.Stop()
 }
 
 func (aw *aggregatedWorker) Stop() {
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -731,7 +731,7 @@ func (aw *aggregatedWorker) Start() error {
 
 func (aw *aggregatedWorker) Run() {
 	aw.Start()
-	d := <-isProcessKilled()
+	d := <-getKillSignal()
 	aw.logger.Info("Worker has been killed", zap.String("Signal", d.String()))
 	aw.Stop()
 }
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -27,6 +27,7 @@ import (
 	"context"
 	"encoding/gob"
 	"fmt"
+	"os"
 	"reflect"
 	"runtime"
 	"strconv"
@@ -199,9 +200,17 @@ func (ww *workflowWorker) Start() error {
 	return nil // TODO: propagate error
 }
 
+func (ww *workflowWorker) Run() {
+	ww.worker.Run()
+}
+
 // Shutdown the worker.
 func (ww *workflowWorker) Stop() {
 	ww.worker.Stop()
+}
+
+func (ww *workflowWorker) Done() <-chan os.Signal {
+	return ww.worker.Done()
 }
 
 func newActivityWorker(
@@ -258,9 +267,19 @@ func (aw *activityWorker) Start() error {
 	return nil // TODO: propagate errors
 }
 
+// Run the worker.
+func (aw *activityWorker) Run() {
+	aw.worker.Run()
+}
+
 // Shutdown the worker.
 func (aw *activityWorker) Stop() {
 	aw.worker.Stop()
+}
+
+// Done returns the exit channel
+func (aw *activityWorker) Done() chan os.Signal {
+	return aw.worker.Done()
 }
 
 type workerFunc func(ctx Context, input []byte) ([]byte, error)
@@ -719,6 +738,13 @@ func (aw *aggregatedWorker) Start() error {
 	return nil
 }
 
+func (aw *aggregatedWorker) Run() {
+	Start()
+	d := <-Done()
+	bw.logger.Info("Worker has been killed: %s", d.String())
+	Stop()
+}
+
 func (aw *aggregatedWorker) Stop() {
 	if !isInterfaceNil(aw.workflowWorker) {
 		aw.workflowWorker.Stop()
@@ -726,6 +752,17 @@ func (aw *aggregatedWorker) Stop() {
 	if !isInterfaceNil(aw.activityWorker) {
 		aw.activityWorker.Stop()
 	}
+}
+
+func (aw *aggregatedWorker) Done() chan os.Signal {
+	c := make(chan os.Signal, 1)
+	select {
+	case w := <-aw.workflowWorker.Done():
+		c <- w
+	case a := <-aw.activityWorker.Done():
+		c <= a
+	}
+	return c
 }
 
 // aggregatedWorker returns an instance to manage the workers.
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -740,8 +740,7 @@ func (aw *aggregatedWorker) Start() error {
 
 func (aw *aggregatedWorker) Run() {
 	aw.Start()
-	d := <-aw.Done()
-	bw.logger.Info("Worker has been killed: %s", d.String())
+	<-aw.Done()
 	aw.Stop()
 }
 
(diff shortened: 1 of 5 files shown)


worker.go
```diff
@@ -31,8 +32,14 @@ import (
 type (
 	// Worker represents objects that can be started and stopped.
 	Worker interface {
-		Stop()
+		// Start starts the worker in a non-blocking fashion
 		Start() error
+		// Run is a blocking start and cleans up resources when killed
+		Run()
```
sivakku: would merging with Start() simplify exposing another method? 
Start(waitForCompletion bool) {...}

madhuravi: I prefer having separate methods in general, boolean arguments determining behavior is not ideal. See https://martinfowler.com/bliki/FlagArgument.html, https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior

venkat1109:
> Today client is killed, defer functions do not get called. The context does not get cancelled and in turn the server is not notified.This adds a blocking Run method that cadence users can use that takes care of clean exit.

When the client is killed with a ctrl-c, os kernel will shutdown all tcp conns / release all fds owned by that process. If the server still waits for pollTimeout period, it makes make think there is a bug on the server wrt handling of TCP_FIN / TCP shutdown, probably within tchannel. Adding a signal handler and forcing clean shutdown on the client will only mitigate the problem, it won't fix the root cause. 

On a side note, exposing Run() and Done() methods on the worker feels a bit weird to me. Done is really just a wrapper around signal.Notify. What does it mean when the app calls Done() more than once in two different go routines ? Done() doesn't mean the worker is done, it just means `a signal is received` with your implementation

Would it rather make sense to not expose the Run()/Done() methods, but have the client lib always register a signal notify handler (as part of Start()) and do graceful cleanup ?

venkat1109:
>  os kernel will shutdown all tcp conns
Actually, now I think about it, this might be true only if TCP_KEEPALIVES are enabled. If that's the case, your fix on the client side does make sense to me. Let me know if you are able to confirm this is the case. If this is the case
* we should also follow up with tchannel on the support for keep alives
* I am cool this PR, except for the Done() method, which feels weird

madhuravi:
@venkat1109 I already spoke to tchannel and they don't have support for this. See https://github.com/uber/tchannel-go/pull/318. I spoke to their team offline as well.

Done() method is for users that don't want a blocking start. Might call start, do some other work, then wait for Done and call Stop. This follows golang's context model https://golang.org/src/context/context.go?s=2437:5880#L52. Another option is just to return a struct channel instead of a signal channel if we don't want users to care about that.

madhuravi:
To address your earlier question, Done is not meant to be called by multiple goroutines. Similar to Start/Stop/Run should be called once during startup. Seems to be the recommended practice in golang for graceful shutdown.

internal_worker_base.go
```diff
@@ -141,33 +150,42 @@ func (bw *baseWorker) Stop() {
 	// TODO: The poll is longer than the 10 seconds, we probably need some way to hard terminate the
 	// poll routines as well.
 
-	if success := awaitWaitGroup(&bw.shutdownWG, 10*time.Second); !success {
+	if success := awaitWaitGroup(&bw.shutdownWG, 2*time.Second); !success {
```
madhuravi: Note to reviewers: I changed this to 2 seconds. No reason for Stop to take very long. And after the bug fix that is part of this PR, Stop usually happens very quickly < 1 sec.

internal_worker_base.go
```diff
@@ -141,33 +150,42 @@ func (bw *baseWorker) Stop() {
...
-			bw.retrier.Succeeded()
-		} else if isClientSideError(err) {
-			bw.logger.Info("Poll and processing failed with client side error", zap.Int(tagRoutineID, routineID), zap.Error(err))
-			// This doesn't count against server failures.
-		} else {
-			bw.logger.Info("Poll and processing task failed with error", zap.Int(tagRoutineID, routineID), zap.Error(err))
-			bw.retrier.Failed()
-		}
+		ch := make(chan struct{})
+		go func() {
```
madhuravi: This is the bug fix for goroutines not getting closed correctly.

internal_worker.go
```diff
@@ -199,11 +200,19 @@ func (ww *workflowWorker) Start() error {
...
+func (ww *workflowWorker) Run() {
+	ww.worker.Run()
+}
+
 // Shutdown the worker.
 func (ww *workflowWorker) Stop() {
 	ww.worker.Stop()
 }
 
+func (ww *workflowWorker) Done() <-chan os.Signal {
```
yiminc: why do we need a chan of os.Signal? What is that signal used for?

internal_worker.go
```diff
@@ -728,6 +755,17 @@ func (aw *aggregatedWorker) Stop() {
...
 }
 
+func (aw *aggregatedWorker) Done() <-chan os.Signal {
+	c := make(chan os.Signal, 1)
+	select {
+	case w := <-aw.workflowWorker.Done():
+		c <- w
+	case a := <-aw.activityWorker.Done():
+		c <- a
+	}
```
yiminc: should it be OR (either one) or AND (both of the 2)?

internal_worker_base.go
```diff
@@ -141,33 +147,36 @@ func (bw *baseWorker) Stop() {
...
```
yiminc: Why do we need a new goroutinue for each task processing? Could you explain (put comments in code) why this is necessary and how it fixes the bug you mentioned.

internal_worker.go
madhuravi: This is not there anymore. See updated diff.

internal_worker.go
madhuravi: This is not there anymore. See updated diff.

internal_worker_base.go
madhuravi: This is basically how shutdown channels are meant to work. Earlier, a shutdown signal will never be received until it finishes the poll and reaches the select statement. So goroutines never get exited since wait timeout (10 sec previously) < poll timeout (60 sec).
Making this a goroutine so that all work is stopped when a shutdown is received. (Not sure if this is worth a comment since this is how goroutines/shutdown channels in go are meant to work)

internal_worker_base.go
```diff
@@ -141,33 +147,36 @@ func (bw *baseWorker) Stop() {
...
+					bw.retrier.Succeeded()
+				} else if isClientSideError(err) {
+					bw.logger.Info("Poll and processing failed with client side error", zap.Int(tagRoutineID, routineID), zap.Error(err))
+					// This doesn't count against server failures.
+				} else {
+					bw.logger.Info("Poll and processing task failed with error", zap.Int(tagRoutineID, routineID), zap.Error(err))
+					bw.retrier.Failed()
+				}
+			}
+			ch <- struct{}{}
```
venkat1109: do you need this when you have defer close(ch) ? Also consider passing the ch as a func param to the go routine

madhuravi: good catch.

internal_worker_base.go
madhuravi: Couldn't find tchannel close, will follow up in separate PR if I find the method.

Spinning up new goroutines is cheap, any specific concerns with that? The advantage here is though the poll doesn't complete, the for loop is killed. With this, the inner goroutine with poll will continue but the outer for loop will not.

venkat1109: tchannel.Close lives here:
https://github.com/uber/tchannel-go/blob/dev/channel.go#L747

but its not a trivial change, because the channel gets created at the app, not the library today, see example here:
https://github.com/samarabbas/cadence-samples/blob/master/cmd/samples/common/factory.go#L96

So, yes, we should tackle this separately. Let's merge this PR / fix

worker.go
yiminc: I think we should merge Run() and Start(). Offering too much feature confuse people. For those who don't want to be blocked, can easily do that by using a separate goroutinue to call Start().

madhuravi: Agree that minimal options are better in general. Here we need:
1. a blocking start method
2. a non blocking start method that returns once the worker is started

The challenge with starting in a goroutine is the user has no way of knowing if the worker even started/errored out etc. So they will proceed doing work when the worker might actually not have started at all.

internal_worker_base.go
sivakku: minor: Can you update/remove the comment above it?

PR #185, opened 2017-07-13
Select Future stale registered callback can modify state

sivakku:
Selector future can have many callbacks registered with the channel for receiving notifications on each ask. One of the stale callback getting executed while closing the channel(after getting a value on the future) can modify the current state that indicates that we have processed this future for the select. This with batch replay can end loosing notifying select.

### DIFF
--- a/internal_coroutines_test.go
+++ b/internal_coroutines_test.go
@@ -1044,24 +1044,24 @@ func TestSelectFuture_WithBatchSets(t *testing.T) {
 
 		s := NewSelector(ctx)
 		s.
-		AddFuture(future1, func(f Future) {
-			var v string
-			err := f.Get(ctx, &v)
-			assert.Nil(t, err)
-			history = append(history, fmt.Sprintf("c1-%v", v))
-		}).
+			AddFuture(future1, func(f Future) {
+				var v string
+				err := f.Get(ctx, &v)
+				assert.Nil(t, err)
+				history = append(history, fmt.Sprintf("c1-%v", v))
+			}).
 			AddFuture(future2, func(f Future) {
-			var v string
-			err := f.Get(ctx, &v)
-			assert.Nil(t, err)
-			history = append(history, fmt.Sprintf("c2-%v", v))
-		}).
+				var v string
+				err := f.Get(ctx, &v)
+				assert.Nil(t, err)
+				history = append(history, fmt.Sprintf("c2-%v", v))
+			}).
 			AddFuture(future3, func(f Future) {
-			var v string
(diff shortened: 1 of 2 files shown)


internal_coroutines_test.go
```diff
@@ -1034,3 +1034,53 @@ func TestDecodeFutureChain(t *testing.T) {
...
 }
+
+func TestSelectFuture_WithBatchSets(t *testing.T) {
+	var history []string
+	d := newDispatcher(background, func(ctx Context) {
+		future1, settable1 := NewFuture(ctx)
+		future2, settable2 := NewFuture(ctx)
+		future3, settable3 := NewFuture(ctx)
+
+		s := NewSelector(ctx)
```
mfateev: What is going to happen if the same future added multiple times?

sivakku: The second time select(more than 1), gets blocked for ever. May be that is acceptable for now. we could detect and panic if that is any better than first one.

mfateev: What is the behaviour of the go select?

sivakku: But GO doesn't have something equivalent to future to compare behaviours. Here is the play ground example for reading on a channel, it seems to always pick first one. https://play.golang.org/p/cSDH3vxMtH
Our go on select future will also pick first callback always.

mfateev: Then I think we should unblock 3 times for 3 callbacks of the same future.

sivakku: hmm...  in the example in the playground, i am writing twice on to channel and hence select executes twice. If i write only once then the select is blocked.
In case of cadence.Future, setting more than once would panic, so  it makes sense that we can only set once on future and that means we can only select once on the same future. 
Unblocking 3 times would add confusion since future is not allowed to set 3 times.

To be fair with our logic that we panic on second Set on future, we can as well panic on second Select on the same future instead of blocking. any thoughts?

mfateev: Select is about reading. Future can be read multiple times as opposed to a channel. So I don't see a problem calling each callback once for a future that is ready.

yiminc: For go's select: If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection.
https://golang.org/ref/spec#Select_statements

sivakku: i think it make sense if we look select as reading multiple times to unblock it 3 times. will do a separate CR tomorrow.
@yiminc - Which is the reason we have our own select to have deterministic order.

PR #186, opened 2017-07-19
threading model

yiminc:
break PollAndProcessSingleTask() into 2 methods: PollTask() and ProcessTask(). Restructure the code to have poller that polls task and send task to taskQueueCh then a task dispatcher is receiving from that task queue channel and dispatch to process it. Task rate limiter is also added.

### DIFF
--- a/internal_task_pollers.go
+++ b/internal_task_pollers.go
@@ -48,9 +48,12 @@ var (
 )
 
 type (
-	// taskPoller interface to poll for a single task
+	// taskPoller interface to poll and process for task
 	taskPoller interface {
-		PollAndProcessSingleTask() error
+		// PollTask polls for one new task
+		PollTask() (interface{}, error)
+		// ProcessTask processes a task
+		ProcessTask(interface{}) error
 	}
 
 	// workflowTaskPoller implements polling/processing a workflow task
@@ -122,8 +125,8 @@ func newWorkflowTaskPoller(taskHandler WorkflowTaskHandler, service m.TChanWorkf
 	}
 }
 
-// PollAndProcessSingleTask process one single task
-func (wtp *workflowTaskPoller) PollAndProcessSingleTask() error {
+// PollTask polls a new task
+func (wtp *workflowTaskPoller) PollTask() (interface{}, error) {
 	startTime := time.Now()
 	defer func() {
 		deltaTime := time.Now().Sub(startTime)
@@ -136,8 +139,14 @@ func (wtp *workflowTaskPoller) PollAndProcessSingleTask() error {
 	// Get the task.
 	workflowTask, err := wtp.poll()
 	if err != nil {
-		return err
+		return nil, err
 	}
+	return workflowTask, nil
+}
+
+// ProcessTask processes a task
+func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
+	workflowTask := task.(*workflowTask)
 	if workflowTask.task == nil {
 		// We didn't have task, poll might have time out.
 		traceLog(func() {
@@ -178,6 +187,7 @@ func (wtp *workflowTaskPoller) PollAndProcessSingleTask() error {
 		return err
 	}
 	return nil
+
 }
 
 // Poll for a single workflow task from the service
@@ -280,8 +290,8 @@ func (atp *activityTaskPoller) poll() (*activityTask, error) {
 	return &activityTask{task: response}, nil
 }
 
-// PollAndProcessSingleTask process one single activity task
-func (atp *activityTaskPoller) PollAndProcessSingleTask() error {
+// PollTask polls a new task
+func (atp *activityTaskPoller) PollTask() (interface{}, error) {
 	startTime := time.Now()
 	defer func() {
 		deltaTime := time.Now().Sub(startTime)
@@ -294,8 +304,14 @@ func (atp *activityTaskPoller) PollAndProcessSingleTask() error {
 	// Get the task.
 	activityTask, err := atp.poll()
 	if err != nil {
-		return err
+		return nil, err
 	}
+	return activityTask, nil
+}
+
+// ProcessTask processes a new task
+func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
+	activityTask := task.(*activityTask)
 	if activityTask.task == nil {
 		// We didn't have task, poll might have time out.
 		traceLog(func() {
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -47,7 +47,7 @@ const (
 	defaultMaxConcurrentActivityExecutionSize = 1000   // Large concurrent activity execution size (1k)
 	defaultMaxActivityExecutionRate           = 100000 // Large activity execution rate (unlimited)
 
-	defaultMaxConcurrentWorkflowExecutionSize = 10     // workflow code normally run fast, so only need small concurrent number
+	defaultMaxConcurrentWorkflowExecutionSize = 50     // hardcoded max workflow execution size.
 	defaultMaxWorkflowExecutionRate           = 100000 // Large workflow execution rate (unlimited)
 )
 
@@ -107,7 +107,7 @@ type (
 		ConcurrentActivityExecutionSize int
 
 		// Defines rate limiting on number of activity tasks that can be executed per second.
-		MaxActivityExecutionsPerSecond int
+		MaxActivityExecutionsPerSecond float32
 
 		// User can provide an identity for the debuggability. If not provided the framework has
 		// a default option.
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -40,7 +40,9 @@ import (
 )
 
 const (
-	defaultConcurrentPollRoutineSize = 1 // set to 1 for now, can adjust later if needed
+	// Set to 1 for now, can adjust later if needed. The typical RTT (round-trip time) is below 1ms within data center.
+	// With 1 poller, we could easily achieve a few hundred RPS.
+	defaultConcurrentPollRoutineSize = 1
 
 	defaultMaxConcurrentActivityExecutionSize = 1000   // Large concurrent activity execution size (1k)
 	defaultMaxActivityExecutionRate           = 100000 // Large activity execution rate (unlimited)
@@ -105,7 +107,7 @@ type (
 		ConcurrentActivityExecutionSize int
 
 		// Defines rate limiting on number of activity tasks that can be executed per second.
-		MaxActivityExecutionRate int
+		MaxActivityExecutionRatePerSecond int
 
 		// User can provide an identity for the debuggability. If not provided the framework has
 		// a default option.
@@ -250,7 +252,7 @@ func newActivityTaskWorker(
 	base := newBaseWorker(baseWorkerOptions{
 		pollerCount:       workerParams.ConcurrentPollRoutineSize,
 		maxConcurrentTask: workerParams.ConcurrentActivityExecutionSize,
-		maxTaskRps:        workerParams.MaxActivityExecutionRate,
+		maxTaskRps:        workerParams.MaxActivityExecutionRatePerSecond,
 		taskWorker:        poller,
 		workflowService:   service,
 		identity:          workerParams.Identity,
@@ -758,7 +760,9 @@ func (aw *aggregatedWorker) Stop() {
 	aw.logger.Info("Stopped Worker")
 }
 
-// aggregatedWorker returns an instance to manage the workers.
+// aggregatedWorker returns an instance to manage the workers. Use defaultConcurrentPollRoutineSize (which is 1) as
+// poller size. The typical RTT (round-trip time) is below 1ms within data center. With 1 poller, we could easily achieve
+// a few hundred RPS.
 func newAggregatedWorker(
 	service m.TChanWorkflowService,
 	domain string,
@@ -767,15 +771,15 @@ func newAggregatedWorker(
 ) (worker Worker) {
 	wOptions := fillWorkerOptionsDefaults(options)
 	workerParams := workerExecutionParameters{
-		TaskList:                        taskList,
-		ConcurrentPollRoutineSize:       defaultConcurrentPollRoutineSize,
-		ConcurrentActivityExecutionSize: wOptions.MaxConcurrentActivityExecutionSize,
-		MaxActivityExecutionRate:        wOptions.MaxActivityExecutionRate,
-		Identity:                        wOptions.Identity,
-		MetricsScope:                    wOptions.MetricsScope,
-		Logger:                          wOptions.Logger,
-		EnableLoggingInReplay:           wOptions.EnableLoggingInReplay,
-		UserContext:                     wOptions.BackgroundActivityContext,
+		TaskList:                          taskList,
+		ConcurrentPollRoutineSize:         defaultConcurrentPollRoutineSize,
+		ConcurrentActivityExecutionSize:   wOptions.MaxConcurrentActivityExecutionSize,
+		MaxActivityExecutionRatePerSecond: wOptions.MaxActivityExecutionRatePerSecond,
+		Identity:              wOptions.Identity,
+		MetricsScope:          wOptions.MetricsScope,
+		Logger:                wOptions.Logger,
+		EnableLoggingInReplay: wOptions.EnableLoggingInReplay,
+		UserContext:           wOptions.BackgroundActivityContext,
 	}
 
 	ensureRequiredParams(&workerParams)
@@ -957,8 +961,8 @@ func fillWorkerOptionsDefaults(options WorkerOptions) WorkerOptions {
 	if options.MaxConcurrentActivityExecutionSize == 0 {
 		options.MaxConcurrentActivityExecutionSize = defaultMaxConcurrentActivityExecutionSize
 	}
-	if options.MaxActivityExecutionRate == 0 {
-		options.MaxActivityExecutionRate = defaultMaxActivityExecutionRate
+	if options.MaxActivityExecutionRatePerSecond == 0 {
+		options.MaxActivityExecutionRatePerSecond = defaultMaxActivityExecutionRate
 	}
 	return options
 }
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -40,9 +40,13 @@ import (
 )
 
 const (
-	defaultConcurrentPollRoutineSize          = 2
-	defaultMaxConcurrentActivityExecutionSize = 10000  // Large execution size(unlimited)
-	defaultMaxActivityExecutionRate           = 100000 // Large execution rate(100K per sec)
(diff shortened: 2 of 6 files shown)


internal_worker_base.go
```diff
@@ -121,18 +133,128 @@ func (bw *baseWorker) Start() {
...
+		bw.taskSemaphore <- struct{}{}
+	}
+
+	for {
+		// wait for new task or shutdown
+		select {
+		case <-bw.shutdownCh:
+			return
+		case task := <-bw.taskQueueCh:
+			bw.pollerRequestCh <- struct{}{}
```
venkat1109: this can result in pre-fetching of tasks by the pollers. You only want to give a new token to the poller after a worker becomes free. In your case, this will likely be inside processTask()

internal_worker_base.go
```diff
@@ -121,18 +133,128 @@ func (bw *baseWorker) Start() {
...
+		return true
+	default:
+		return false
+	}
+}
+
+func (bw *baseWorker) runPoller() {
+	defer bw.shutdownWG.Done()
+
+	bw.pollerRequestCh <- struct{}{} // initial poll request to kick off the poller loop
```
venkat1109: couple of problems
* this cannot be inside runPoller, this is something that must be shared across all the poller routines
* you need to initialize this with maxConcurrentTask tokens (or) you are limiting itself to only one outstanding poll request at any given point in time. This won't keep all of your workers busy

internal_worker_base.go
```diff
@@ -121,18 +133,128 @@ func (bw *baseWorker) Start() {
...
+
+	if task != nil {
+		bw.taskQueueCh <- task
+	} else {
+		bw.pollerRequestCh <- struct{}{} // poll failed, trigger a new pool
+	}
+}
+
+func (bw *baseWorker) processTask(task interface{}) {
+	err := bw.options.taskWorker.ProcessTask(task)
```
venkat1109: it would be cool to separate out the RespondDecisionCompleted/RespondActivityTaskCompleted calls as well into a new stage in the task pipeline. May be for later, when we need it.

internal_worker.go
```diff
@@ -757,13 +767,15 @@ func newAggregatedWorker(
...
 	workerParams := workerExecutionParameters{
-		TaskList:                  taskList,
-		ConcurrentPollRoutineSize: defaultConcurrentPollRoutineSize,
-		Identity:                  wOptions.Identity,
-		MetricsScope:              wOptions.MetricsScope,
-		Logger:                    wOptions.Logger,
-		EnableLoggingInReplay:     wOptions.EnableLoggingInReplay,
-		UserContext:               wOptions.BackgroundActivityContext,
+		TaskList:                        taskList,
+		ConcurrentPollRoutineSize:       defaultConcurrentPollRoutineSize,
```
venkat1109: add method header comments on the default values for the pollerCount/concurrentTask count etc. Also would be cool to provide a rough estimation of the task throughput they can achieve with the default values and when they should consider changing this value.

internal_worker_base.go
```diff
@@ -107,11 +114,16 @@ func createPollRetryPolicy() backoff.RetryPolicy {
...
 func newBaseWorker(options baseWorkerOptions, logger *zap.Logger) *baseWorker {
 	return &baseWorker{
-		options:     options,
-		shutdownCh:  make(chan struct{}),
-		rateLimiter: common.NewTokenBucket(1000, common.NewRealTimeSource()),
-		retrier:     backoff.NewConcurrentRetrier(pollOperationRetryPolicy),
-		logger:      logger.With(zapcore.Field{Key: tagWorkerType, Type: zapcore.StringType, String: options.workerType}),
+		options:         options,
+		shutdownCh:      make(chan struct{}),
+		pollRateLimiter: common.NewTokenBucket(1000, common.NewRealTimeSource()),
```
venkat1109: why need this and why 1000 ?

worker.go
```diff
@@ -46,14 +46,14 @@ type (
...
-		// default: defaultMaxConcurrentActivityExecutionSize(10k)
+		// default: defaultMaxConcurrentActivityExecutionSize(1k)
 		MaxConcurrentActivityExecutionSize int
 
 		// Optional: Sets the rate limiting on number of activities that can be executed per second.
 		// This can be used to protect down stream services from flooding.
 		// The zero value of this uses the default value.
 		// default: defaultMaxActivityExecutionRate(100k)
-		MaxActivityExecutionRate float32
+		MaxActivityExecutionRate int
```
venkat1109: Can we rename this to MaxActivityExecutionRatePerSec , then we don't need the additional comment, its self documenting

venkat1109: Also add a warning here that the timeout starts ticking if a task worker is blocked due to rate limiting

internal_worker_base.go
yiminc: I think we want to protect our server and randomly put 1000 at first, this is what it was, i didn't change it in this diff.

internal_worker_base.go
yiminc: I thought about that. The problem is that it won't work nicely with rate limiter. In the case of low RPS with high maxConcurrentTask, it will end up polling down a bunch of tasks that will only be blocked by rate limiter. The dispatcher does not actually process the task, it just receive from the channel and start a new routine to process it. With current setup, as soon as poller sends the task to the channel, dispatcher will receive it and it will send signal to pollerRequestCh immediately after which will cause the poller to continue polling so there is no blocking at all.

internal_worker_base.go
yiminc: One task pre-fetching at worst with this approach. If we do it after processTask, we are starving pollers.

yiminc: We can do better by checking if the concurrent task count is still less than the max, and if so we send poller request before process task, otherwise we send poller request after process task.

internal_task_pollers.go
```diff
@@ -48,9 +48,12 @@ var (
 )
 
 type (
-	// taskPoller interface to poll for a single task
+	// taskPoller interface to poll and process for task
 	taskPoller interface {
-		PollAndProcessSingleTask() error
+		// PollTask polls for one new task
+		PollTask() (interface{}, error)
```
madhuravi: Can we avoid empty interfaces here? It is much slower and inefficient.

internal_worker.go
```diff
@@ -97,9 +103,11 @@ type (
...
 		ConcurrentPollRoutineSize int
 
-		// Defines how many executions for task list by this worker.
-		// TODO: In future we want to separate the activity executions as they take longer than polls.
-		// ConcurrentExecutionRoutineSize int
+		// Defines how many concurrent executions for task list by this worker.
+		ConcurrentActivityExecutionSize int
+
+		// Defines rate limiting on number of activity tasks that can be executed per second.
+		MaxActivityExecutionRatePerSecond int
```
madhuravi: nit: Should either be MaxActivityExecutionsPerSecond or MaxActivityExecutionRate to be meaningful.

internal_task_pollers.go
yiminc: Any suggestion are welcome :)

internal_worker.go
yiminc: it was MaxActivityExecutionRate. too many chefs....

yiminc: I will revert back to MaxActivityExecutionRate

internal_worker_base.go
```diff
@@ -121,18 +135,137 @@ func (bw *baseWorker) Start() {
...
 	}
 
-	// Launch the routines to do work
-	for i := 0; i < bw.options.routineCount; i++ {
+	for i := 0; i < bw.options.pollerCount; i++ {
 		bw.shutdownWG.Add(1)
-		go bw.execute(i)
+		go bw.runPoller()
 	}
+	bw.shutdownWG.Add(1)
```
madhuravi: This seems wrong? It should be in the loop to match poller count.

internal_worker_base.go
```diff
@@ -121,18 +135,137 @@ func (bw *baseWorker) Start() {
...
+			return
+		case task := <-bw.taskQueueCh:
+			// got new task, wait for taskSemaphore or shutdown
+			select {
+			case <-bw.shutdownCh:
+				return
+			case <-bw.taskSemaphore:
+				// block until taskRateLimiter satisfied
+				for !bw.taskRateLimiter.Consume(1, time.Millisecond) {
+					if bw.isShutdown() {
```
madhuravi: Not sure what's happening here, why do we need so many checks on the shutdown channel?

madhuravi: Why not just make this an if statement and pass to the outer for loop if not satisfied?

if bw.taskRateLimiter.Consume(1, time.Millisecond {
  go bw.ProcessTask(task)
}

No isShutdown check necessary.

internal_worker_base.go
```diff
@@ -121,18 +135,137 @@ func (bw *baseWorker) Start() {
...
+						return
+					}
+				}
+				go bw.processTask(task)
```
madhuravi: Shouldn't we make sure this goroutine completed?

worker.go
```diff
@@ -46,14 +46,15 @@ type (
 	WorkerOptions struct {
 		// Optional: To set the maximum concurrent activity executions this host can have.
 		// The zero value of this uses the default value.
```
madhuravi: minor: IMO this line is confusing and we can remove.

internal_worker_base.go
```diff
@@ -121,18 +135,137 @@ func (bw *baseWorker) Start() {
...
+	if task != nil {
+		bw.taskQueueCh <- task
+	} else {
+		bw.pollerRequestCh <- struct{}{} // poll failed, trigger a new pool
+	}
+}
+
+func (bw *baseWorker) processTask(task interface{}) {
+	pollerRequested := false
+	if bw.concurrentTaskCount.Inc() < int32(bw.options.maxConcurrentTask) {
```
venkat1109: You don't need this if you make the cap(pollRequestCh) same as worker count. This logic is confusing and it also cannot keep the workers busy, see example below:

pollerCount=1, workerCount=2

1. worker1 gets task1
2. worker1 finds that TaskCount.Inc() < max, so creates another pollToken for poller
3. worker2 gets task2
4. worker2 finds that TaskCount.Inc()  > max, so, won't enqueue another pollToken 
```5. worker1 completes, but it's idle now until worker2 completes and puts a token back```

File: internal_worker_base.go (hunk 10 above)
Comment (madhuravi) [inline internal_worker_base.go]: Never mind, that won't work since we lose the task.

File: internal_worker_base.go (hunk 1 above)
Comment (venkat1109) [inline internal_worker_base.go]: I think starving pollers is fine, pollers need to do work only when workers are idle. And in the common case _pollLatency << processTaskLatency_, so I wouldn't worry about pre-fetching until we have data or a use-case that proves our assumption is wrong.  

To start with, we should shoot for correctness i.e never timeout tasks due to prefetching regardless of your worker configuration. We can always add prefetching support later, when we need it.

And even when we do prefetching, we probably will do it *smart* based on some smoothed out throughput estimation, just so that we can avoid unnecessary task timeouts.

File: internal_worker.go (hunk 8 above)
Comment (venkat1109) [inline internal_worker.go]: cool.. my rule on this one is - any config that's based off time should have units as the suffix or it won't make sense to the reader :)

File: internal_worker_base.go [hunk 14]
```diff
@@ -121,18 +135,137 @@ func (bw *baseWorker) Start() {
...
+}
+
+func (bw *baseWorker) runTaskDispatcher() {
+	defer bw.shutdownWG.Done()
+
+	for i := 0; i < bw.options.maxConcurrentTask; i++ {
+		bw.taskSemaphore <- struct{}{}
+	}
+
+	for {
```
madhuravi: A simpler alternative:
for {
  go func() {
    task := <-bw.taskQueueCh
    <-bw.taskSemaphore
    for !bw.taskRateLimiter.Consume(1, time.Millisecond) {}
    processTask(task)
   done <- struct{}{}
  }()
  select {
    case <- shutdownCh:
      return
    case <- done:
  }
}

internal_worker_base.go
yiminc: this is for the dispatcher, we have one for each poller separately.

internal_worker_base.go
yiminc: not sure what you mean? it will eventually complete, right?

internal_worker_base.go
madhuravi: Ignore ^^, I see it inside the loop as well.

internal_worker_base.go
madhuravi: I see you're achieving this with taskSemaphore.

internal_worker_base.go
```diff
@@ -121,18 +135,137 @@ func (bw *baseWorker) Start() {
...
+			}
+		}
+	}
+}
+
+func (bw *baseWorker) runTaskDispatcher() {
+	defer bw.shutdownWG.Done()
+
+	for i := 0; i < bw.options.maxConcurrentTask; i++ {
+		bw.taskSemaphore <- struct{}{}
```
madhuravi: Can we get rid of taskSemaphore by making taskQueue a buffered channel of size maxConcurrentTask?
See https://gobyexample.com/worker-pools for an example

internal_worker_base.go
yiminc: if pollLatency=5ms, two poller is able to produce 400 task per second, which will keep all worker busy. The dispatcher is only receiving from the task queue, it is not the worker. Once a worker starts, it checks if the concurrent task count is less than max, and if so, it sends poller request before process the task.

internal_worker_base.go
yiminc: 5 is not true, after worker 1 completes, it will put a request (see end of this method), we don't wait for worker 2, and that is desired behavior because before worker 1 is completed, we don't want poll. The problem of use cap of worker count and pre fill pollRequestCh has the problem that i described earlier that it will result refetch in case of large maxConcurrentTask.

internal_worker_base.go
yiminc: with this, the dispatcher will be blocked while the processing the task. this will reduce the throughput to as if there is only a single worker.

internal_worker_base.go
venkat1109: > 5 is not true, after worker 1 completes, it will put a request (see end of this method)
I am confused, this is the end of the function

```
if !pollerRequested { --> won't this be false ? 
}
```

yiminc: I confused myself. Yes, you are correct in this case, the worker 1 will idle until worker 2 complete and put up new poll request. ideally, when worker 1 finishes, it should be smart enough to put a request. but in this case, it was actually even smarter to put up a poll request even before processing task. The problem is that we only have one poll request initially (same as pollerCount). if we make the initial poll request count same as maxConcurrentTask, then we have the problem i described in the PR for large maxConcurrentTask.

internal_task_pollers.go
mfateev: This is relatively low rate method. So interface is not going to affect perf.

internal_worker.go
```diff
@@ -40,9 +40,15 @@ import (
...
-	defaultMaxConcurrentActivityExecutionSize = 10000  // Large execution size(unlimited)
-	defaultMaxActivityExecutionRate           = 100000 // Large execution rate(100K per sec)
+	// Set to 1 for now, can adjust later if needed. The typical RTT (round-trip time) is below 1ms within data center.
+	// With 1 poller, we could easily achieve a few hundred RPS.
+	defaultConcurrentPollRoutineSize = 1
+
+	defaultMaxConcurrentActivityExecutionSize = 1000   // Large concurrent activity execution size (1k)
+	defaultMaxActivityExecutionRate           = 100000 // Large activity execution rate (unlimited)
+
+	defaultMaxConcurrentWorkflowExecutionSize = 10     // workflow code normally run fast, so only need small concurrent number
```
mfateev: It is not the case when pagination is involved

worker.go
```diff
@@ -46,14 +46,15 @@ type (
...
+		// default: defaultMaxConcurrentActivityExecutionSize(1k)
 		MaxConcurrentActivityExecutionSize int
 
 		// Optional: Sets the rate limiting on number of activities that can be executed per second.
 		// This can be used to protect down stream services from flooding.
 		// The zero value of this uses the default value.
 		// default: defaultMaxActivityExecutionRate(100k)
-		MaxActivityExecutionRate float32
+		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to rate limiting.
+		MaxActivityExecutionsPerSecond int
```
mfateev: It was float to support rates like 1 activity in 10 seconds. Sometimes customers need to rate limit below one request per second.

internal_worker.go
yiminc: true. May be set this to 100? This is a value currently not exposed to user, and is hard coded here.

worker.go
yiminc: I change this because our current RateLimiter implementation does not support that. Maybe we should change our RateLimiter implementation and keep this as float.

internal_worker_base.go
yiminc: Alternative solution: set size of pollerRequestCh to be same as maxConcurrentTask, and initially fill pollerRequestCh so we have plenty of poll request. Then we set size of taskQueueCh to no buffer, so that only after a task is dispatched can the poller put new task. That means, the max tasks blocked by rate limiter will be the number of pollers. The pros is both poller and worker will keep busy.

venkat1109: I think that's an acceptable tradeoff. Another option would be to have worker first block on rate limit, then grant a poller token, with the assumption that our pollTask() latencies will be negligible. I am fine with either of the approaches, but lets not compromise on keeping the workers busy part :)

internal_worker_base.go
```diff
@@ -121,18 +136,119 @@ func (bw *baseWorker) Start() {
...
+	}
+
+	for {
+		// wait for new task or shutdown
+		select {
+		case <-bw.shutdownCh:
+			return
+		case task := <-bw.taskQueueCh:
+			// block until taskRateLimiter satisfied
+			for !bw.taskRateLimiter.Consume(1, time.Millisecond) {
```
venkat1109: can we increase the timeout to 100ms instead of 1ms ? This will avoid spinning in a loop, if we are contending for tokens, while at the same time allows us to shutdown with 100ms.

internal_worker.go
```diff
@@ -748,7 +760,9 @@ func (aw *aggregatedWorker) Stop() {
 	aw.logger.Info("Stopped Worker")
 }
 
-// aggregatedWorker returns an instance to manage the workers.
+// aggregatedWorker returns an instance to manage the workers. Use defaultConcurrentPollRoutineSize (which is 1) as
+// poller size. The typical RTT (round-trip time) is below 1ms within data center. With 1 poller, we could easily achieve
```
venkat1109: more interesting is the median poll API latency, which based on our dashboards is ~5-6ms..

PR #189, opened 2017-07-22
Allow custom names for register and execute workflows and activities

madhuravi:


activity.go
```diff
@@ -42,9 +42,14 @@ type (
 		ActivityID        string
 		ActivityType      ActivityType
 	}
+
+	// RegisterActivityOptions consists of options for registering an activity
+	RegisterActivityOptions struct {
+		Name string
```
mfateev: Add comment if it is required and what is the default value

internal_worker.go
```diff
@@ -348,43 +348,59 @@ func (th *hostEnvImpl) AddActivityRegistrationInterceptor(i interceptorFn) {
...
 	fnType := reflect.TypeOf(af)
 	if err := validateFnFormat(fnType, true); err != nil {
 		return err
 	}
-	// Check if already registered
-	fnName := getFunctionName(af)
-	if _, ok := th.getWorkflowFn(fnName); ok {
-		return fmt.Errorf("workflow type \"%v\" is already registered", fnName)
+	names := make([]string, 0, 2)
+	names = append(names, getFunctionName(af))
```
mfateev: I think when name is present in the options there is no need to register workflow under fn name.
If somebody really needs (for example for migration) they can just call Register twice with the same function.

internal_worker.go
```diff
@@ -348,43 +348,59 @@ func (th *hostEnvImpl) AddActivityRegistrationInterceptor(i interceptorFn) {
...
 	fnType := reflect.TypeOf(af)
 	if err := validateFnFormat(fnType, false); err != nil {
 		return err
 	}
-	// Check if already registered
-	fnName := getFunctionName(af)
-	if _, ok := th.getActivityFn(fnName); ok {
-		return fmt.Errorf("activity type \"%v\" is already registered", fnName)
+	names := make([]string, 0, 2)
+	names = append(names, getFunctionName(af))
```
mfateev: Ditto

activity.go
```diff
@@ -56,8 +61,14 @@ type (
 // Serialization of all primitive types, structures is supported ... except channels, functions, variadic, unsafe pointer.
 // This method calls panic if activityFunc doesn't comply with the expected format.
 func RegisterActivity(activityFunc interface{}) {
+	RegisterActivityWithOptions(activityFunc, RegisterActivityOptions{})
+}
+
+// RegisterActivityWithOptions registers an activity function with the framework with provided
+// options.
+func RegisterActivityWithOptions(activityFunc interface{}, options RegisterActivityOptions) {
```
molteanu: Should this method that allows us to register with a name be on the Worker instead of global? Wasn't the concern with "custom names" that there could much easier be conflicts if we allow global registration?

madhuravi: Was planning on moving the register functions under worker in a separate PR.

internal_worker.go
madhuravi: Based on offline discussion, this behavior will stay. Register will implicitly add the fully qualified function name and user will not need to call Register twice.

activity.go
madhuravi: Decided to add it here, latest commit has that change.

activity.go
```diff
@@ -42,26 +42,12 @@ type (
 		ActivityID        string
 		ActivityType      ActivityType
 	}
-)
 
-// RegisterActivity - register a activity function with the framework.
-// A activity takes a context and input and returns a (result, error) or just error.
-// Examples:
-//	func sampleActivity(ctx context.Context, input []byte) (result []byte, err error)
```
sivakku: Can we preserve these comments in the workflow client? they can look at ExecuteActivity, but probably better to have in two places.

internal_public.go
```diff
@@ -148,27 +162,27 @@ func NewActivityTaskHandler(service m.TChanWorkflowService, identity string, log
 func AddWorkflowRegistrationInterceptor(
```
sivakku: Should these workflow/activity interceptors be moved to per worker? instead of global, since they are bound to registration scope.

internal_worker.go
```diff
@@ -799,73 +844,49 @@ func newAggregatedWorker(
...
+		if testTags != nil && len(testTags) > 0 {
+			workflowWorker = newWorkflowWorkerWithPressurePoints(
+				service,
+				domain,
+				workerParams,
+				testTags,
+				env,
+			)
 		} else {
-			logger.Warn("Workflow worker is enabled but no workflow is registered. Use cadence.RegisterWorkflow() to register your workflow.")
```
sivakku: Does this mean we can register workflows after it is started?

internal_workflow.go
```diff
@@ -922,7 +923,11 @@ func newWorkflowDefinition(workflow workflow) workflowDefinition {
 	return &syncWorkflowDefinition{workflow: workflow}
 }
 
-func getValidatedWorkerFunction(workflowFunc interface{}, args []interface{}) (*WorkflowType, []byte, error) {
+func getValidatedWorkerFunction(
```
yiminc: does it make sense to make it a method on hostEnvImpl struct?

internal_workflow.go
```diff
@@ -1032,7 +1040,7 @@ func (d *decodeFutureImpl) Get(ctx Context, value interface{}) error {
 		return errors.New("value parameter is not a pointer")
 	}
 
-	err := deSerializeFunctionResult(d.fn, d.futureImpl.value.([]byte), value)
+	err := deSerializeFunctionResult(d.fn, d.futureImpl.value.([]byte), value, d.env)
```
yiminc: and this method as well, as a method of the host env?

internal_workflow_testsuite.go
```diff
@@ -276,7 +277,8 @@ func (env *testWorkflowEnvironmentImpl) executeWorkflow(workflowFn interface{},
 		// auto register workflow if it is not already registered
 		fnName := getFunctionName(workflowFn)
 		if _, ok := s.hostEnv.getWorkflowFn(fnName); !ok {
-			s.RegisterWorkflow(workflowFn)
+			wo := RegisterWorkflowOptions{}
+			s.RegisterWorkflow(workflowFn, wo)
```
yiminc: would be nice if we could pass in nil

internal_worker.go
yiminc: also needs to update the message as it is refer to cadence.Register()

activity.go
madhuravi: Yes, done.

internal_public.go
madhuravi: I have retained all functions in internal_public separately since they were to be used for catalyst. When the catalyst story is solidified, we can decide to remove this potentially.

On the worker, this is exposed as part of the hostEnv. Does this need to be exposed to users? What is the use case?

internal_worker.go
madhuravi: This means we can register workflows after object creation. Since Register* functions are on the object itself, we don't know yet what the workflows/activities are.

However I am planning to add a check so Register panics if called after worker.Start. That is the behavior we want to prevent.

internal_workflow_testsuite.go
madhuravi: I agree with your underlying point, it's nice to not have users pass in an empty struct if they don't care about the names. I think it is good practice to avoid nils though. An option we discussed is exposing two methods
RegisterWorkflowWithOptions(fn, opts)
RegisterWorkflow(fn) 

However this would mean these two sets of methods on the worker and client interfaces. We can discuss this if you think that's a good idea.

internal_workflow.go
madhuravi: Thought about this as well, makes sense.

internal_workflow.go
madhuravi: This probably makes more sense as a method on decodeFutureImpl.

madhuravi: Just looked at the usage. You're right, moved to host env.

internal_public.go
sivakku: This is only for catalyst, makes sense to revisit after the story is clear.

internal_workflow_client.go
```diff
@@ -63,6 +64,16 @@ type (
 	}
 )
 
+// RegisterWorkflow registers the workflow function with options
```
sivakku: minor: add comments about the parameters as well like rest of methods in wcclient. Mostly user lands on this class.

madhuravi:
@yiminc Yes you are right, this is something I had in mind already. Planning to separate encoding out later. Decided not to bundle it with this one since it is already very extensive.

PR #192, opened 2017-07-25
update rate limiter

yiminc:
update rate limiter to support RPS less than 1.


internal_worker.go
```diff
@@ -106,8 +107,11 @@ type (
 		// Defines how many concurrent executions for task list by this worker.
 		ConcurrentActivityExecutionSize int
 
-		// Defines rate limiting on number of activity tasks that can be executed per second.
-		MaxActivityExecutionsPerSecond float32
+		// Defines rate limiting on number of activity tasks that can be executed per MaxActivityExecutionRateRefreshDuration.
+		MaxActivityExecutionRate int
```
venkat1109: Since you are anyways renaming this, can you remove *Activity* from  the name. Makes the reader think this applies only to activities, when in fact, this is really the taskExecutionRate

internal_worker_base.go
```diff
@@ -214,7 +218,8 @@ func (bw *baseWorker) pollTask() {
 	var err error
 	var task interface{}
 	bw.retrier.Throttle()
-	if bw.pollRateLimiter.Consume(1, time.Millisecond*100) {
+	if bw.pollLimiter.Wait(bw.limiterContext) == nil {
```
venkat1109: this changes behavior a bit. Previously, it looks like we waited for 100ms and then initiated backoff. Now we wait infinitely. This is fine, except now, the backoff logic is useless. It also doesn't look like backoff belongs here, so you might as well remove it.

internal_worker_base.go
```diff
@@ -214,7 +218,8 @@ func (bw *baseWorker) pollTask() {
...
+		//if bw.pollRateLimiter.Consume(1, time.Millisecond*100) {
```
venkat1109: remove this

worker.go
```diff
@@ -49,12 +50,18 @@ type (
 		// default: defaultMaxConcurrentActivityExecutionSize(1k)
 		MaxConcurrentActivityExecutionSize int
 
-		// Optional: Sets the rate limiting on number of activities that can be executed per second.
+		// Optional: Sets the rate limiting on number of activities that can be executed per refresh duration.
 		// This can be used to protect down stream services from flooding.
 		// The zero value of this uses the default value.
 		// default: defaultMaxActivityExecutionRate(100k)
 		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to rate limiting.
-		MaxActivityExecutionsPerSecond float32
```
sivakku: "perSecond" is more readable than "rate", event for fractions.

worker.go
```diff
@@ -49,12 +50,18 @@ type (
...
+		MaxActivityExecutionRate int
+
+		// Optional: Sets the refresh duration for rate limit. If not specified, it uses 1s as default.
+		// Use this to fine tune your rate limiter. For example, you could set this duration to 10s with max rate set to
+		// 1 which means we rate limit the activity task to at most 1 per every 10s. You could also set this duration to
+		// 100ms with max rate set wo 10 which means rate limit to 10 activity task execution per every 100ms.
+		MaxActivityExecutionRateRefreshDuration time.Duration
```
sivakku: Rate exposing through two variables is confusing on API. Why can't we have just one that can take 0.1 instead of 1 in 10sec.

internal_worker.go
sivakku: For decisions we should figure out a high number that we manage by ourself, which might not be same as this and let users worry about only activity rate.

yiminc: This is only apply to activity task. For workflow task, we probably don't need to expose to user, for now we use hardcoded rate at 1000 per second.

internal_worker_base.go
yiminc: oops

worker.go
yiminc: we don't use float (fraction), it is not per 'second' anymore. It is per refresh duration. You can set rate to 1 and duration to 10s which means 0.1 RPS.

worker.go
yiminc: good point. however, golang's rate limiter is actually using 2 concept for the rate and duration.

yiminc: using 2 variables of rate and refill duration gives user more precise control of the behavior, and clear behavior expectation. It is unclear what it means for 1.1 RPS if we use only one float variable for rate.

internal_worker_base.go
venkat1109: ah.. the diff viewer made me think the else{} was for the if {rateLimiter.Wait()} - nvm

venkat1109: yes, the backoff on failure makes sense.. but backoff on ratelimit doesn't. Your Wait() will *never* fail, so else is redundant.

yiminc: did not realize the else case for rate limiter, will remove

PR #196, opened 2017-08-01
validate domain existence at worker start time

venkat1109:
This patch adds validation for domain existence at worker start time. If the server throws either one of EntityNotExist (or) BadRequest errors, then worker.Start() will fail immediately. If the server throws any other transient error, we ignore the error and continue.

### DIFF
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -37,6 +37,7 @@ import (
 	"github.com/uber-go/tally"
 	m "go.uber.org/cadence/.gen/go/cadence"
 	"go.uber.org/cadence/.gen/go/shared"
+	"go.uber.org/cadence/common/backoff"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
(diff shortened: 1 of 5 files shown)


internal_worker.go
```diff
@@ -160,6 +161,31 @@ func ensureRequiredParams(params *workerExecutionParameters) {
...
+		if _, ok := err.(*shared.EntityNotExistsError); ok {
+			logger.Error("domain does not exist", zap.String("domain", domain), zap.Error(err))
+			return err
+		}
+		if _, ok := err.(*shared.BadRequestError); ok {
+			logger.Error("domain does not exist", zap.String("domain", domain), zap.Error(err))
+			return err
+		}
+		// on any other error, just return true
+		logger.Warn("unable to verify if domain exist", zap.String("domain", domain), zap.Error(err))
```
yiminc: should we retry on transient error?

internal_worker.go
```diff
@@ -160,6 +161,31 @@ func ensureRequiredParams(params *workerExecutionParameters) {
...
+	if err != nil {
+		if _, ok := err.(*shared.EntityNotExistsError); ok {
+			logger.Error("domain does not exist", zap.String("domain", domain), zap.Error(err))
+			return err
+		}
+		if _, ok := err.(*shared.BadRequestError); ok {
+			logger.Error("domain does not exist", zap.String("domain", domain), zap.Error(err))
+			return err
+		}
+		// on any other error, just return true
```
madhuravi: why? shouldnt this be an error?

venkat1109: the assumption is - all other errors are transient. That's the server side contract. The server, for the most part, also wrap all these other errors within InternalServiceError{}. In the case of a transient error, there is not much we can do. Following the idl contract:
```
  shared.DescribeDomainResponse DescribeDomain(1: shared.DescribeDomainRequest describeRequest)
    throws (
      1: shared.BadRequestError badRequestError,
      2: shared.InternalServiceError internalServiceError,
      3: shared.EntityNotExistsError entityNotExistError,
    )
```

internal_worker.go
venkat1109: newTChannelContext() comes with a defaultRetryOption - which will retry on transient connect errors. As for InternalServiceErrors, yes, I can add a retry, but the expiry will be < 10 secs. We cannot block worker.Start() for a long time.

venkat1109: Added support for backoff/retry

PR #197, opened 2017-08-02
add client side metrics

yiminc:
add client side metrics.

### DIFF
--- a/common/metrics/constants.go
+++ b/common/metrics/constants.go
@@ -22,19 +22,36 @@ package metrics
 
 // Workflow Creation metrics
 const (
-	WorkflowsStartTotalCounter      = "workflows-start-total"
-	ActivitiesTotalCounter          = "activities-total"
-	DecisionsTotalCounter           = "decisions-total"
-	DecisionsTimeoutCounter         = "decisions-timeout"
-	WorkflowsCompletionTotalCounter = "workflows-completion-total"
-	WorkflowEndToEndLatency         = "workflows-endtoend-latency"
-	ActivityEndToEndLatency         = "activities-endtoend-latency"
-	DecisionsEndToEndLatency        = "decisions-endtoend-latency"
-	ActivityPollLatency             = "activities-poll-latency"
-	DecisionsPollLatency            = "decisions-poll-latency"
-	ActivityExecutionLatency        = "activities-execution-latency"
-	DecisionsExecutionLatency       = "decisions-execution-latency"
-	ActivityResponseLatency         = "activities-response-latency"
-	DecisionsResponseLatency        = "decisions-response-latency"
-	UnhandledSignalsTotalCounter    = "unhandled-signals-total"
+	WorkflowStartCounter     = "workflow-start"
+	WorkflowCompletedCounter = "workflow-completed"
+	WorkflowEndToEndLatency  = "workflow-endtoend-latency" // measure workflow execution from start to close
+	DecisionTimeoutCounter   = "decision-timeout"
+
+	DecisionPollCounter            = "decision-poll-total"
+	DecisionPollFailedCounter      = "decision-poll-failed"
+	DecisionPollNoTaskCounter      = "decision-poll-no-task"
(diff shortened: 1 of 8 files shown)


internal_task_pollers.go
```diff
@@ -186,21 +174,20 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
 		}, serviceOperationRetryPolicy, isServiceTransientError)
 
 	if err != nil {
+		wtp.metricsScope.Counter(metrics.DecisionResponseFailedCounter).Inc(1)
 		return err
 	}
-	return nil
 
+	wtp.metricsScope.Timer(metrics.DecisionResponseLatency).Record(time.Now().Sub(responseStartTime))
```
sivakku: This would include retries sleep as well. should we moved this to inside RespondDecisionCompelted method?

internal_workflow.go
```diff
@@ -410,33 +411,26 @@ func getDispatcher(ctx Context) dispatcher {
...
-	}
-
+	env := getWorkflowEnvironment(ctx)
 	panicErr := dispatcher.ExecuteUntilAllBlocked()
 	if panicErr != nil {
-		env := getWorkflowEnvironment(ctx)
-		env.GetLogger().Error("Dispatcher panic.",
-			zap.String("PanicError", panicErr.Error()),
-			zap.String("PanicStack", panicErr.StackTrace()))
-		checkUnhandledSigFn(ctx)
```
sivakku: it is probably valuable to have warning about unhandled signals, when it panic'ing.

yiminc: They are moved to task handler layer. Because the env.GetLogger() will not log in replay, but we want to log anyway for all panic.

internal_task_pollers.go
yiminc: I think it still make sense to include that time as part of decision response latency because any retry indicate there are issues with responding.

common/metrics/constants.go
```diff
@@ -22,19 +22,36 @@ package metrics
...
+	ActivityPollNoTaskCounter      = "activity-poll-no-task"
+	ActivityPollSucceedCounter     = "activity-poll-succeed"
+	ActivityPollLatency            = "activity-poll-latency"
+	ActivityExecutionFailedCounter = "activity-execution-failed"
+	ActivityExecutionLatency       = "activity-execution-latency"
+	ActivityResponseLatency        = "activity-response-latency"
+	ActivityResponseFailedCounter  = "activity-response-failed"
+	ActivityEndToEndLatency        = "activity-endtoend-latency"
+	ActivityTaskPanicCounter       = "activity-task-panic"
+	ActivityTaskCompletedCounter   = "activity-task-completed"
```
yiminc: Will add failed counter. But won't be able to add timeout as that is determined by server.

common/metrics/constants.go
```diff
@@ -22,19 +22,36 @@ package metrics
...
+	DecisionPollNoTaskCounter      = "decision-poll-no-task"
+	DecisionPollSucceedCounter     = "decision-poll-succeed"
+	DecisionPollLatency            = "decision-poll-latency" // measure succeed poll request latency
+	DecisionExecutionFailedCounter = "decision-execution-failed"
+	DecisionExecutionLatency       = "decision-execution-latency"
+	DecisionResponseFailedCounter  = "decision-response-failed"
+	DecisionResponseLatency        = "decision-response-latency"
+	DecisionEndToEndLatency        = "decision-endtoend-latency" // measure from poll request start to response completed
+	DecisionTaskPanicCounter       = "decision-task-panic"
+	DecisionTaskCompletedCounter   = "decision-task-completed"
```
yiminc: will add latency/count for get history.

internal_workflow.go
yiminc: ok, on a second thought, i don't think it is meaningful to warn on unhandled signals when panic. as the panic is most likely the reason that there is unhandled signals.

PR #200, opened 2017-08-03
add service wrapper that emit metrics automatically

yiminc:
add tchannel service wrapper that automatically emit metrics on all service calls.

This will emit metrics like:
PollForActivityTask.cadence-request
RespondDecisionTaskCompleted.cadence-latency
....

### DIFF
--- a/common/metrics/service_wrapper.go
+++ b/common/metrics/service_wrapper.go
@@ -37,6 +37,27 @@ type workflowServiceMetricsWrapper struct {
 	mutex       sync.Mutex
 }
 
+const (
+	scopeNameDeprecateDomain                = "DeprecateDomain"
+	scopeNameDescribeDomain                 = "DescribeDomain"
+	scopeNameGetWorkflowExecutionHistory    = "GetWorkflowExecutionHistory"
+	scopeNameListClosedWorkflowExecutions   = "ListClosedWorkflowExecutions"
+	scopeNameListOpenWorkflowExecutions     = "ListOpenWorkflowExecutions"
+	scopeNamePollForActivityTask            = "PollForActivityTask"
+	scopeNamePollForDecisionTask            = "PollForDecisionTask"
+	scopeNameRecordActivityTaskHeartbeat    = "RecordActivityTaskHeartbeat"
+	scopeNameRegisterDomain                 = "RegisterDomain"
+	scopeNameRequestCancelWorkflowExecution = "RequestCancelWorkflowExecution"
+	scopeNameRespondActivityTaskCanceled    = "RespondActivityTaskCanceled"
+	scopeNameRespondActivityTaskCompleted   = "RespondActivityTaskCompleted"
+	scopeNameRespondActivityTaskFailed      = "RespondActivityTaskFailed"
+	scopeNameRespondDecisionTaskCompleted   = "RespondDecisionTaskCompleted"
+	scopeNameSignalWorkflowExecution        = "SignalWorkflowExecution"
+	scopeNameStartWorkflowExecution         = "StartWorkflowExecution"
+	scopeNameTerminateWorkflowExecution     = "TerminateWorkflowExecution"
+	scopeNameUpdateDomain                   = "UpdateDomain"
+)
+
 // NewWorkflowServiceWrapper creates a new wrapper to WorkflowService that will emit metrics for each service call.
 func NewWorkflowServiceWrapper(service m.TChanWorkflowService, scope tally.Scope) m.TChanWorkflowService {
 	return &workflowServiceMetricsWrapper{service: service, scope: scope, childScopes: make(map[string]tally.Scope)}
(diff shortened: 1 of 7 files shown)


common/metrics/service_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+		return scope
+	}
+	scope = w.scope.SubScope(scopeName)
+	w.childScopes[scopeName] = scope
+	w.mutex.Unlock()
+	return scope
+}
+
+func (w *workflowServiceMetricsWrapper) DeprecateDomain(ctx thrift.Context, request *shared.DeprecateDomainRequest) error {
+	scope := w.getScope("DeprecateDomain")
```
sivakku: Can we  make const out of scope name for all these strings easier to look what all being reported.

common/metrics/service_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+	scope.Counter(CadenceRequest).Inc(1)
+	startTime := time.Now()
+	err := w.service.DeprecateDomain(ctx, request)
+	scope.Timer(CadenceLatency).Record(time.Now().Sub(startTime))
+	if err != nil {
```
sivakku: We should not count error if the error is EntityNotExistsError, that is a valid response.

common/metrics/service_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	scope = w.scope.SubScope(scopeName)
+	w.childScopes[scopeName] = scope
+	w.mutex.Unlock()
+	return scope
+}
+
+func (w *workflowServiceMetricsWrapper) DeprecateDomain(ctx thrift.Context, request *shared.DeprecateDomainRequest) error {
+	scope := w.getScope("DeprecateDomain")
+	scope.Counter(CadenceRequest).Inc(1)
```
sivakku: May be wrap these three things in a function that takes scope name  may be a function to execute that wraps cadence call.

common/metrics/service_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+	return err
+}
+
+func (w *workflowServiceMetricsWrapper) DescribeDomain(ctx thrift.Context, request *shared.DescribeDomainRequest) (*shared.DescribeDomainResponse, error) {
+	scope := w.getScope("DescribeDomain")
+	scope.Counter(CadenceRequest).Inc(1)
+	startTime := time.Now()
+	result, err := w.service.DescribeDomain(ctx, request)
+	scope.Timer(CadenceLatency).Record(time.Now().Sub(startTime))
+	if err != nil {
```
sivakku: same here for: EntityNotExistsError

common/metrics/service_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+	return result, err
+}
+
+func (w *workflowServiceMetricsWrapper) GetWorkflowExecutionHistory(ctx thrift.Context, request *shared.GetWorkflowExecutionHistoryRequest) (*shared.GetWorkflowExecutionHistoryResponse, error) {
+	scope := w.getScope("GetWorkflowExecutionHistory")
+	scope.Counter(CadenceRequest).Inc(1)
+	startTime := time.Now()
+	result, err := w.service.GetWorkflowExecutionHistory(ctx, request)
+	scope.Timer(CadenceLatency).Record(time.Now().Sub(startTime))
+	if err != nil {
```
sivakku: same here: for workflow not found.

common/metrics/service_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+	return err
+}
+
+func (w *workflowServiceMetricsWrapper) RespondActivityTaskCompleted(ctx thrift.Context, request *shared.RespondActivityTaskCompletedRequest) error {
+	scope := w.getScope("RespondActivityTaskCompleted")
+	scope.Counter(CadenceRequest).Inc(1)
+	startTime := time.Now()
+	err := w.service.RespondActivityTaskCompleted(ctx, request)
+	scope.Timer(CadenceLatency).Record(time.Now().Sub(startTime))
+	if err != nil {
```
sivakku: same here for: ActivityNotExist for all above activity methods.

common/metrics/service_wrapper.go
yiminc: I tried that and the result does not look better than this.

common/metrics/service_wrapper.go
yiminc: Why is EntityNotExistsError count as succeed? It indicate something wrong with the request.

common/metrics/service_wrapper.go
yiminc: same comments as above. This is not a succeed case.

yiminc:
@sivakku , for this wrapper the main idea is we want to capture a generic info regarding every service call. This is handled in the same way as it is on server side (metricClient).

common/metrics/service_wrapper.go
sivakku: These counters indicate server failures to respond, but there could be valid response that the domain doesn't exist and the user is retrying crazily on it and that shouldn't count against the failures of the calls. we do this on server side as well, take a look at some of those front-end handler metrics. These failures can cause clients to contact us, but those are valid.

yiminc: Those metrics are meant to be used as "tracing level" metrics for us (cadence team). Our users are recommended to monitor based on the higher level metrics that we logged.

PR #202, opened 2017-08-04
Client library to use Thrift serializer for Thrift types

sivakku:
The user inputs/error messages for activity/workflow function defaults to using GOB encoding. We want to use thrift serialization for thrift specific structures, so some of these activities can be cross platform. This is a step towards that, the worker checks if the provided types are thrift types then it would automatically use thrift serialization instead of GOB.


common/thrift_util.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// TListSerialize is used to serialize list of thrift TStruct to []byte
+func TListSerialize(ts []thrift.TStruct) (b []byte, err error) {
+	if ts == nil {
+		return
+	}
+
+	t := thrift.NewTSerializer()
+	t.Transport.Reset()
```
yiminc: is this reset necessary on a newly created instance?

common/thrift_util.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if err = t.Protocol.Flush(); err != nil {
+		return
+	}
+
+	if err = t.Transport.Flush(); err != nil {
+		return
+	}
+
+	b = append(b, t.Transport.Bytes()...)
```
yiminc: can this be b = t.Transport.Bytes()

internal_worker.go
```diff
@@ -546,20 +554,59 @@ func (th *hostEnvImpl) registerEncodingTypes(fnType reflect.Type) error {
...
+		}
+		rType := reflect.Indirect(val).Type()
+		return th.registerType(rType, encoder)
+	}
+	return nil
 }
 
 // register type with our encoder.
-func (th *hostEnvImpl) registerType(t reflect.Type) error {
+func (th *hostEnvImpl) registerType(t reflect.Type, encoder encoding) error {
```
yiminc: should this be a method on encoding instead of on hostEnvImpl?

common/thrift_util.go
sivakku: it might be redundant, i looked at how thrift is doing at it and seems to be the recurring pattern else where, thought this is a better safety check.
https://github.com/apache/thrift/blob/master/lib/go/thrift/serializer.go#L58

common/thrift_util.go
sivakku: Again i used the pattern from https://github.com/apache/thrift/blob/master/lib/go/thrift/serializer.go#L58. the append makes sense if the buffer is getting stitched.

internal_worker.go
sivakku: The encoder interface was kept generic taking interface{} instead of reflection types. some of this is specific to GO, i would leave the encoding interface as it is, since registration in some serialization in advance might not make sense.

common/thrift_util.go
sivakku: i did realize, the append would make a copy of the slice, instead of ptr to slice, that might be better to avoid re-using buffers.

yiminc: I think it is not suitable in this case. The append confuses people who read this code, it is also not as performant as it makes unnecessary copies.

common/thrift_util.go
yiminc: sg

internal_worker.go
yiminc: sg.

PR #205, opened 2017-08-07
Add registration functions to support custom names for workflows/activities

madhuravi:


internal_worker.go
```diff
@@ -925,50 +1029,41 @@ func newAggregatedWorker(
...
+		if testTags != nil && len(testTags) > 0 {
+			workflowWorker = newWorkflowWorkerWithPressurePoints(
+				service,
+				domain,
+				workerParams,
+				testTags,
+				hostEnv,
+			)
 		} else {
-			logger.Warn("Workflow worker is enabled but no workflow is registered. Use cadence.RegisterWorkflow() to register your workflow.")
```
yiminc: those warning log is very useful for user to realize their setup has issues.

internal_workflow_testsuite.go
```diff
@@ -257,19 +258,34 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(optio
...
 
-func (env *testWorkflowEnvironmentImpl) setActivityTaskList(tasklist string, activityFns ...interface{}) {
+func (env *testWorkflowEnvironmentImpl) setActivityTaskList(
+	tasklist string,
+	ao RegisterActivityOptions,
+	activityFns ...interface{},
+) {
 	for _, activityFn := range activityFns {
-		env.testSuite.RegisterActivity(activityFn)
+		env.getHostEnv().RegisterActivityWithOptions(activityFn, ao)
```
yiminc: we want test to use both the private host env and also the global host env as well, so that activity/workflow registered by init() method should be available to tests and don't require tests to register them via the private host again.

activity.go
```diff
@@ -57,8 +62,27 @@ type (
 // Serialization of all primitive types, structures is supported ... except channels, functions, variadic, unsafe pointer.
 // This method calls panic if activityFunc doesn't comply with the expected format.
 func RegisterActivity(activityFunc interface{}) {
```
madhuravi: The intention was not backwards compatibility. In a lot of cases users will not provide a name, having to add an empty struct is not a great user experience. So they can simple call RegisterActivity(fn) if they dont have an external name.

internal_worker.go
madhuravi: It might be a false warning here depending on how their code is set up. For instance they might not have Register in init function. What we need to do is add this warning in Start. If the worker is started and no workflows/activities are registered, issue a warning. I can add that.

internal_event_handlers.go
```diff
@@ -89,6 +88,7 @@ type (
 		enableLoggingInReplay bool // flag to indicate if workflow should enable logging in replay mode
 
 		metricsScope tally.Scope
+		hostEnv      *hostEnvImpl
```
madhuravi: Sounds like this is related to your earlier comment about using getHostEnvironment everywhere.

Creating a reference of host environment in the entry point and propagating that through all other workers and pollers makes testing a lot easier. So when we use a test workflow environment, we can create a task handler with the same host env instead of the global one for instance. Was almost impossible to get tests working without this.

This is also nice with design since it makes future refactoring easier and ensures all the objects bundled together use the same host environment They just all happen to use the global one since thats what the aggregated worker decides.

internal_workflow_testsuite.go
madhuravi: What use case does that solve? It's very beneficial to have self-contained tests. Tests should register workflows/activities that they are using. We can always use reusable helper methods to make that easier.

internal_workflow_testsuite.go
```diff
@@ -326,11 +327,27 @@ func (env *testWorkflowEnvironmentImpl) executeWorkflowInternal(workflowType str
 	env.startMainLoop()
 }
 
+func (env *testWorkflowEnvironmentImpl) getWorkflowDefinition(wt WorkflowType) (workflowDefinition, error) {
+	hostEnv := env.getHostEnv()
```
madhuravi: See comment above. Let's discuss this there as well.

internal_public.go
```diff
@@ -118,11 +118,7 @@ func NewWorkflowTaskHandler(domain string, identity string, logger *zap.Logger)
 		Identity: identity,
 		Logger:   logger,
 	}
-	return newWorkflowTaskHandler(
-		getWorkflowDefinitionFactory(newRegisteredWorkflowFactory()),
-		domain,
-		params,
-		nil)
+	return newWorkflowTaskHandler(domain, params, nil, getHostEnvironment())
```
madhuravi: See my reply on Siva's comment. Seems to be similar to what he was asking.

internal_workflow_testsuite.go
yiminc: This was added before we were able to mock activity. With the ability to mock activity/workflow, we don't need that private host env any more. :)

internal_workflow_testsuite.go
yiminc: please remove this env.getHostEnv() and use the global version of getHostEnvironment().

internal_workflow_testsuite_test.go
```diff
@@ -77,6 +77,7 @@ func (s *WorkflowTestSuiteUnitTest) Test_ActivityMockFunction() {
 
 func (s *WorkflowTestSuiteUnitTest) Test_ActivityMockValues() {
 	env := s.NewTestWorkflowEnvironment()
+	env.SetActivityTaskList(defaultTestTaskList, RegisterActivityOptions{}, testActivityHello)
```
madhuravi: This was due to a change in behavior causing testWorkflowEnv not to work properly. Fixed now.

internal_worker.go
madhuravi: Added this change to the PR.

internal_event_handlers.go
yiminc: the test framework should work with the global host env as we are deprecate the private host env.

internal_worker.go
```diff
@@ -405,53 +396,81 @@ func (th *hostEnvImpl) AddActivityRegistrationInterceptor(i interceptorFn) {
...
+		return fmt.Errorf("activity type \"%v\" is already registered", registerName)
 	}
 	// Register args with encoding.
 	if err := th.registerEncodingTypes(fnType); err != nil {
 		return err
 	}
-	fnName, af = th.invokeInterceptors(fnName, af, th.activityRegistrationInterceptors)
-	th.addActivityFn(fnName, af)
+	registerName, af = th.invokeInterceptors(registerName, af, th.activityRegistrationInterceptors)
+	th.addActivityFn(registerName, af)
```
yiminc: I think we might have problem with this interceptor. We transform the rergisterName when register, but we do not transform when invoke. cc @mfateev  (Note: this is not related to this PR)

internal_worker.go
```diff
@@ -615,6 +687,16 @@ func (th *hostEnvImpl) isUseThriftDecoding(objs []interface{}) bool {
 	return true
 }
 
+func (th *hostEnvImpl) getWorkflowDefinition(wt WorkflowType) (workflowDefinition, error) {
+	wf, ok := th.getWorkflowFn(wt.Name)
```
yiminc: that wt.Name could be an alias? Do we handle alias here?

internal_worker.go
```diff
@@ -765,13 +854,19 @@ var once sync.Once
...
+}
+
 func getHostEnvironment() *hostEnvImpl {
 	once.Do(func() {
-		thImpl = &hostEnvImpl{
-			workflowFuncMap: make(map[string]interface{}),
-			activityFuncMap: make(map[string]interface{}),
-			encoding:        gobEncoding{},
-		}
+		thImpl = newHostEnvironment()
```
yiminc: can we directly assign newHostEnvironment() to var thImpl and remove this once.Do()

internal_workflow_testsuite.go
```diff
@@ -604,6 +618,10 @@ func (env *testWorkflowEnvironmentImpl) CompleteActivity(taskToken []byte, resul
 	return nil
 }
 
+func (env *testWorkflowEnvironmentImpl) getHostEnv() *hostEnvImpl {
+	return env.testSuite.hostEnv
```
yiminc: we shouldn't need this method. We should use the global one.

internal_worker.go
madhuravi: Why do you think that? The interceptor tests are still working after this change.

internal_workflow_testsuite.go
madhuravi: Let's keep this discussion on your PR. I added some thoughts on global vs keeping the private env there. This shouldn't be a blocker on this one. We can make the change in parallel.

internal_worker.go
madhuravi: That would create a new host environment everytime instead of reusing the existing global one. Not the behavior we want.

PR #207, opened 2017-08-07
adding more client metrics

yiminc:
Adding more client side metrics for worker start counter and poller start counter.


internal_worker.go
```diff
@@ -977,6 +978,20 @@ func newAggregatedWorker(
 	}
 }
 
+func tagScope(metricsScope tally.Scope, tags ...string) tally.Scope {
```
sivakku: Where do we need tags being more than one set?

yiminc: We don't need more than one for now.

PR #208, opened 2017-08-08
remove private hostEnv from test suite

yiminc:
We added the private hostEnv to test suite before we were able to mock activity/workflow. Now with mock, we don't really need this separate hostEnv.

### DIFF
--- a/internal_worker_test.go
+++ b/internal_worker_test.go
@@ -41,27 +41,40 @@ import (
 	"go.uber.org/zap"
 )
 
-// Used to test registration listeners
-var registeredActivities []string
-var registeredWorkflows []string
-
 func init() {
-	registerWithSample(getHostEnvironment())
-}
-
-func registerWithSample(hostEnv *hostEnvImpl) {
-	hostEnv.RegisterWorkflowWithOptions(
+	RegisterWorkflowWithOptions(
 		sampleWorkflowExecute,
 		RegisterWorkflowOptions{Name: "sampleWorkflowExecute"},
 	)
-	hostEnv.AddWorkflowRegistrationInterceptor(
-		func(workflowName string, workflow interface{}) (string, interface{}) {
-			registeredWorkflows = append(registeredWorkflows, workflowName)
-			return workflowName, workflow
-		},
-	)
-	hostEnv.RegisterWorkflow(testReplayWorkflow)
+	RegisterWorkflow(testReplayWorkflow)
 
+	RegisterActivityWithOptions(
+		testActivity,
+		RegisterActivityOptions{Name: "testActivity"},
+	)
+	RegisterActivity(testActivityByteArgs)
+	RegisterActivityWithOptions(
+		testActivityMultipleArgs,
+		RegisterActivityOptions{Name: "testActivityMultipleArgs"},
+	)
+	RegisterActivity(testActivityReturnString)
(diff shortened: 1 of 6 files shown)


madhuravi:
Another option is to merge this in after my PR is in. Let me know what you prefer.

internal_worker_test.go
```diff
@@ -63,38 +61,15 @@ func init() {
 	RegisterActivity(testActivityReturnString)
 	RegisterActivity(testActivityReturnEmptyString)
 	RegisterActivity(testActivityReturnEmptyStruct)
-}
 
-func TestActivityRegistrationListener(t *testing.T) {
```
madhuravi: Where do we test the registration interceptors now?

internal_workflow_testsuite.go
```diff
@@ -343,8 +341,10 @@ func (env *testWorkflowEnvironmentImpl) executeActivity(
 		input,
 	)
 
-	// ensure activityFn is registered to defaultTestTaskList
-	env.testSuite.RegisterActivity(activityFn)
+	if _, ok := getHostEnvironment().getActivityFn(fnName); !ok {
```
madhuravi: Why not continue using the env on testWorkflowEnvironment and let the default be the global GetHostEnvironment() and allow someone to override it if required? This will allow us to test with isolated host environments.

yiminc: We don't need that anymore since we are able to mock.

madhuravi: Private env allows more flexibility than mock though. For eg. an empty test env with no workflows or activities. I don't see any downside to keeping this around and it provides added flexibility. In the newTestWorkflowEnv we can always set private env to the global getHostEnv so thats always the case by default. If we choose to override in tests now or later, we still can.

yiminc: We don't really have a use case for that private host env. As we are settling down to global registration, we should remove that private version for test suite. We could always revisit and refactor the code if we decided to do differently in the future. Offering more functionality confuse people, for example, the Registering method on TestSuite. More code also make it more likely to have more bug in it.

madhuravi: That's valid, we can remove the Register* methods on the TestSuite. I want to keep the functionality to isolate tests if possible. Having a private env on the testsuite should be simple and not confusing. Self-contained tests enable us to do a lot eg. run tests in parallel.

madhuravi: Also testing registration interceptors is a great example of when self contained tests are simple and easy to understand.

yiminc: If we remove the Register* method from test suite, how can you make use of the private hostEnv?

madhuravi: Made this change in my PR itself. Removed the Register* methods. For a self contained test, you can still create WorkflowTestSuite instance with a new host environment and register on that environment.

yiminc: The WorkflowTestSuite is our public API for our external users to do unit test. Exposing internal stuff like host env is not a good idea, and also it won't work for external use case.

madhuravi: Makes sense, my PR doesnt expose host env as it is. Let's discuss this in person.

internal_worker_test.go
sivakku: any idea we removed this? we would interceptors for the catalyst.

PR #209, opened 2017-08-08
propagate timeouts in test suite

yiminc:
propagate timeouts

### DIFF
--- a/internal_workflow_testsuite.go
+++ b/internal_workflow_testsuite.go
@@ -335,12 +335,18 @@ func (env *testWorkflowEnvironmentImpl) executeActivity(
 		panic(err)
 	}
 
+	params := executeActivityParameters{
+		ActivityType: ActivityType{Name: fnName},
+		Input:        input,
+		ScheduleToCloseTimeoutSeconds: 600,
+		StartToCloseTimeoutSeconds:    600,
+	}
+
 	task := newTestActivityTask(
 		defaultTestWorkflowID,
 		defaultTestRunID,
 		"0",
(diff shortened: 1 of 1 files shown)


internal_workflow_testsuite.go
```diff
@@ -956,20 +961,20 @@ func (env *testWorkflowEnvironmentImpl) newTestActivityTaskHandler(taskList stri
...
 		TaskToken:                     []byte(activityID), // use activityID as TaskToken so we can map TaskToken in heartbeat calls.
-		ActivityType:                  &shared.ActivityType{Name: common.StringPtr(activityType)},
-		Input:                         input,
+		ActivityType:                  &shared.ActivityType{Name: common.StringPtr(params.ActivityType.Name)},
+		Input:                         params.Input,
 		ScheduledTimestamp:            common.Int64Ptr(time.Now().UnixNano()),
-		ScheduleToCloseTimeoutSeconds: common.Int32Ptr(60),
+		ScheduleToCloseTimeoutSeconds: common.Int32Ptr(params.ScheduleToCloseTimeoutSeconds),
 		StartedTimestamp:              common.Int64Ptr(time.Now().UnixNano()),
-		StartToCloseTimeoutSeconds:    common.Int32Ptr(60),
```
sivakku: Can you add a unit test very like to break these kind of changes?

yiminc: not sure I understand. It is not easy to trigger timeout in unit test. It is only possible to trigger if you are debug it and step though which takes long time.

PR #213, opened 2017-08-11
Fix and add negative test case for fully qualified function name string

madhuravi:


internal_worker.go
```diff
@@ -547,14 +547,6 @@ func (th *hostEnvImpl) lenWorkflowFns() int {
 	return len(th.workflowFuncMap)
 }
 
-func (th *hostEnvImpl) lookupActivity(fnName string) (activity, bool) {
```
sivakku: I see.. this one makes it work, either way, which we want to break.

internal_workflow_testsuite.go
```diff
@@ -277,12 +277,10 @@ func (env *testWorkflowEnvironmentImpl) executeWorkflow(workflowFn interface{},
 	case reflect.String:
 		workflowType = workflowFn.(string)
 	case reflect.Func:
-		// auto register workflow if it is not already registered
```
sivakku: This makes it test code simpler, i was hoping we make it easier not to register explicitly for test code.

madhuravi: let's chat about this.

PR #215, opened 2017-08-15
Increase poll backoff exponential timeouts.

sivakku:


internal_worker_base.go
```diff
@@ -36,8 +36,8 @@ import (
 )
 
 const (
-	retryPollOperationInitialInterval    = time.Millisecond
-	retryPollOperationMaxInterval        = 1 * time.Second
+	retryPollOperationInitialInterval    = 20 * time.Millisecond
+	retryPollOperationMaxInterval        = 60 * time.Second
```
samarabbas: Can we use 60 seconds as the expiration interval and 10 seconds as max interval?
We should also emit a log if poll fails after all retries are still exhausted.

sivakku: This fix didn't help for the memory leak. but i am going still land this adjustments, will be helpful.

Changed the max to 10 seconds, we don't use the retryPollOperationExpirationInterval in poll so didn't change it. Since poll anyway is retrying we didn't use the policy retrier. We have log for each poll failure and that is too verbose to look at it.

Added metric to differentiate between transient and non-transient poll failures.

internal_task_pollers.go
```diff
@@ -204,7 +205,11 @@ func (wtp *workflowTaskPoller) poll() (*workflowTask, error) {
 
 	response, err := wtp.service.PollForDecisionTask(ctx, request)
 	if err != nil {
-		wtp.metricsScope.Counter(metrics.DecisionPollFailedCounter).Inc(1)
+		if isServiceTransientError(err) {
+			wtp.metricsScope.Counter(metrics.DecisionPollTransientFailedCounter).Inc(1)
```
samarabbas: Can you create a separate counter for server busy.

PR #219, opened 2017-08-24
add cadence- prefix to all metrics emit from library

yiminc:
add "cacence-" prefix to all metrics emit by client library code.


common/metrics/service_wrapper.go
```diff
@@ -45,24 +45,24 @@ type (
 )
 
 const (
-	scopeNameDeprecateDomain                = "DeprecateDomain"
```
madhuravi: Check https://github.com/uber-go/tally#report-your-metrics

It is possible to add a prefix when creating the tally scope. That would be a cleaner approach than adding this string everywhere.

yiminc: This tally scope is also used by user code and we don't want their metrics to have this prefix.

madhuravi: Can we create a subscope then? Once we get the users scope, create a subscope with the cadence name and use that everywhere.

At the very least, this should be a constant that  all metrics concatenate instead of pasting the same string everywhere.

yiminc: I tried to use subscope, it is not as easy as it sounds. The way our code is implemented, the metrics scope is shared by our library code and the user code. If we want to use subscope, we would need to create a new filed for multiple places to hold the original user's metrics and the subscope. I would rather not do in this change and maybe revisit if we want to make that change later.
Added constant for the prefix definition to avoid putting string everywhere.

common/metrics/service_wrapper.go
```diff
@@ -45,24 +45,24 @@ type (
...
-	scopeNameRequestCancelWorkflowExecution = "RequestCancelWorkflowExecution"
-	scopeNameRespondActivityTaskCanceled    = "RespondActivityTaskCanceled"
-	scopeNameRespondActivityTaskCompleted   = "RespondActivityTaskCompleted"
-	scopeNameRespondActivityTaskFailed      = "RespondActivityTaskFailed"
-	scopeNameRespondDecisionTaskCompleted   = "RespondDecisionTaskCompleted"
-	scopeNameSignalWorkflowExecution        = "SignalWorkflowExecution"
-	scopeNameStartWorkflowExecution         = "StartWorkflowExecution"
-	scopeNameTerminateWorkflowExecution     = "TerminateWorkflowExecution"
-	scopeNameUpdateDomain                   = "UpdateDomain"
+	scopeNameDeprecateDomain                = "cadence-DeprecateDomain"
```
madhuravi: Use CadencePrefix constant here as well.

PR #226, opened 2017-09-13
query API client impl

yiminc:
this CR does not include encoding stuff, which will be addressed in a separate CR shortly after.

### DIFF
--- a/internal_event_handlers.go
+++ b/internal_event_handlers.go
@@ -39,8 +39,6 @@ import (
 var _ workflowEnvironment = (*workflowEnvironmentImpl)(nil)
 var _ workflowExecutionEventHandler = (*workflowExecutionEventHandlerImpl)(nil)
 
-const queryTypeStackTrace string = "__stack_trace"
-
 type (
 	// completionHandler Handler to indicate completion result
 	completionHandler func(result []byte, err error)
@@ -580,7 +578,7 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 }
 
 func (weh *workflowExecutionEventHandlerImpl) ProcessQuery(queryType string, queryArgs []byte) ([]byte, error) {
-	if queryType == queryTypeStackTrace {
+	if queryType == QueryTypeStackTrace {
 		return getHostEnvironment().encodeArg(weh.StackTrace())
 	}
 	return weh.queryHandler(queryType, queryArgs)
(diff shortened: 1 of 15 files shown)


internal_event_handlers.go
```diff
@@ -39,6 +39,8 @@ import (
 var _ workflowEnvironment = (*workflowEnvironmentImpl)(nil)
 var _ workflowExecutionEventHandler = (*workflowExecutionEventHandlerImpl)(nil)
 
+const queryTypeStackTrace string = "__stack_trace"
```
mfateev: I think we can make constants for stack trace (and other built in queries) public

workflow.go
```diff
@@ -625,3 +626,43 @@ const DefaultVersion Version = -1
 func GetVersion(ctx Context, changeID string, minSupported, maxSupported Version) Version {
 	return getWorkflowEnvironment(ctx).GetVersion(changeID, minSupported, maxSupported)
 }
+
+// SetQueryHandler sets the query handler to handle workflow query. The queryType specify which query type this handler
```
mfateev: I believe that we discussed that query function is executed outside of dispatcher context. We have somehow document that it cannot use any workflow blocking functions like channel.get, etc.

workflow.go
```diff
@@ -625,3 +626,43 @@ const DefaultVersion Version = -1
...
+// All the input parameter must be serializable. You should call cadence.SetQueryHandler() at the beginning of the workflow
+// code.
+// Example of workflow code that support query type "current_state":
+//  func MyWorkflow(ctx cadence.Context, input string) error {
+//    currentState := "started" // this could be any serializable struct
+//    err := cadence.SetQueryHandler(ctx, "current_state", func() (string, error) {
+//      return currentState, nil
+//    })
+//    if err != nil {
+//      currentState = "failed to register query handler"
```
mfateev: Should we support querying closed and failed workflows? And what is the semantic of it?
If not this example doesn't make sense as it updates currentState before closing the workflow.

internal_event_handlers.go
yiminc: will change to public and move to Client.go

workflow.go
yiminc: Will update comments to point that out. However, if user do try to use Channel.Get, the code will block indefinitely. We probably need to error out in that case, meaning that we need to detect if those function is called outside of a workflow context. I will address that in a separate CR.

workflow.go
yiminc: Technically, we could support. The current implementation does support it. The use case is that user don't need to check the open/closed state of a workflow before sending out query. User could also use query to return workflow result or activity result if they wish to. (However, extending visibility API to return workflow result would be much cheaper.) I don't know, I don't feel strong about this. We can discuss again on this.

PR #227, opened 2017-09-19
detect blocking call in query handler

yiminc:
add code to detect illegal access to coroutinue from outside of workflow context, this is to prevent blocking calls from within query handler code.

### DIFF
--- a/internal_workflow.go
+++ b/internal_workflow.go
@@ -40,6 +40,8 @@ import (
 
 const (
 	defaultSignalChannelSize = 100000 // really large buffering size(100K)
+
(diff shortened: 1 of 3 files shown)


internal_workflow.go
```diff
@@ -461,7 +461,11 @@ func getState(ctx Context) *coroutineState {
 	if s == nil {
 		panic("getState: not workflow context")
 	}
-	return s.(*coroutineState)
+	state := s.(*coroutineState)
+	if !state.dispatcher.executing {
+		panic("getState: access from outside of workflow context")
```
mfateev: It would be nice to produce more descriptive error when called from query function.

PR #230, opened 2017-09-21
fix rate limiter bug

yiminc:
This is to fix the bug we found in canary alert.
The way we created the RateLimiter is incorrect. Here is the API:
// NewLimiter returns a new Limiter that allows events up to rate r and permits
// bursts of at most b tokens.
func NewLimiter(r Limit, b int) *Limiter

Here is how we use it:
rate.NewLimiter(rate.Every(maxTaskRateRefreshDuration), maxTaskRate)
In canary case, refresh duration is 1s, and the maxTaskRate is 100k. This means no rate limit for first 100K calls, then it will limit to 1 request for every 1s.

The correct usage should be:
rate.NewLimiter(maxTaskPerSecond, 1)
This means allow maxTaskPerSeconds for every 1s, and no burst.

### DIFF
--- a/internal_worker_base.go
+++ b/internal_worker_base.go
@@ -82,13 +82,12 @@ type (
 
 	// baseWorkerOptions options to configure base worker.
 	baseWorkerOptions struct {
-		pollerCount                int
-		maxConcurrentTask          int
-		maxTaskRate                int
-		maxTaskRateRefreshDuration time.Duration
-		taskWorker                 taskPoller
-		identity                   string
-		workerType                 string
+		pollerCount       int
+		maxConcurrentTask int
+		maxTaskPerSecond  float64
+		taskWorker        taskPoller
+		identity          string
+		workerType        string
 	}
 
 	// baseWorker that wraps worker activities.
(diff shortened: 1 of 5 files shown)


worker.go
```diff
@@ -53,18 +52,12 @@ type (
...
 		// default: defaultMaxActivityExecutionRate(100k)
 		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to rate limiting.
-		MaxActivityExecutionRate int
-
-		// Optional: Sets the refresh duration for rate limit. If not specified, it uses 1s as default.
-		// Use this to fine tune your rate limiter. For example, you could set this duration to 10s with max rate set to
-		// 1 which means we rate limit the activity task to at most 1 per every 10s. You could also set this duration to
-		// 100ms with max rate set wo 10 which means rate limit to 10 activity task execution per every 100ms.
-		MaxActivityExecutionRateRefreshDuration time.Duration
+		MaxActivityExecutionPerSecond float64
```
venkat1109: can you add a comment / sample code on how to specify a limit < 1 rps ?

internal_worker_base.go
```diff
@@ -127,8 +126,8 @@ func newBaseWorker(options baseWorkerOptions, logger *zap.Logger, metricsScope t
 	return &baseWorker{
 		options:         options,
 		shutdownCh:      make(chan struct{}),
-		pollLimiter:     rate.NewLimiter(rate.Every(time.Millisecond*100), 100),
-		taskLimiter:     rate.NewLimiter(rate.Every(options.maxTaskRateRefreshDuration), options.maxTaskRate),
+		pollLimiter:     rate.NewLimiter(1000, 1),
+		taskLimiter:     rate.NewLimiter(rate.Limit(options.maxTaskPerSecond), 1),
```
samarabbas: Do we really need two separate rate limiters? This could cause issues of task sitting in the buffer and getting timedout.

yiminc: https://github.com/uber-go/cadence-client/issues/231

PR #232, opened 2017-09-21
default to json encoding

yiminc:
switch to use json encoding as default encoder.

### DIFF
--- a/internal_worker.go
+++ b/internal_worker.go
@@ -360,8 +360,8 @@ type hostEnvImpl struct {
 	workflowAliasMap                 map[string]string
 	activityFuncMap                  map[string]activity
 	activityAliasMap                 map[string]string
-	encoding                         jsonEncoding
-	tEncoding                        thriftEncoding
+	encoding                         encoding
+	tEncoding                        encoding
 	activityRegistrationInterceptors []interceptorFn
 	workflowRegistrationInterceptors []interceptorFn
 }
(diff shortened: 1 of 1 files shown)


internal_worker.go
```diff
@@ -359,7 +360,7 @@ type hostEnvImpl struct {
 	workflowAliasMap                 map[string]string
 	activityFuncMap                  map[string]activity
 	activityAliasMap                 map[string]string
-	encoding                         gobEncoding
+	encoding                         jsonEncoding
```
mfateev: shouldn't this be of encoding type?

yiminc: yes, it should be using the interface encoding here and then assign a concrete implementation to it.

internal_worker.go
```diff
@@ -850,7 +851,8 @@ func newHostEnvironment() *hostEnvImpl {
 		workflowAliasMap: make(map[string]string),
 		activityFuncMap:  make(map[string]activity),
 		activityAliasMap: make(map[string]string),
-		encoding:         gobEncoding{},
+		encoding:         jsonEncoding{},
+		tEncoding:        thriftEncoding{},
```
yiminc: just find out that we never set the thriftEncoding (it is always nil), and since nobody complained about it, we are sure no one is using thrift encoding at this point.

mfateev: It could be using it directly.

Issue #236, opened 2017-09-27
flakey unit test case


yiminc:
I have seeing this unit test failure couple times, but it does not repro easily.

2017-09-27T13:14:26.614-0700	DEBUG	_obj_test/internal_workflow_testsuite.go:547	Auto fire timer	{"TimerID": 0, "TimerDuration": "1m0s", "TimeSkipped": "59.999s"}
--- FAIL: TestUnitTestSuite (0.15s)
    --- FAIL: TestUnitTestSuite/Test_OnActivityStartedListener (0.01s)
        Error Trace:    internal_workflow_testsuite_test.go:113
    	            	internal_workflow_testsuite.go:897
    	            	internal_workflow_testsuite.go:509
    	            	internal_workflow_testsuite.go:462
    	            	internal_workflow_testsuite.go:360
    	            	internal_workflow_testsuite.go:341
    	            	workflow_testsuite.go:215
    	            	internal_workflow_testsuite_test.go:122
    	Error:      	Received unexpected error:
    	            	unable to decode argument: 0, *string, with json error: EOF
        Error Trace:    internal_workflow_testsuite_test.go:125
    	Error:      	Not equal:
    	            	expected: []string{"testActivityHello:msg1", "testActivityHello:msg2", "testActivityHello:msg3"}
    	            	actual: []string{"testActivityHello:msg1", "testActivity:", "testActivityHello:msg2", "testActivityHello:msg3"}

    	            	Diff:
    	            	--- Expected
    	            	+++ Actual
    	            	@@ -1,3 +1,4 @@
    	            	-([]string) (len=3) {
    	            	+([]string) (len=4) {
    	            	  (string) (len=22) "testActivityHello:msg1",
    	            	+ (string) (len=13) "testActivity:",
    	            	  (string) (len=22) "testActivityHello:msg2",
FAIL

Issue #237, opened 2017-09-28
RegisterActivity is not thread safe


yiminc-zz:
If there are 2 goroutinues doing RegisterActivity() for 2 different functions with same friendly name, one of them should fail. But with current implementation, we don't use lock correctly, it is possible that both call will succeed and one will override the other.

PR #238, opened 2017-09-28
fix flakey unit test

yiminc:
This is to fix https://github.com/uber-go/cadence-client/issues/236

The bug is a regression introduced in https://github.com/uber-go/cadence-client/commit/0116f4710cdf2942d11fa5cb7ee731cf2f1dcee1

The test should not modify the global map of activities, before the above commit, the tests created a copy of activities and wrap them with test wrapper for activity handler, the above mentioned commit change to directly modify the global map.

This fix only fix part of the issue that currently is exposed. The deeper issue is the way we lock the global map. If we have multiple threads doing RegisterActivity(), then there will be race condition. The RegisterActivity() will check if activity is registered and add to the map only if it is not already registered. However, the RegisterActivity() does not use a lock. This needs to be fixed but it should be in its own CR. File new issue for that: https://github.com/uber-go/cadence-client/issues/237

### DIFF
--- a/internal_task_handlers.go
+++ b/internal_task_handlers.go
@@ -87,13 +87,14 @@ type (
 
 	// activityTaskHandlerImpl is the implementation of ActivityTaskHandler
 	activityTaskHandlerImpl struct {
-		taskListName string
-		identity     string
-		service      m.TChanWorkflowService
-		metricsScope tally.Scope
-		logger       *zap.Logger
-		userContext  context.Context
-		hostEnv      *hostEnvImpl
+		taskListName     string
+		identity         string
+		service          m.TChanWorkflowService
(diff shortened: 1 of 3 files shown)


internal_task_handlers.go
```diff
@@ -87,13 +87,14 @@ type (
...
-		userContext  context.Context
-		hostEnv      *hostEnvImpl
+		taskListName     string
+		identity         string
+		service          m.TChanWorkflowService
+		metricsScope     tally.Scope
+		logger           *zap.Logger
+		userContext      context.Context
+		hostEnv          *hostEnvImpl
+		activityProvider func(name string) activity
```
samarabbas: Can you define a type for this handler?

PR #239, opened 2017-09-28
adding context as first parameter to all methods on our client

yiminc:
Adding context.Context as first parameter to all public methods of Client and DomainClient. No change to worker API.

### DIFF
--- a/internal_task_handlers.go
+++ b/internal_task_handlers.go
@@ -892,7 +892,7 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 
 func (i *cadenceInvoker) internalHeartBeat(details []byte) (bool, error) {
 	isActivityCancelled := false
-	err := recordActivityHeartbeat(i.service, i.identity, i.taskToken, details, i.retryPolicy)
+	err := recordActivityHeartbeat(context.Background(), i.service, i.identity, i.taskToken, details, i.retryPolicy)
 
 	switch err.(type) {
 	case *CanceledError:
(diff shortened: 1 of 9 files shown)


internal_task_handlers.go
```diff
@@ -892,7 +892,7 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 
 func (i *cadenceInvoker) internalHeartBeat(details []byte) (bool, error) {
```
samarabbas: I thought the idea is for user to pass in the context.  Don't you need to update internalHeartBeat to also take in context along with details?

internal_task_pollers.go
```diff
@@ -217,10 +217,10 @@ func (wtp *workflowTaskPoller) poll() (*workflowTask, error) {
 		Identity: common.StringPtr(wtp.identity),
 	}
 
-	ctx, cancel := newTChannelContext(tchanTimeout(pollTaskServiceTimeOut), tchanRetryOption(retryNeverOptions))
+	tchCtx, cancel := newTChannelContext(context.Background(), tchanTimeout(pollTaskServiceTimeOut), tchanRetryOption(retryNeverOptions))
```
samarabbas: Should we also accept context from user for poll requests?  This will make graceful shutdown easier in the future.

internal_task_handlers.go
yiminc: no, this change only affect our Client/DomainClient interface. Poll task and respond task completed are not exposed through our Client interface. Workers are long running process and does not take context from user.

internal_task_pollers.go
yiminc: We can discuss that separately, but it is not in the scope of this change.

PR #241, opened 2017-10-02
no retry when context is done

yiminc:
retry should return when ctx is done

### DIFF
--- a/common/backoff/retry.go
+++ b/common/backoff/retry.go
@@ -21,6 +21,7 @@
 package backoff
 
 import (
+	"context"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


common/backoff/retry.go
```diff
@@ -106,6 +108,18 @@ func Retry(operation Operation, policy RetryPolicy, isRetryable IsRetryable) err
 			return err
 		}
 
+		// check if ctx is done
+		if ctxDone := ctx.Done(); ctxDone != nil {
```
molteanu: Could you add a test to cover this new logic?

Issue #243, opened 2017-10-03
Workflow Query: Potential coroutine leak


samarabbas:
If I query for callstack for workflow execution I'm seeing stack traces for coroutines which shouldn't be there.  For instances completed activities.

Here is an example
./cmd/cli/cadence-cli --env staging --domain cadence-canary query -qt __stack_trace -w cron.workflow.sanity -r 269b9e25-4836-4813-8a08-150c2cd65f78
Query result:
coroutine 1 [blocked on chan-15.Receive]:
code.uber.internal/devexp/cadence-canary/vendor/go.uber.org/cadence.(*futureImpl).Get(0xc420b4d040, 0x125b2a0, 0xc4208a4f00, 0x0, 0x0, 0xc420d21c20, 0x12607c0)
	code.uber.internal/devexp/cadence-canary/vendor/go.uber.org/cadence/internal_workflow.go:235 +0x70
code.uber.internal/devexp/cadence-canary/vendor/go.uber.org/cadence.Sleep(0x125b2a0, 0xc4208a4f00, 0xdbcac8e00, 0xddead0, 0xc420b4ceb0)
	code.uber.internal/devexp/cadence-canary/vendor/go.uber.org/cadence/workflow.go:441 +0x7d
code.uber.internal/devexp/cadence-canary/canary.cronWorkflow(0x125b2a0, 0xc4208a4f00, 0x14ea30978f4719fa, 0xc420b4c0f0, 0xc4208ec090, 0xf, 0xee6b2800, 0x0, 0x0)
	code.uber.internal/devexp/cadence-canary/canary/cron.go:49 +0x41b
reflect.Value.call(0xc62060, 0xddead8, 0x13, 0xd5bcf0, 0x4, 0xc420d0a000, 0x5, 0x5, 0x0, 0xc62060, ...)
	/usr/lib/go-1.8/src/reflect/value.go:434 +0x91f
reflect.Value.Call(0xc62060, 0xddead8, 0x13, 0xc420d0a000, 0x5, 0x5, 0x4, 0x4, 0x4)
	/usr/lib/go-1.8/src/reflect/value.go:302 +0xa4
[...]

PR #245, opened 2017-10-04
fix coroutinue leak before workflow completed

yiminc:
we create new coroutinue for every ExecuteActivity/ChildWorkflow and NewTimer to cancel them if the workflow context is canceled. The coroutinue should exit when activity/childworkflow/timer is completed.

This is to fix https://github.com/uber-go/cadence-client/issues/243

I have verified this fix with the query API that those coroutinue no longer show up.

### DIFF
--- a/workflow.go
+++ b/workflow.go
@@ -301,11 +301,14 @@ func ExecuteActivity(ctx Context, f interface{}, args ...interface{}) Future {
 		settable.Set(r, e)
 	})
 	Go(ctx, func(ctx Context) {
-		if ctx.Done() == nil {
-			return // not cancellable.
-		}
-		if ctx.Done().Receive(ctx, nil); ctx.Err() == ErrCanceled {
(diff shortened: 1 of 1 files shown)


workflow.go
```diff
@@ -418,13 +423,14 @@ func NewTimer(ctx Context, d time.Duration) Future {
 	})
 	if t != nil {
 		Go(ctx, func(ctx Context) {
-			if ctx.Done() == nil {
-				return // not cancellable.
+			if ctxDone := ctx.Done(); ctxDone != nil {
+				NewSelector(ctx).AddReceive(ctxDone, func(c Channel, more bool) {
+					// We will cancel the timer either it is explicit cancellation (or) we are closed.
```
sivakku: Can you add unit test for activity/timer/child workflow? you can use the API to query the stack and check of the co-routine still exist.

PR #249, opened 2017-10-16
Passing headers through yarpc call options

sivakku:


internal_utils.go
```diff
@@ -54,32 +55,48 @@ var retryDefaultOptions = &tchannel.RetryOptions{
...
+	Timeout time.Duration
+
+	// ParentContext to build the new context from. If empty, context.Background() is used.
+	// The new (child) context inherits a number of properties from the parent context:
+	//   - context fields, accessible via `ctx.Value(key)`
+	ParentContext context.Context
+}
+
+func (cb *contextBuilder) Build() (context.Context, context.CancelFunc) {
+	parent := cb.ParentContext
```
madhuravi: Wouldn't context.WithTimeout itself cover this? You shouldnt need these few lines. And as a consequence, is contextBuilder required at all?

yiminc:
looks good, but having the same question as Mahdu.

internal_utils.go
sivakku: you have two points.
(1) What you are saying is context.WithTimeout() can take first argument which can be nil? i had this throwing segmentation violation code. The playground sample: https://play.golang.org/p/IxNixZfxp2
(2) i have seen this context builder(with thrift to eventually grow up with more options) like restricting retries for some calls(like Poll) and they tend to keep adding on.. having builder will allow to pass options seamlessly in future.

sivakku: From context builder perspective it needs to work with if the parent context is not available, there is another wrapper function newChannelContext() that takes context, so in our cases we have it available b/c of wrapper one but for the builder it might be optional. 

Once we run yarpc(changes) with stress, i am sure we will discover to configure some knobs through options which needs context builder.

PR #250, opened 2017-10-17
Fix make rules for collecting and reporting code coverage

molteanu:
The current make rules only report code coverage for the last test folder they process.
This change collects all coverage data collected into a central file and reports that
to coveralls.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,3 +106,4 @@ fmt:
 
 clean:
 	rm -rf cadence-client
+	rm -Rf $(BUILD)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -71,21 +71,21 @@ bins: glide thriftc copyright lint
 	done;
 
 cover_profile: clean copyright lint glide
+	@mkdir -p $(BUILD)
```
venkat1109: can you also update the `clean:` target to delete the build folder ?

PR #251, opened 2017-10-17
marshal and unmarshal before mock setup

yiminc:
this is to work around the time difference before/after marshal/unmarshal. It would also fix for other custom struct that has private fields.


yiminc:
closing this PR for now, since we decided not to do it and let the error message be more clear when mock expectation is not met.

PR #262, opened 2017-10-25
remove startedEventId

yiminc:
Yesterday, we decided to remove startedEventId from PollForActivityTaskResponse.
When I work on the change, I realize there are some other places in history event attributes that we reference to startedEventId. They are not used anywhere in our client library.
I tried to remove them as well.


yiminc:
We also return startedEventId in PollForDecisionTaskResponse, as well as the DecisionTaskCompleted[TimedOut/Failed]EventAttributes. But they are not affected by the buffered event change. However, they are also not used by client either. I think might also want to remove them as well.

PR #267, opened 2017-11-03
add sticky decision task list

yiminc:
This is part one of the sticky decision change. This change added the host specific sticky decision task list.
This part does not include the partial history. The sticky decision task will still contains full history.
Partial history will be worked on in part 2 of the change.


internal_task_pollers.go
```diff
@@ -204,6 +223,48 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
...
 
+// getNextPollRequest returns appropriate next poll request based on poller configuration.
+// Simple rules:
+// 1) if sticky execution is disabled, always poll for regular task list
+// 2) otherwise:
+//   2.1) make sure there is at least one pending request for regular task list
+//   2.2) make sure there is at least one pending request for sticky task list
+//   2.3) poll from the tasklist that has less pending requests, if they are the same, poll from sticky task list.
+// TODO: make this more smart to auto adjust based on poll latency
+func (wtp *workflowTaskPoller) getNextPollRequest() (request *s.PollForDecisionTaskRequest, release func()) {
```
mfateev: I don't understand the polling logic fully. What is the number of polling threads in this case? I think we need separate goroutine count for default and process specific task lists. Then the logic outlined in the comment wouldn't be needed as task list would be passed as an argument.

internal_utils.go
```diff
@@ -101,11 +102,23 @@ func newChannelContext(ctx context.Context, options ...func(builder *contextBuil
...
 	}
-	return fmt.Sprintf("%d@%s@%s", os.Getpid(), hostName, tasklistName)
+	return hostName
+}
+
+// worker uuid per process
+var workerUUID string = uuid.New()
+
+func getWorkerTaskList(tasklist string) string {
+	// includes original tasklist and hostname for debuggability, workerUUID guarantees the uniqueness
```
mfateev: Why is it task list specific? We can have more than one global task list polled by a process. I don't think it means that we need to have multiple worker specific task lists. Having that number of workers can be potentially large doubling or even multiplying number of worker specific task lists could be not a good idea.

internal_task_pollers.go
yiminc: Currently we are using 2 poller threads (goroutinues) per worker. The number of threads is controlled by the baseWorker class. The baseWorker is used by both activity worker and workflow worker. The sticky task list only apply to workflow worker. BaseWorker does not aware of the sticky task list. I'm making the poller to make the decision for each poll request on which task list to poll from. So the base worker don't need to be aware of the sticky task list thing.

internal_utils.go
yiminc: Are you concerning that we have too many task list? I think the task list is super cheap on server. Our currently use case are all having one task list per worker host. But I agree that we could have one worker specific task list for multiple task lists. I can remove the task list from here if that is your concern.

PR #273, opened 2017-11-10
add lru cache

yiminc:
lru cache, copy from server repo. this will be needed to store sticky execution state.

### DIFF
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


common/cache/cache.go
```diff
@@ -0,0 +1,72 @@
...
+	// Put adds an element to the cache, returning the previous element
+	Put(key string, value interface{}) interface{}
+
+	// PutIfNotExist puts a value associated with a given key if it does not exist
+	PutIfNotExist(key string, value interface{}) (interface{}, error)
+
+	// Delete deletes an element in the cache
+	Delete(key string)
+
+	// Release decrements the ref count of a pinned element. If the ref count
```
mfateev: Which function pins the element. I couldn't find any reference in the docs.

yiminc: When you create the LRU cache, you supply optional Options. One of the option is to enable pin. When pin is enabled for a cache, all get request is a pin (with a refcount increase).

madhuravi:
On a tangential note, should we create a cadence-utils library of sorts? We seem to have a bit of duplicate code in general, will be harder to maintain.

PR #277, opened 2017-11-10
partial history

yiminc:
add partial history support

tested 3 cases for partial history
1) happy case;
2) disable cache (need code change) to mimic the cache been evicted case
3) always destroy cached state after each task (need code change) to mimic the error case
All cases worked.

open issues:
1) need better eviction policy for LRU, for now using fixed size (1000)
2) query task is not sticky (need minor server side change to make sure query task does not include new events that could potentially change state)
3) client side need update once query task is possible to be sticky
4) close decision could fail, but cache is unloaded anyway, need server side change to return error/flag to indicate that.
5) need way to remove stickiness in case of error.

### DIFF
--- a/internal_task_handlers.go
+++ b/internal_task_handlers.go
@@ -36,6 +36,7 @@ import (
 	s "go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/common"
 	"go.uber.org/cadence/common/backoff"
+	"go.uber.org/cadence/common/cache"
 	"go.uber.org/cadence/common/metrics"
 	"go.uber.org/cadence/common/util"
 	"go.uber.org/zap"
@@ -45,7 +46,6 @@ const (
 	defaultHeartBeatIntervalInSec = 10 * 60
 )
 
-// interfaces
 type (
 	// workflowExecutionEventHandler process a single event.
 	workflowExecutionEventHandler interface {
@@ -61,9 +61,9 @@ type (
 
 	// workflowTask wraps a decision task.
 	workflowTask struct {
-		task               *s.PollForDecisionTaskResponse
-		getHistoryPageFunc GetHistoryPage
-		pollStartTime      time.Time
+		task            *s.PollForDecisionTaskResponse
+		historyIterator HistoryIterator
+		pollStartTime   time.Time
 	}
 
 	// activityTask wraps a activity task.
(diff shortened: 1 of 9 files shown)


yiminc:
tested 3 cases for partial history
1) happy case;
2) disable cache (need code change) to mimic the cache been evicted case
3) always destroy cached state after each task (need code change) to mimic the error case
All cases worked.

mfateev:
I'm confused. The change contains refactoring of the iterator only. But the description claims that it contains the full support for workflow state caching.

yiminc:
The changes for caching is in internal_task_handlers.go. The diff view collapse the file when there is more than a few hundred lines of changes.

internal_task_handlers.go
```diff
@@ -308,83 +324,242 @@ func newWorkflowTaskHandler(
...
-// ProcessWorkflowTask processes each all the events of the workflow task.
-func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
-	task *s.PollForDecisionTaskResponse,
-	getHistoryPage GetHistoryPage,
-	emitStack bool,
-) (result interface{}, stackTrace string, err error) {
-	if task == nil {
-		return nil, "", errors.New("nil workflowtask provided")
+// TODO: need a better eviction policy based on memory usage
+var workflowCache = cache.New(1000, &cache.Options{
```
mfateev: Shouldn't cache be per worker?

internal_task_handlers.go
```diff
@@ -308,83 +324,242 @@ func newWorkflowTaskHandler(
...
+	workflowContext := &workflowExecutionContext{workflowStartTime: wfStartTime, workflowInfo: workflowInfo, wth: wth}
+	workflowContext.resetWorkflowState()
 
-	isWorkflowCompleted := false
-	var completionResult []byte
-	var failure error
+	return workflowContext, nil
+}
+
+func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(task *s.PollForDecisionTaskResponse,
```
mfateev: While I couldn't find any obvious problems with this method I believe it could be refactored to improve it readability.

yiminc: the problem is the query is not sticky for now which make it more complex than it needs to be.

internal_task_handlers.go
yiminc: i think it should be per process, or we could have multiple caches which could make it hard if we ever want to do something based on memory stats (which is global view).

mfateev:
OK. I don't have any more questions. Approved.

PR #299, opened 2017-12-02
handle ErrActivityResultPending in TestActivityEnvironment

yiminc:
When using TestActivityEnvironment to test activity and activity returns ErrActivityResultPending, it should not be treated as error.


internal_task_pollers.go
```diff
@@ -499,7 +499,7 @@ func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
 	}
 	atp.metricsScope.Timer(metrics.ActivityExecutionLatency).Record(time.Now().Sub(executionStartTime))
 
-	if request == nil {
+	if request == ErrActivityResultPending {
 		// this could be true when activity returns ErrActivityResultPending.
```
mfateev: Remove comment

PR #307, opened 2017-12-07
fix unit test race condition

yiminc:
this is the same issue as previous fix. Move the runningCount inc/dec into the main loop and only modify the counter in the main loop.

change type of runningCount from atomic.Int32 to int and make sure all inc/dec happen in main thread. Hopefully, this fix should close this data race issue.

Have verified the fix on Wenquan's machine and the issue is not repro anymore with this fix.


wxing1292:
i have tested this commit locally 37 times, works.

yiminc:
change type of runningCount from  atomic.Int32 to int and make sure all inc/dec happen in main thread. Hopefully, this fix should close this issue.

PR #314, opened 2017-12-14
add CancellationAlreadyRequestedError to no retry list

yiminc:


yiminc:
I did check that before I sent this PR. We defined 8 exceptions, 6 of them are listed here, other 2 are  ServiceBusyError, InternalServiceError.

PR #316, opened 2017-12-14
separate library version from feature version

wxing1292:
separate the library version from feature version, also include the client lang in the rpc call to service.

### DIFF
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -38,23 +38,45 @@ import (
 	"golang.org/x/net/context"
 )
 
-// versionHeaderName refers to the name of the
-// tchannel / http header that contains the client
-// library version
-const versionHeaderName = "cadence-client-version"
+const (
+	// libraryVersionHeaderName refers to the name of the
+	// tchannel / http header that contains the client
+	// library version
+	libraryVersionHeaderName = "cadence-client-library-version"
 
-// defaultRPCTimeout is the default tchannel rpc call timeout
-const defaultRPCTimeout = 10 * time.Second
+	// featureVersionHeaderName refers to the name of the
+	// tchannel / http header that contains the client
+	// feature version
+	featureVersionHeaderName = "cadence-client-feature-version"
 
-// retryNeverOptions - Never retry the connection
-var retryNeverOptions = &tchannel.RetryOptions{
-	RetryOn: tchannel.RetryNever,
-}
+	// languageHeaderName refers to the name of the
+	// tchannel / http header that contains the client
+	// language
+	languageHeaderName  = "cadence-client-language"
+	languageHeaderValue = "go"
 
-// retryDefaultOptions - retry with default options.
-var retryDefaultOptions = &tchannel.RetryOptions{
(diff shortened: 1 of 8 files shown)


internal/internal_utils.go
```diff
@@ -38,23 +38,45 @@ import (
...
+
+	// featureVersionHeaderName refers to the name of the
+	// tchannel / http header that contains the client
+	// feature version
+	featureVersionHeaderName = "cadence-client-feature-version"
+
+	// languageHeaderName refers to the name of the
+	// tchannel / http header that contains the client
+	// language
+	languageHeaderName  = "cadence-client-language"
```
mfateev: I would rename it to cadence-client-name
and value be something like uber-go, uber-java, as someone else can create mylibrary-java, etc.

internal/version.go
```diff
@@ -20,13 +20,20 @@
...
+
+// LibraryVersion is a semver string that represents
+// the version of this cadence client library
 const LibraryVersion = "v0.5.0"
+
+// FeatureVersion is a int that represents the
+// feature set of this cadence client library support.
+// This can be used for client capibility check, for
+// backward compatibility
+const FeatureVersion = "1"
```
mfateev: I think it should also be semantic version with MAJOR.MINOR version of the server it was tested against. The server itself would have MAJOR.MINOR.PATCH version, where PATCH would be for any changes that are not API changing.

internal/version.go
```diff
@@ -20,13 +20,20 @@
...
+// of headers in every rpc call made by this client to
+// cadence server.
+
+// Update to the metadata below is typically done
 // by the cadence team as part of a major feature or
 // behavior change
+
+// LibraryVersion is a semver string that represents
+// the version of this cadence client library
 const LibraryVersion = "v0.5.0"
```
mfateev: This version should represent API changes visibile to Cadence client side library consumers. I.e. developers that are writing workflows. So every time we change API that can affect them we have to change this number. And we can even add code to validate that workflow definition matches it in the future.

internal/internal_utils.go
mfateev: May be even call it user-agent to match HTTP?

internal/version.go
wxing1292: so what about this:

internally, we keep 3 version number on the client side (major, minor, patch), and when exposing to customer who relying on the version number, we just exposed_version = sprintf("v%s.%s.%s"), i.e. "v1.2.3"

when piggy the version back to server, we send back 3 version separately, like:
"cadence-client-major-ver": "1"
"cadence-client-minor-ver": "2"
"cadence-client-patch-ver": "3"
to avoid parsing.

mfateev: I'm confused what version are you talking about?
For feature version only two numbers are necessary, and parsing is trivial. 
For client version there is no need for parsing yet.

internal/internal_utils.go
wxing1292: changed to be "cadence-client-name" : "uber-go"

PR #317, opened 2017-12-14
fail decision task when decider panic

yiminc:
fail decision task when decider panic, verified that failed decision task event is appended to history with details.
when server receives the fail decision, it will generate a new decision task, to avoid spin around the failed decision task, the worker will not respond the the second attempt and the second decision task will timeout. we might want to backoff generating new decisions in this case, but that would be a separate task.


internal/internal_task_handlers.go
```diff
@@ -943,43 +873,97 @@ func isDecisionMatchEvent(d *s.Decision, e *s.HistoryEvent, strictMode bool) boo
...
+
+	if panicErr, ok := workflowContext.err.(*PanicError); ok {
+		// Workflow panic
+		wth.metricsScope.Counter(metrics.DecisionTaskPanicCounter).Inc(1)
+		wth.logger.Error("Workflow panic.",
+			zap.String("PanicError", panicErr.Error()),
+			zap.String("PanicStack", panicErr.StackTrace()))
+		failedCause := s.DecisionTaskFailedCauseWorkflowWorkerUnhandledFailure
+		_, details := getErrorDetails(panicErr)
+		completeRequest = &s.RespondDecisionTaskFailedRequest{
```
mfateev: I would change it to return to avoid all these elses. So code would look like:

if ... {
   return &s....
}
if ... {
   return &s...
}

internal/internal_task_pollers.go
```diff
@@ -190,6 +190,17 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
 			defer cancel()
 			var err1 error
 			switch request := completedRequest.(type) {
+			case *s.RespondDecisionTaskFailedRequest:
+				// Only fail decision on first attempt, subsequent failure on the same decision task will timeout.
```
mfateev: I'm confused. Wasn't it something that was fixed on the server side? I thought the model is that client always fails and server controls the backoff semantic.

yiminc: At least it is not implemented that way now. Server will immediately generate a new one. Without this trick here, it will rapidly spin.

mfateev: I see. Then it makes sense.

PR #328, opened 2018-01-04
sync thrift files, add RunWorkflow API which starts a workflow and ea…

wxing1292:
…it until the workflow finishes, return the last event


internal/internal_workflow_client.go
```diff
@@ -139,6 +141,49 @@ func (wc *workflowClient) StartWorkflow(
...
+		return nil, NewRunWorkflowError(executionInfo.RunID, "Error when tracking workflow history: "+err.Error())
+	}
+
+	closeEvent = history.Events[0]
+	errorMessage := ""
+	switch closeEvent.GetEventType() {
+	case s.EventTypeWorkflowExecutionCompleted:
+		return EncodedValue(closeEvent.WorkflowExecutionCompletedEventAttributes.Result), nil
+	case s.EventTypeWorkflowExecutionCanceled:
+		errorMessage = "Workflow execution is cancelled."
```
yiminc: It would be good if we could expose that error type as an enum field of the RunWorkflowError.

internal/client.go
```diff
@@ -50,8 +50,32 @@ type (
...
+		// WorkflowRun has 2 methods:
+		//  - GetRunID() string: which return the first started workflow run ID (please see below)
+		//  - Get(ctx context.Context, valuePtr interface{}) error: which will fill the workflow
+		//    execution result to valuePtr, if workflow execution is a success, or return corresponding
+		//    error. This is a blocking API.
+		// NOTE: if the started workflow return ContinueAsNewError during the workflow execution, the
+		// return result of GetRunID() will NOT be updated, however, Get(ctx context.Context, valuePtr interface{})
+		// will use the new run ID for workflow execution result.
+		// NOTE: DO NOT USE THIS API INSIDE A WORKFLOW
+		ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error)
```
yiminc: I thought we discussed and agreed on using RunWorkflow as method name and returning an encoded.Value and error. Looks like I missed some context here?

wxing1292: mfateev, samarabbas and wxing1292 had a discussion on last Friday, the idea is to make the ExecuteWorkflow (renamed) API having the same user experience as the ExecuteChildWorkflow

internal/internal_workflow_client.go
```diff
@@ -496,3 +585,126 @@ func getRunID(runID string) *string {
...
+	}
+
+	// we have cached events
+	if iter.nextEventIndex < len(iter.events) {
+		index := iter.nextEventIndex
+		iter.nextEventIndex++
+		return iter.events[index], nil
+	} else if iter.err != nil {
+		// we have err, clear that iter.err and return err
+		err := iter.err
```
mfateev: Why do we clear iter.err? Is it possible to reuse iterator after it returned error?

internal/client.go
wxing1292: @yiminc FYI, i am adding unit test

internal/internal_workflow_client.go
```diff
@@ -496,3 +585,126 @@ func getRunID(runID string) *string {
...
+		err = newTerminatedError()
+	case s.EventTypeWorkflowExecutionTimedOut:
+		attributes := closeEvent.WorkflowExecutionTimedOutEventAttributes
+		err = NewTimeoutError(attributes.GetTimeoutType())
+	case s.EventTypeWorkflowExecutionContinuedAsNew:
+		attributes := closeEvent.WorkflowExecutionContinuedAsNewEventAttributes
+		workflowRun.currentRunID = attributes.GetNewExecutionRunId()
+		return workflowRun.Get(ctx, valuePtr)
+	default:
+		err = errors.New("Unknown error occur when fetching workflow execution result")
```
mfateev: Could you provide more descriptive message (like the type of event returned instead of expected)

internal/client.go
```diff
@@ -50,8 +50,32 @@ type (
...
 		StartWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (*WorkflowExecution, error)
 
+		// ExecuteWorkflow starts a workflow execution and return a WorkflowRun instance and error
+		// The user can use this to start using a function or workflow type name.
+		// Either by
+		//     ExecuteWorkflow(ctx, options, "workflowTypeName", input)
+		//     or
+		//     ExecuteWorkflow(ctx, options, workflowExecuteFn, arg1, arg2, arg3)
+		// The errors it can return:
+		//	- EntityNotExistsError
```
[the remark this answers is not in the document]
wxing1292: added, and yes if domain does not exists, as far as i see

internal/client.go
```diff
@@ -50,8 +50,32 @@ type (
 		//	- EntityNotExistsError
 		//	- BadRequestError
 		//	- WorkflowExecutionAlreadyStartedError
+		//	- InternalServiceError
 		StartWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (*WorkflowExecution, error)
 
+		// ExecuteWorkflow starts a workflow execution and return a WorkflowRun instance and error
+		// The user can use this to start using a function or workflow type name.
+		// Either by
+		//     ExecuteWorkflow(ctx, options, "workflowTypeName", input)
```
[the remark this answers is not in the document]
wxing1292: ok

internal/client.go
```diff
@@ -50,8 +50,32 @@ type (
...
+		//
+		// WorkflowRun has 2 methods:
+		//  - GetRunID() string: which return the first started workflow run ID (please see below)
+		//  - Get(ctx context.Context, valuePtr interface{}) error: which will fill the workflow
+		//    execution result to valuePtr, if workflow execution is a success, or return corresponding
+		//    error. This is a blocking API.
+		// NOTE: if the started workflow return ContinueAsNewError during the workflow execution, the
+		// return result of GetRunID() will NOT be updated, however, Get(ctx context.Context, valuePtr interface{})
+		// will use the new run ID for workflow execution result.
+		// NOTE: DO NOT USE THIS API INSIDE A WORKFLOW
```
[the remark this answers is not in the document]
wxing1292: added "USE workflow.ExecuteChildWorkflow instead"

internal/client.go
```diff
@@ -80,14 +104,12 @@ type (
...
+		// GetWorkflowHistory gets the point in time snapshot of history of a particular workflow
 		// - workflow ID of the workflow.
-		// - runID can be default(empty string). if empty string then it will pick the running execution of that workflow ID.
-		// The errors it can return:
-		//	- EntityNotExistsError
-		//	- BadRequestError
-		//	- InternalServiceError
-		GetWorkflowHistory(ctx context.Context, workflowID string, runID string) (*s.History, error)
+		// - runID can be default(empty string). if empty string then it will pick the last running execution of that workflow ID.
+		// - whether use long poll for tracking new events
```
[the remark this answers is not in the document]
wxing1292: changed to:

// GetWorkflowHistory gets the point in time snapshot of history of a particular workflow
// - workflow ID of the workflow.
// - runID can be default(empty string). if empty string then it will pick the last running execution of that workflow ID.
// - whether use long poll for tracking new events: when the workflow is running, there can be new events generated during iteration
// 	 of HistoryEventIterator, if isLongPoll == true, then iterator will do long poll, tracking new history event, i.e. the iteration
//   will not be finished until workflow is finished; if isLongPoll == false, then iterator will only return current history events.
// - whether return all history events or just the last event, which contains the workflow execution end result
// Example:-
//	To iterate all events,
//		iter := GetWorkflowHistory(ctx, workflowID, runID, isLongPoll, filterType)
//		events := []*shared.HistoryEvent{}
//		for iter.HasNext() {
//			event, err := iter.Next()
//			if err != nil {
//				return err
//			}
//			events = append(events, event)
//		}

internal/internal_workflow_client.go
```diff
@@ -59,6 +61,57 @@ type (
...
+		GetRunID() string
+
+		// Get will fill the workflow execution result to valuePtr,
+		// if workflow execution is a success, or return corresponding,
+		// error. This is a blocking API.
+		Get(ctx context.Context, valuePtr interface{}) error
+
+		// NOTE: if the started workflow return ContinueAsNewError during the workflow execution, the
+		// return result of GetRunID() will NOT be updated, however, Get(ctx context.Context, valuePtr interface{})
+		// will use the new run ID for workflow execution result.
```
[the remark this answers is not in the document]
wxing1292: ok, changed the explanation to:

// NOTE: if the started workflow return ContinueAsNewError during the workflow execution, the
		// return result of GetRunID() will be the started workflow run ID, not the new run ID caused by ContinueAsNewError,
		// however, Get(ctx context.Context, valuePtr interface{}) will return result from the run which did not return ContinueAsNewError.
		// Say ExecuteWorkflow started a workflow, in its first run, has run ID "run ID 1", and returned ContinueAsNewError,
		// the second run has run ID "run ID 2" and return some result other than ContinueAsNewError:
		// GetRunID() will always return "run ID 1" and  Get(ctx context.Context, valuePtr interface{}) will return the result of second run.

internal/internal_workflow_client.go
```diff
@@ -496,3 +585,126 @@ func getRunID(runID string) *string {
...
+	}
+	return history, nil
+}
+
+func (iter *historyEventIteratorImpl) HasNext() bool {
+	return iter.nextEventIndex < len(iter.events) || len(iter.nexttoken) != 0 || iter.err != nil || !iter.initialized
+}
+
+func (iter *historyEventIteratorImpl) Next() (*s.HistoryEvent, error) {
+	// if caller call the Next() when iteration is over, just return nil, nil
```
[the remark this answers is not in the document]
wxing1292: ok changed to:
panic("HistoryEventIterator Next() called without checking HasNext()")

internal/internal_workflow_client.go
wxing1292: if the iterator returned error, and caller continue to use the iterator, it will panic, as specified above. so the iterator can only be used once.

internal/internal_workflow_client.go
wxing1292: changed to:
err = fmt.Errorf("Unexpected event type %s when handling workflow execution result", closeEvent.GetEventType())

internal/internal_workflow_client.go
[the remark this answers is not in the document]
mfateev: Looks good

internal/internal_workflow_client.go
wxing1292: and the the events and err should be only returned once

internal/client.go
```diff
@@ -268,6 +269,23 @@ type (
...
+
+	// WorkflowIDReusePolicy defines workflow ID reuse behavior.
+	WorkflowIDReusePolicy int
+)
+
+const (
+	// WorkflowIDReusePolicyAllowDuplicateFailedOnly allow start a workflow execution
+	// when workflow not running, and the last execution close state is in
+	// [terminated, cancelled, timeouted, failed].
+	WorkflowIDReusePolicyAllowDuplicateFailedOnly WorkflowIDReusePolicy = 0
```
mfateev: I think the pattern is to use iota for enums

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -907,7 +938,16 @@ struct DescribeWorkflowExecutionRequest {
   20: optional WorkflowExecution execution
 }
 
+struct PendingActivityInfo {
+  10: optional string activityID
+  20: optional ActivityType activityType
+  30: optional PendingActivityState state
+  40: optional binary heartbeatDetails
+  50: optional i64 (js.type = "Long") lastHeartbeatTimestamp
```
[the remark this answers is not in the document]
wxing1292: this is not part of my change, i only sync-ed the change from server side to client side.
@yiminc  can you provide more info here?

internal/client.go
wxing1292: ok, updated

idl/github.com/uber/cadence/shared.thrift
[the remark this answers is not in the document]
yiminc: added issue to track this. Should be pretty straightforward to do. https://github.com/uber/cadence/issues/501

PR #331, opened 2018-01-05
 Set up activity throttle limit as input and propagate to server

madhuravi:
Related to https://github.com/uber/cadence/pull/432


internal/internal_public.go
```diff
@@ -104,7 +104,8 @@ func NewActivityTaskWorker(
 		TaskList:                        taskList,
 		ConcurrentPollRoutineSize:       defaultConcurrentPollRoutineSize,
 		ConcurrentActivityExecutionSize: wOptions.MaxConcurrentActivityExecutionSize,
-		MaxActivityExecutionPerSecond:   wOptions.MaxActivityExecutionPerSecond,
+		WorkerActivitiesPerSecond:       wOptions.WorkerActivitiesPerSecond,
+		ActivitiesPerSecond:             wOptions.ActivitiesPerSecond,
```
mfateev: I would rename it to TaskListActivitiesPerSecond

internal/worker.go
```diff
@@ -49,14 +49,27 @@ type (
...
 		MaxConcurrentActivityExecutionSize int
 
-		// Optional: Sets the rate limiting on number of activities that can be executed per second. Notice that the
-		// number is represented in float, so that you can set it to less than 1 if needed. For example, set the number
-		// to 0.1 means you want your activity to be executed once for every 10 seconds. This can be used to protect
-		// down stream services from flooding.
+		// Optional: Sets the rate limiting on number of activities that can be executed per second per
+		// worker. Notice that the number is represented in float, so that you can set it to less than
+		// 1 if needed. For example, set the number to 0.1 means you want your activity to be executed
+		// once for every 10 seconds. This can be used to protect down stream services from flooding.
```
mfateev: This can be used to limit amount of resources used by a worker.

internal/worker.go
```diff
@@ -49,14 +49,27 @@ type (
...
+
+		// Optional: Sets the rate limiting on number of activities that can be executed per second.
+		// This is managed by the server and controls activities per second for your entire tasklist
+		// whereas WorkerActivityTasksPerSecond controls activities only per worker.
+		// Notice that the number is represented in float, so that you can set it to less than
+		// 1 if needed. For example, set the number to 0.1 means you want your activity to be executed
+		// once for every 10 seconds. This can be used to protect down stream services from flooding.
+		// The zero value of this uses the default value.
+		// default: defaultMaxActivityExecutionRate(100k)
+		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to
```
mfateev: I don't think it is true for global throttling.

internal/worker.go
```diff
@@ -49,14 +49,27 @@ type (
...
-		// down stream services from flooding.
+		// Optional: Sets the rate limiting on number of activities that can be executed per second per
+		// worker. Notice that the number is represented in float, so that you can set it to less than
+		// 1 if needed. For example, set the number to 0.1 means you want your activity to be executed
+		// once for every 10 seconds. This can be used to protect down stream services from flooding.
 		// The zero value of this uses the default value.
 		// default: defaultMaxActivityExecutionRate(100k)
-		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to rate limiting.
-		MaxActivityExecutionPerSecond float64
+		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to
```
mfateev: We need to fix it. There should be prefetch buffer size and when it is set to 0 poll shouldn't be issued if throttling limit is reached.

internal/worker.go
madhuravi: If throttling kicks in, the task will not be delivered to the poll channel and the activity's StartToCloseTimeout starts ticking till the poll times out.

madhuravi: It will actually be the *ScheduleToStart timeout in this case I think. Since the task is in there, but fails to get picked up by the worker since throttling is in effect.

internal/worker.go
madhuravi: Not sure what you mean here. Why would prefetch buffer size be set to 0? Poll will not be issued if throttling limit is reached, regardless of that.

internal/worker.go
mfateev: It is what I mean. We don't need this warning as activity just stays queued in the task list.

internal/worker.go
mfateev: The this warning is wrong.

internal/worker.go
```diff
@@ -66,10 +65,8 @@ type (
 		// 1 if needed. For example, set the number to 0.1 means you want your activity to be executed
 		// once for every 10 seconds. This can be used to protect down stream services from flooding.
 		// The zero value of this uses the default value.
-		// default: defaultMaxActivityExecutionRate(100k)
-		// Warning: activity's StartToCloseTimeout starts ticking even if a task is blocked due to
-		// rate limiting.
-		ActivitiesPerSecond float64
+		// default: defaultMaxActivityExecutionRate(100k).
```
mfateev: defaultTaskListActivitiesPerSecond

madhuravi: Thinking about it, doesn't make sense for a private variable name to be in public docs, will just remove the name and keep the value.

PR #337, opened 2018-01-09
catch worker panic

yiminc:
catch worker panic


internal/internal_worker_base.go
```diff
@@ -246,6 +247,18 @@ func (bw *baseWorker) pollTask() {
 }
 
 func (bw *baseWorker) processTask(task interface{}) {
+	defer func() {
+		if p := recover(); p != nil {
+			bw.metricsScope.Counter(metrics.WorkerPanicCounter).Inc(1)
```
mfateev: could we call fail task from here?

yiminc: We can not do it here, but we could put panic handler one layer lower to the activityPoller and workflowPoller.

internal/internal_event_handlers.go
```diff
@@ -436,6 +436,18 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	if event == nil {
 		return nil, errors.New("nil event provided")
 	}
+	defer func() {
+		if p := recover(); p != nil {
+			weh.metricsScope.Counter(metrics.DecisionTaskPanicCounter).Inc(1)
```
madhuravi: I would recommend checking weh.metricsScope != nil and weh.workflowInfo != nil
Panic could have occured due to those being nil. And you will run into issues inside recover.

yiminc: They should not be nil. We set the metricsScope even if user do not supply one. The workflowInfo is also set before we do any event processing. If they are not, that would be bug in our code and it might be better to let it crash. This panic handler is meant to catch issues in user code, for example invalid state transition due to ill history events.

PR #340, opened 2018-01-12
adding local activity

yiminc:
unit test support is not done yet, want to send out this earlier so i can get feedback sooner. Will work on test framework shortly.


internal/internal_task_handlers.go
```diff
@@ -651,14 +660,62 @@ ProcessEvents:
...
+	eventDecisions, err := workflowContext.eventHandler.ProcessLocalActivityResult(lar)
+	if err != nil {
+		return nil, err
+	}
+
+	if eventDecisions != nil {
+		workflowContext.newDecisions = append(workflowContext.newDecisions, eventDecisions...)
+	}
+
+	return workflowContext.CompleteDecisionTask(), nil
```
mfateev: I think I'm missing  something, but I cannot find where eventLoop is called when a local activity completes.

internal/internal_task_pollers.go
```diff
@@ -164,6 +200,18 @@ func (wtp *workflowTaskPoller) PollTask() (interface{}, error) {
 
 // ProcessTask processes a task
 func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
+	if lar, ok := task.(*localActivityResult); ok {
```
mfateev: Could we have two functions like ProcessTask and ProcessActivityResult which would delegate to some internal method for common parts. Having method that switches on the type of parameter is confusing.

internal/internal_task_pollers.go
```diff
@@ -100,8 +100,44 @@ type (
...
+		logger       *zap.Logger
+	}
+
+	localActivityResult struct {
+		result []byte
+		err    error
+		task   *localActivityTask
+	}
+
+	localActivityChannel struct {
```
mfateev: Rename, as initially I assumed that it is just a wrapper of a single channel.

internal/internal_task_pollers.go
```diff
@@ -235,13 +297,115 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
...
+		laStartTime := time.Now()
+		laResult, err = ae.ExecuteWithActualArgs(ctx, task.params.InputArgs)
+		lath.metricsScope.Timer(metrics.LocalActivityExecutionLatency).Record(time.Now().Sub(laStartTime))
+		close(ch)
+	}(doneCh)
+
+	// block until previous poll completed or return immediately when shutdown
+	select {
+	case <-ctx.Done():
+		// context is done
```
mfateev: May be asynchronously check for doneCh here to ensure that we are not throwing away result.

internal/internal_worker.go
```diff
@@ -220,6 +221,24 @@ func newWorkflowWorkerInternal(
 	return newWorkflowTaskWorkerInternal(taskHandler, service, domain, params)
 }
 
+var localActivityChannelMap = make(map[string]*localActivityChannel)
```
mfateev: Why do we have global map? instead we could have this shared between workflow and local activity worker during initialization and do not have task list name at all.

internal/internal_task_handlers.go
```diff
@@ -651,14 +660,62 @@ ProcessEvents:
...
+}
+
+func (w *workflowExecutionContext) CompleteDecisionTask() interface{} {
+	if !w.isWorkflowCompleted && len(w.eventHandler.pendingLaTasks) > 0 {
+		lac := getLocalActivityChannel(w.workflowInfo.TaskListName)
+		for _, task := range w.eventHandler.pendingLaTasks {
```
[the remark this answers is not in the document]
yiminc: will do

internal/internal_task_pollers.go
yiminc: will do

internal/internal_worker.go
yiminc: will remove

internal/internal_task_handlers.go
```diff
@@ -606,8 +614,9 @@ ProcessEvents:
 				return nil, "", err
 			}
 		}
-		isInReplay := reorderedEvents[0].GetEventId() < reorderedHistory.LastNonReplayedID()
+
 		for i, event := range reorderedEvents {
+			isInReplay := reorderedEvents[0].GetEventId() < reorderedHistory.LastNonReplayedID() || isLocalActivityMarkerEvent(event)
```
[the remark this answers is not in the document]
yiminc: we put DecisionTaskCompleted before new decision events, so a lastCompletedEventID will not help. We will need an explicit id (last decision event) if we want.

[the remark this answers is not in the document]
yiminc: Updated the history's LastNonReplayedID() to IsReplayEvent() which checks if the eventID is less or equal to PreviousStartedEventId, or if the event is decision event.

internal/internal_task_pollers.go
```diff
@@ -235,13 +313,121 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
...
+	}
+	task.cancelFunc = cancel
+	task.Unlock()
+
+	var laResult []byte
+	var err error
+	doneCh := make(chan struct{})
+	go func(ch chan struct{}) {
+		laStartTime := time.Now()
+		laResult, err = ae.ExecuteWithActualArgs(ctx, task.params.InputArgs)
```
mfateev: We are making an assumption that activity is going to cancel itself once context expires. I think it is reasonable assuption, but let's add a warning when activity actual execution time exceeded the deadline by some value.

PR #344, opened 2018-01-13
bugfix: registered activity with alias in test suite

wxing1292:


internal/internal_workflow_testsuite.go
```diff
@@ -353,15 +353,27 @@ func (env *testWorkflowEnvironmentImpl) executeActivity(
 	activityFn interface{},
 	args ...interface{},
 ) (encoded.Value, error) {
-	fnName := getFunctionName(activityFn)
+	var activityType string
+	fnType := reflect.TypeOf(activityFn)
+	switch fnType.Kind() {
```
yiminc: can we use getValidatedActivityFunction() here?

PR #355, opened 2018-01-18
Add isStickyTask to poll request

vancexu:
related server change: https://github.com/uber/cadence/pull/510/files


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -735,6 +735,7 @@ struct PollForDecisionTaskRequest {
   10: optional string domain
   20: optional TaskList taskList
   30: optional string identity
+  40: optional bool isStickyTask
```
yiminc: I thought we decided that we need an enum, not a boolean.

yiminc: also, could it make sense to add the enum to the TaskList struct, instead of here?

vancexu: I agree it make sense to add that info to TaskList, changed and tested

internal/internal_task_pollers.go
```diff
@@ -242,27 +242,22 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
...
-
-func (wtp *workflowTaskPoller) release(isStickyTask bool) {
+func (wtp *workflowTaskPoller) release(kind s.TaskListKind) {
 	if wtp.disableStickyExecution {
 		return
 	}
 
 	wtp.requestLock.Lock()
-	if isStickyTask {
+	if kind.Equals(s.TaskListKindWorker) {
```
yiminc: use == to compare for enum type

internal/internal_task_pollers.go
```diff
@@ -242,27 +242,22 @@ func (wtp *workflowTaskPoller) ProcessTask(task interface{}) error {
...
 	} else {
 		wtp.pendingRegularPollCount--
 	}
 	wtp.requestLock.Unlock()
 }
 
-func (wtp *workflowTaskPoller) updateBacklog(isStickyTask bool, backlogCountHint int64) {
-	if !isStickyTask || wtp.disableStickyExecution {
+func (wtp *workflowTaskPoller) updateBacklog(taskListKind s.TaskListKind, backlogCountHint int64) {
+	if taskListKind.Equals(s.TaskListKindNormal) || wtp.disableStickyExecution {
```
yiminc: use ==

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -196,6 +196,11 @@ enum HistoryEventFilterType {
   CLOSE_EVENT,
 }
 
+enum TaskListKind {
+  NORMAL,
+  WORKER,
```
yiminc: rename WORKER to STICKY

internal/internal_task_pollers.go
vancexu: changed, what's the benefits of using ==? Equals function itself use == inside.

PR #356, opened 2018-01-19
only check function name of child workflow for non-deterministic detection

yiminc:
we already loose the check for activity to only check the function name of activity, so we should do the same for child workflow.


yiminc:
@mfateev , that is what it is now. The Domain and TaskList is only checked in strict mode which is always false in our code.

PR #364, opened 2018-01-22
bugfix: test tag should have type string

wxing1292:


internal/internal_worker.go
```diff
@@ -58,6 +58,8 @@ const (
 
 	defaultMaxConcurrentWorkflowExecutionSize = 50     // hardcoded max workflow execution size.
 	defaultMaxWorkflowExecutionRate           = 100000 // Large workflow execution rate (unlimited)
+
+	testTagsContextKey = "testTags"
```
yiminc: to avoid key conflict, we should use a more explicit key name, like "cadence_testTags".

wxing1292: ok

PR #365, opened 2018-01-23
fix nil argument to activity

### LINKED ISSUE #360: ExecuteLocalActivity panics if nil is passed as argument to activity
Took me a while to track down what was happening.  https://github.com/uber-go/cadence-client/blob/master/internal/internal_activity.go#L176 panics if nil is an argument to an activity.  If nil is not allowed, it should ideally print a more user friendly message.  If nil is allowed, should do some safer type checking at https://github.com/uber-go/cadence-client/blob/master/internal/internal_activity.go#L204.

yiminc:
fix #360


yiminc:
nil will work for func as well by this change. However, if activity's argument is func and the value is not nil, the encoding part will fail because encoder cannot encode func.

PR #369, opened 2018-01-26
adding mock clients

yiminc:
mockery is not able to generate mock for type alias.
we generate mock and put it in internal package, then define type alias to expose it to public.


yiminc:
discard this PR in favor of https://github.com/uber-go/cadence-client/pull/370

PR #370, opened 2018-01-26
redefine Client/DomainClient as public interface

yiminc:
also generate mocks and modified the mock to use public alias types.

if we take this change, then will discard https://github.com/uber-go/cadence-client/pull/369


internal/workflow_testsuite.go
```diff
@@ -105,6 +105,7 @@ func (t *TestActivityEnvironment) ExecuteActivity(activityFn interface{}, args .
 
 // SetWorkerOptions sets the WorkerOptions that will be use by TestActivityEnvironment. TestActivityEnvironment will
 // use options of Identity, MetricsScope and BackgroundActivityContext on the WorkerOptions. Other options are ignored.
+// Note: WorkerOptions is defined in internal package, you can use public type worker.Options instead.
```
mfateev: I would change "you can use" to "use"

PR #371, opened 2018-01-26
fix potential nil pointer error

yiminc:
fix potential nil pointer bug

### DIFF
--- a/internal/internal_activity.go
+++ b/internal/internal_activity.go
@@ -175,7 +175,7 @@ func getValidatedLocalActivityOptions(ctx Context) (*executeLocalActivityParams,
 
 func validateFunctionArgs(f interface{}, args []interface{}, isWorkflow bool) error {
 	fType := reflect.TypeOf(f)
-	if fType.Kind() != reflect.Func {
+	if fType == nil || fType.Kind() != reflect.Func {
 		return fmt.Errorf("Provided type: %v is not a function type", f)
 	}
 	fnName := getFunctionName(f)
@@ -215,7 +215,7 @@ func validateFunctionArgs(f interface{}, args []interface{}, isWorkflow bool) er
 func getValidatedActivityFunction(f interface{}, args []interface{}) (*ActivityType, []byte, error) {
 	fnName := ""
 	fType := reflect.TypeOf(f)
-	switch fType.Kind() {
+	switch getReflectTypeKind(fType) {
 	case reflect.String:
 		fnName = reflect.ValueOf(f).String()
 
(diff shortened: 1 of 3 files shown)


internal/internal_activity.go
```diff
@@ -175,7 +175,7 @@ func getValidatedLocalActivityOptions(ctx Context) (*executeLocalActivityParams,
 
 func validateFunctionArgs(f interface{}, args []interface{}, isWorkflow bool) error {
 	fType := reflect.TypeOf(f)
-	if fType.Kind() != reflect.Func {
+	if fType == nil || fType.Kind() != reflect.Func {
```
madhuravi: Curious when will this be nil? If f is nil? In that case, should we return an error? Is nil not a valid argument?

internal/internal_activity.go
```diff
@@ -240,9 +240,17 @@ func getValidatedActivityFunction(f interface{}, args []interface{}) (*ActivityT
 	return &ActivityType{Name: fnName}, input, nil
 }
 
+func getReflectTypeKind(fType reflect.Type) reflect.Kind {
```
madhuravi: you can just name this "kind", might be simpler to read. We can infer it is for reflect.Type from the function argument.

internal/internal_activity.go
yiminc: If you pass nil as activity in ExecuteLocalActivity or ExecuteLocalActivity.
I run into this accidentally in a unit test. I defined an dummy activity in the test that takes no parameter and returns nil. I should have call workflow.ExecuteLocalActivity(ctx, dummyActivity), but I used dummyActivity() which returns nil and did workflow.ExecuteLocalActivity(ctx, dummyActivity()). The workflow panic with nil pointer error but have no clue where the error comes from.

internal/internal_activity.go
yiminc: sure

internal/internal_activity.go
```diff
@@ -240,9 +240,17 @@ func getValidatedActivityFunction(f interface{}, args []interface{}) (*ActivityT
 	return &ActivityType{Name: fnName}, input, nil
 }
 
+func getKind(fType reflect.Type) reflect.Kind {
+	fTypeKind := reflect.Invalid
+	if fType != nil {
+		fTypeKind = fType.Kind()
+	}
+	return fTypeKind
+}
```
mfateev: What about:

```
  if fType == nil {
    return reflect.Invalid
  }
  return fType.Kind()
```

PR #373, opened 2018-01-27
Activity options fix for #372

mfateev:
Fix for #372. 
The problem was that activity and workflow options are not copied but passed by reference when With... methods are called. Which means overriding an option in a child context would update it in every context.


ryanwalls:
👍

PR #384, opened 2018-01-31
test framework support for SignalExternalWorkflow

yiminc:


internal/internal_workflow_testsuite.go
```diff
@@ -1189,25 +1202,58 @@ func (env *testWorkflowEnvironmentImpl) RequestCancelWorkflow(domainName, workfl
...
+	if childHandle, ok := env.runningWorkflows[workflowID]; ok {
+		// target workflow is a child
+		childEnv := childHandle.env
+		if childEnv.isTestCompleted {
+			// child already completed (NOTE: we have only one failed cause now)
+			err := fmt.Errorf("signal external workflow failed, %v", shared.SignalExternalWorkflowExecutionFailedCauseUnknownExternalWorkflowExecution)
+			callback(nil, err)
+		} else {
+			childEnv.signalHandler(signalName, input)
+			//childEnv.signalWorkflow(signalName, input)
```
vancexu: remove

internal/internal_workflow_testsuite_test.go
```diff
@@ -1247,3 +1247,124 @@ func (s *WorkflowTestSuiteUnitTest) Test_WorkflowLocalActivityWithMockAndListene
 	s.NoError(err)
 	s.Equal("hello mock", result)
 }
+
+func (s *WorkflowTestSuiteUnitTest) Test_SignalChildWorkflow() {
+	// This test will send signal from parent to child, and then child will send back signal to ack. No mock is needed.
+	signalName := "test-signal-name"
+	signalData := "test-signal-data"
+	childWorkflowFn := func(ctx Context, parentExec WorkflowExecution) (string, error) {
```
mfateev: Shouldn't parent ids be in the context?

internal/internal_workflow_testsuite_test.go
```diff
@@ -1247,3 +1247,124 @@ func (s *WorkflowTestSuiteUnitTest) Test_WorkflowLocalActivityWithMockAndListene
...
+		childFuture := ExecuteChildWorkflow(ctx, childWorkflowFn, GetWorkflowInfo(ctx).WorkflowExecution)
+
+		// wait until child starts
+		var childWE WorkflowExecution
+		if err := childFuture.GetChildWorkflowExecution().Get(ctx, &childWE); err != nil {
+			return err
+		}
+
+		// send signal to child workflow
+		signalFuture := SignalExternalWorkflow(ctx, childWE.ID, childWE.RunID, signalName, signalData)
```
mfateev: Let's add SignalChildWorkflow method to the childFuture. Internally it would wait for child start.
You can make it a separate PR.

internal/workflow_testsuite.go
```diff
@@ -190,6 +190,32 @@ func (t *TestWorkflowEnvironment) OnWorkflow(workflow interface{}, args ...inter
...
+// However, it does not know what to do if your tested workflow code is sending signal to external unknown workflows.
+// In that case, you will need to setup mock for those signal calls.
+// Some examples of how to setup mock:
+//
+// * mock for specific target workflow that matches specific signal name and signal data
+// 	 env.OnSignalExternalWorkflow("test-domain", "test-workflow-id1", "test-runid1", "test-signal", "test-data").Return(nil).Once()
+// * mock for anything and succeed the send
+// 	 env.OnSignalExternalWorkflow(mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Once()
+// * mock for anything and fail the send
+// 	 env.OnSignalExternalWorkflow(mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errors.New("unknown external workflow")).Once()
```
mfateev: Does it convert to an error that Cadence actually returns from Signal?

internal/internal_workflow_testsuite_test.go
yiminc: yes, already plan to do so.

internal/workflow_testsuite.go
yiminc: no, it directly returns what ever mock supplied. We don't have an error type defined for signal error. In production code we use fmt.Errorf(...) to create error for signal failure.

internal/internal_workflow_testsuite_test.go
yiminc: no, it is not there, server has that info, but it is not sending down in the DecisionTask.
Task created: https://github.com/uber/cadence/issues/540

yiminc: In the test environment, yes, all info are available. But in production environment, it is not.

PR #385, opened 2018-01-31
adding helper methods to check error type

yiminc:
i realize we defined errors in 2 packages: workflow package and cadence package. But I want to have those helper method defined in one place so people can easily find them.

### DIFF
--- a/error.go
+++ b/error.go
@@ -22,6 +22,7 @@ package cadence
 
 import (
 	"go.uber.org/cadence/internal"
+	"go.uber.org/cadence/workflow"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -43,3 +44,45 @@ func NewCustomError(reason string, details ...interface{}) *CustomError {
...
+}
+
+// IsPanicError return if the err is a PanicError
+func IsPanicError(err error) bool {
+	_, ok := err.(*workflow.PanicError)
+	return ok
+}
+
+// IsContinueAsNewError return if the err is a ContinueAsNewError
+func IsContinueAsNewError(err error) bool {
```
mfateev: this one is never received by user code. Why do we need it?

PR #386, opened 2018-01-31
TimeoutError.Details() no panic

yiminc:
Add HasDetails to 3 error types: TimeoutError, CustomError, CanceledError.
Also added HasValue/HasValues to encoded.Value and encoded.Values.

### DIFF
--- a/error.go
+++ b/error.go
@@ -32,6 +32,9 @@ type (
 	CanceledError = internal.CanceledError
 )
 
+// ErrNoData is returned when trying to extract strong typed data while there is no data available.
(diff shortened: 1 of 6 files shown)


error.go
```diff
@@ -32,6 +32,9 @@ type (
 	CanceledError = internal.CanceledError
 )
 
+// ErrNoData is returned when trying to extract strong typed data while there is no data available.
+var ErrNoData = internal.ErrNoData
```
mfateev: Do we need IssErrData method?

yiminc: i think not.

PR #387, opened 2018-01-31
send signal to child workflow from ChildWorkflowFuture

yiminc:
Add SignalChildWorkflow() to ChildWorkflowFuture interface.


internal/internal_workflow.go
```diff
@@ -329,6 +329,17 @@ func (f childWorkflowFutureImpl) GetChildWorkflowExecution() Future {
 	return f.executionFuture
 }
 
+func (f childWorkflowFutureImpl) SignalChildWorkflow(ctx Context, signalName string, data interface{}) Future {
+	var childExec WorkflowExecution
+	if err := f.GetChildWorkflowExecution().Get(ctx, &childExec); err != nil {
+		future, settable := NewFuture(ctx)
```
mfateev: Why just returning f.GetChildWorkflowExecution future doesn't work in this case?

yiminc: it should work, didn't think about that.

PR #389, opened 2018-02-01
shrink decision state machine collections

yiminc:
when decision state machine is in completed state, we no longer need it, so should remove to save memory

### DIFF
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -21,6 +21,7 @@
 package internal
 
 import (
+	"container/list"
 	"fmt"
 
 	s "go.uber.org/cadence/.gen/go/shared"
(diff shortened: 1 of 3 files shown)


internal/internal_decision_state_machine.go
```diff
@@ -917,22 +919,26 @@ func (h *decisionsHelper) handleChildWorkflowExecutionCanceled(workflowID string
...
 		if decision != nil {
 			result = append(result, decision)
 		}
-	}
 
-	if markAsSent {
-		h.handleDecisionsSent()
-	}
+		if markAsSent {
+			d.handleDecisionSent()
```
madhuravi: This seems like a change in behavior? Earlier all decisions the orderedDecisions were updated. Why is that?

yiminc: there is no behavior change. Previously it was done in 2 for loops (second one done in separate function), now they are done in one loop.

yiminc: @madhuravi , previously it is done out of the first for loop.

PR #395, opened 2018-02-02
Added missing fields into activity.Info

mfateev:


internal/internal_activity.go
```diff
@@ -90,14 +91,19 @@ type (
...
-		isLocalActivity   bool
+		taskToken          []byte
+		workflowExecution  WorkflowExecution
+		activityID         string
+		activityType       ActivityType
+		serviceInvoker     ServiceInvoker
+		logger             *zap.Logger
+		metricsScope       tally.Scope
+		isLocalActivity    bool
+		heartbeatTimeout   time.Duration
```
madhuravi: How about embedding activityInfo in here instead of duplicating the fields?

testsuite/activity.go
```diff
@@ -37,9 +37,10 @@ type ServiceInvoker = internal.ServiceInvoker
 func WithActivityTask(
```
yiminc: this function is not used anywhere, we should remove it.

internal/internal_public.go
```diff
@@ -58,7 +58,7 @@ type (
 		// - RespondActivityTaskCompletedRequest
 		// - RespondActivityTaskFailedRequest
 		// - RespondActivityTaskCancelRequest
-		Execute(task *s.PollForActivityTaskResponse) (interface{}, error)
+		Execute(taskList string, task *s.PollForActivityTaskResponse) (interface{}, error)
```
madhuravi: We can make tasklist also part of PollForActivityTaskResponse. All other params are in there, tasklist seems relevant as well.

mfateev: I think it is overkill as poller knows which task list it polls on.

internal/internal_activity.go
mfateev: I think it is mostly about layering. ActivityInfo doesn't exist at the level of this function.

madhuravi: Not sure what you mean? They're both in the same package, struct definitions can be embedded.

internal/internal_public.go
madhuravi: Makes sense.

Issue #396, opened 2018-02-05
SignalChildWorkflow and CancelChildWorkflow behavior definition


wxing1292:
currently SignalChildWorkflow and CancelChildWorkflow's behavior is to some degree misleading: workflow run ID is used for existing API call / decision, however, the run ID will not be refreshed if child workflow execution is continued-as-newed. One possible solution is to use the empty string and let server decide the current running one.

Issue #388, opened 2018-01-31
testsuite: implement cancel external non child workflow


wxing1292:
currently the testsuite only support cancel current workflow and child workflow, it would be good if we also support cancel external non child workflow

Issue #326, opened 2018-01-03
RequestCancelWorkflow should return error when request failed


vancexu:
Currently in `workflowEnvironmentImpl` RequestCancelWorkflow return nil, which leads to no way of knowing running status for client. Should return a `Future`.

PR #401, opened 2018-02-11
Fix client side cancellation && signal

wxing1292:
bugfix: child / external workflow signal / cancellation combination were not handled correctly if child does continue as new.
bugfix: child workflow cancellation does not validate input run ID, which can corrupt internal state machine.
separate child workflow cancellation logic from child workflow state machine logic.
bugfix: *childWorkflowFutureImpl should be use as function receiver, instead of childWorkflowFutureImpl
add RequestCancelChildWorkflow to ChildWorkflowFuture
remove child workflow state cancellation logic since this is deprecated
add testsuite test for workflow cancellation

server side change: https://github.com/uber/cadence/pull/544

solves: #396, #326, #388 and #405


internal/workflow.go
```diff
@@ -549,20 +554,47 @@ func Sleep(ctx Context, d time.Duration) (err error) {
...
+
+	resultCallback := func(result []byte, err error) {
+		settable.Set(result, err)
+	}
+
+	getWorkflowEnvironment(ctx).RequestCancelExternalWorkflow(
+		*options.domain,
+		workflowID,
+		runID,
+		false, // this means we are not limited to child workflow
```
yiminc: hard code? should it be the childWorkflowOnly argument?

internal/internal_workflow_testsuite.go
```diff
@@ -1296,10 +1300,7 @@ func (env *testWorkflowEnvironmentImpl) getActivityInfo(activityID, activityType
 
 func (env *testWorkflowEnvironmentImpl) cancelWorkflow() {
 	env.postCallback(func() {
-		// RequestCancelWorkflow needs to be run in main thread
-		env.RequestCancelWorkflow(env.workflowInfo.Domain,
-			env.workflowInfo.WorkflowExecution.ID,
-			env.workflowInfo.WorkflowExecution.RunID)
+		env.workflowCancelHandler()
```
yiminc: the old code would call onChildWorkflowCanceledListener() call

internal/internal_workflow_testsuite.go
```diff
@@ -1208,11 +1210,13 @@ func (env *testWorkflowEnvironmentImpl) RequestCancelWorkflow(domainName, workfl
 		childHandle.handled = true
 		childEnv := childHandle.env
 		childEnv.cancelWorkflow()
+		callback(nil, nil)
```
yiminc: this should be handled within childEnv.cancelWorkflow()

internal/internal_event_handlers.go
```diff
@@ -1003,3 +1028,13 @@ func (weh *workflowExecutionEventHandlerImpl) handleSignalExternalWorkflowExecut
 
 	return nil
 }
+
+// getCancellationID, helper function determine the ID used for workflow cancellation
+// this function exists for backward compatibility, i.e. prev generation of client use workflow ID as identifier,
+// and there is no control (being empty / missing), so we have to use the workflow ID
+func getCancellationID(control []byte, workflowID string) string {
```
yiminc: this might cause problem. for example, replay the history that was generated by the old version code, so the control will be empty from history, but the state machine is keyed with the cancellation id.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -734,12 +739,22 @@ struct UpdateDomainInfo {
   20: optional string ownerEmail
 }
 
+struct ClusterReplicationConfiguration {
+ 10: optional string clusterName
+}
+
+struct DomainReplicationConfiguration {
```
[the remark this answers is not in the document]
wxing1292: i only synced the idl file.

do you prefer me removing this ?

internal/internal_decision_state_machine.go
```diff
@@ -778,71 +748,49 @@ func (h *decisionsHelper) handleStartChildWorkflowExecutionFailed(workflowID str
...
-	decision, ok := h.decisions[makeDecisionID(decisionTypeChildWorkflow, workflowID)]
-	if ok {
-		// for child workflow
-		decision.handleCancelInitiatedEvent()
-	} else {
-		// for non-child external workflow
-		decision = h.getDecision(makeDecisionID(decisionTypeExternalWorkflow, workflowID))
-		decision.handleInitiatedEvent()
-	}
+func (h *decisionsHelper) handleRequestCancelExternalWorkflowExecutionInitiated(cancellationID string) {
```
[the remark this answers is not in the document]
wxing1292: already tested using canary, you are on the reviewer list

internal/workflow.go
wxing1292: there are 2 API, one is childWorkflowFuture.RequestCancelChildWorkflow
another is workflow.RequestCancelExternalWorkflow

this is RequestCancelExternalWorkflow, so hard coded to false

internal/internal_workflow_testsuite.go
wxing1292: let us wait on testsuite, i still have some issue dealing with it

internal/internal_workflow_testsuite.go
wxing1292: same

internal/internal_event_handlers.go
wxing1292: existing client use workflow ID for indexing.
new client use control for indexing.

existing client will always use workflow ID for indexing, because existing client do not use control at all. so existing client should not break.

new client will work if replay using events from a new client, since both are using control for indexing.

new client will still work if replay using events from a old client, because old client will not touch the control, leaving that empty, so new client will use workflow ID for indexing, same as old client.

wxing1292: one more thing, control will only be set by new client, and will be copy pasted by service
so as long as control is set on the decision (by new client), it will appear on all related events, if not (by the old client), it will not appear.

wxing1292: ok there might be a issue, let me check

wxing1292: i think it will be ok if in the reply mode, client initialize 2 cancellation state machine, one using workflow ID, one using cancellation ID. once the Cancellation Initiated Event is received, one of the above 2 state machine can be removed.

internal/internal_decision_state_machine.go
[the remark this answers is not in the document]
yiminc: your canary case is not what i worried. The case is: parent starts a child workflow, before child completed, someone use API to cancel the child workflow outside of decider code.

internal/workflow.go
yiminc: i don't get it. then why do you need the childWorkflowOnly pass in as parameter?

idl/github.com/uber/cadence/shared.thrift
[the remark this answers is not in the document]
yiminc: did not know that server side idl was already landed, it is ok to keep it. However, it is always recommended to keep your diff short and focus only one theme at a time.

internal/workflow.go
wxing1292: childWorkflowOnly is used on the server side for this case:
parent workflow want to cancel a workflow if and only if the target workflow is its child workflow.

internal/internal_workflow_testsuite.go
```diff
@@ -1228,6 +1264,16 @@ func (env *testWorkflowEnvironmentImpl) SignalExternalWorkflow(domainName, workf
 		return
 	}
 
+	// here we are cancelling a child workflow but we cannot find it
```
yiminc: update the comment to signal

internal/internal_decision_state_machine.go
```diff
@@ -472,55 +473,24 @@ func (d *childWorkflowDecisionStateMachine) getDecision() *s.Decision {
...
 func (d *childWorkflowDecisionStateMachine) cancel() {
-	switch d.state {
-	case decisionStateStarted:
-		d.moveState(decisionStateCanceledAfterStarted, eventCancel)
-	default:
-		d.decisionStateMachineBase.cancel()
-	}
+	// the child workflow state machine should be alterd when trying to
+	// cancel child workflow, should use requestCancelExternalWorkflow instead
+	panic("unsupported operation")
```
yiminc: what happen for below code:
childCtx, childCancel := workflow.WithCancel(ctx)
workflow.ExecuteChildWorkflow(childCtx, ChildWorkflow, ...)
childCancel()

in old behavior, it will cancel the state machine and not even send out decision to start child.
in this new behavior, it will schedule the child even if cancel is already requested, and then wait until it started then send decision to cancel it.

This is breaking change.

Could you double check with Max.
cc @mfateev

internal/internal_decision_state_machine.go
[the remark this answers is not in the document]
yiminc: I miss understood. There is no problem here.

internal/internal_workflow_testsuite.go
wxing1292: done

internal/internal_decision_state_machine.go
wxing1292: sure, i think i and maxim had a conversation about this, the conclusion was to record every events, so no action will be omitted

wxing1292: @yiminc  talked to @mfateev, the change of behavior is not a huge issue, the issue is actually, when doing replay, the new client will encounter missing events, causing the child workflow future Get() stuck.

internal/internal_decision_state_machine.go
```diff
@@ -778,71 +781,94 @@ func (h *decisionsHelper) handleStartChildWorkflowExecutionFailed(workflowID str
...
+		decision := h.getDecision(makeDecisionID(decisionTypeChildWorkflow, workflowID))
 		decision.cancel()
-		return true, decision
+		return decision
 	}
 
-	// this is for external non-child workflow
+	// sanity check that cancellation ID is set
+	if len(cancellationID) == 0 {
+		panic("cancellation on all possible workflow should use cancellation ID")
```
yiminc: change "all possible" to "external"

internal/internal_event_handlers.go
```diff
@@ -977,6 +978,55 @@ func (weh *workflowExecutionEventHandlerImpl) handleChildWorkflowExecutionTermin
...
+}
+
+func (weh *workflowExecutionEventHandlerImpl) handleRequestCancelExternalWorkflowExecutionFailed(event *m.HistoryEvent) error {
+	// For cancellation of child workflow only, we do not use cancellation ID
+	// for cancellation of all possible workflow, we have to use cancellation ID
+	attributes := event.RequestCancelExternalWorkflowExecutionFailedEventAttributes
+	workflowID := attributes.WorkflowExecution.GetWorkflowId()
+	cancellationID := string(attributes.Control)
+	decision := weh.decisionsHelper.handleRequestCancelExternalWorkflowExecutionFailed(attributes.GetInitiatedEventId(), workflowID)
+	if len(cancellationID) != 0 {
```
yiminc: it would be clearer if we could encapsulate this logic into one separate method. for example:
func isForChildWorkflow(event *m.HistoryEvent) bool, and change all the places where you do this check to that method call.

internal/internal_decision_state_machine.go
wxing1292: sure

internal/internal_event_handlers.go
wxing1292: sure

internal/internal_decision_state_machine.go
```diff
@@ -778,71 +781,94 @@ func (h *decisionsHelper) handleStartChildWorkflowExecutionFailed(workflowID str
...
-		return true, decision
+		return decision
 	}
 
-	// this is for external non-child workflow
+	// sanity check that cancellation ID is set
+	if len(cancellationID) == 0 {
+		panic("cancellation on external workflow should use cancellation ID")
+	}
+	// this is aimed for any external workflow (including child workflow)
```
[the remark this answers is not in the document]
wxing1292: ok sure

PR #404, opened 2018-02-13
disconnected context

yiminc:
also manually tested with real workflow against local server and it works.

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -186,6 +186,14 @@ func WithCancel(parent Context) (ctx Context, cancel CancelFunc) {
 }
 
 // NewDisconnectedContext returns a new context that won't propagate parent's cancellation to the new child context.
+// One common use case is to do cleanup work after workflow is cancelled.
+//  err := workflow.ExecuteActivity(ctx, ActivityFoo).Get(ctx, &activityFooResult)
+//  if err != nil && cadence.IsCanceledError(ctx.Err()) {
+//    // activity failed, and workflow context is canceled
+//    disconnectedCtx, _ := workflow.newDisconnectedContext(ctx);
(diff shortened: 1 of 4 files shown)


internal/context.go
```diff
@@ -185,6 +185,12 @@ func WithCancel(parent Context) (ctx Context, cancel CancelFunc) {
 	return c, func() { c.cancel(true, ErrCanceled) }
 }
 
+// NewDisconnectedContext returns a new context that won't propagate parent's cancellation to the new child context.
```
mfateev: Would you add the following sample:

```
       err := workflow.ExecuteActivity(ctx, activitity1).Get(ctx, &activity1Result)
	if err != nil && cadence.IsCanceledError(err) {
                disconnectedCtx, _ := workflow.newDisconnectedContext(ctx);
		workflow.ExecuteActivity(disconnectedCtx, handleCancellation)
		return err
	}
```

mfateev: I think in this particular case it is equivalent, but uglier.

yiminc: There is slightly different. Activity could voluntarily return CanceledError. In that case, workflow context is not canceled.

yiminc: Also the workflow needs to wait until the handleCancellation activity returns, otherwise workflow returns CanceledError that would close the workflow and no activity would be scheduled..

ryanwalls:
@yiminc Should this work with `ExecuteLocalActivity` as well?

mfateev:
Yes, as well as with all other blocking operations like Workflow.sleep.

ryanwalls:
@mfateev @yiminc Thanks again for the quick fix.  Just an update, this is working great for us.

PR #406, opened 2018-02-14
add more description about timer

wxing1292:


workflow/deterministic_wrappers.go
```diff
@@ -106,6 +106,18 @@ func Now(ctx Context) time.Time {
 // is canceled, the returned Future become ready, and Future.Get() will return *CanceledError.
```
yiminc: we already mention the cancelled case here.

yiminc:
I'm not sure this change is needed. The doc already mention that the Future could return *CanceledError. It is not a big deal if you do not check the timer's return error and proceed, because your next operation will also return CanceledError.

workflow/deterministic_wrappers.go
wxing1292: i believe by documenting some common pitfalls, developers will have less chance wonder what happened.

PR #407, opened 2018-02-20
expose encoded values

yiminc:
User could use cadenceClient service API to get the closing event of a given workflow, however, that would only give you the blob data of the closing event. We need a way to convert that blob to strong typed values for user to consume.

Move the interface definition to internal to avoid circular dependency.


mfateev:
I don't understand the need for this change. The encoded package contains only interfaces. So I'm not sure how it can have circular dependency.

PR #409, opened 2018-02-22
non-deterministic check for signal external workflow

yiminc:
the non-deterministic check does not check for DecisionTypeSignalExternalWorkflowExecution so any code that send signal will fail in replay. This fix that.  Also relax the check for RequestCancelExternalWorkflowExecution to not check on the RunID.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -914,7 +914,8 @@ func isDecisionMatchEvent(d *s.Decision, e *s.HistoryEvent, strictMode bool) boo
 		}
 		eventAttributes := e.SignalExternalWorkflowExecutionInitiatedEventAttributes
 		decisionAttributes := d.SignalExternalWorkflowExecutionDecisionAttributes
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -902,8 +902,20 @@ func isDecisionMatchEvent(d *s.Decision, e *s.HistoryEvent, strictMode bool) boo
...
+
+		return true
+
+	case s.DecisionTypeSignalExternalWorkflowExecution:
+		if e.GetEventType() != s.EventTypeSignalExternalWorkflowExecutionInitiated {
+			return false
+		}
+		eventAttributes := e.SignalExternalWorkflowExecutionInitiatedEventAttributes
+		decisionAttributes := d.SignalExternalWorkflowExecutionDecisionAttributes
+		if eventAttributes.GetSignalName() != decisionAttributes.GetSignalName() ||
```
vancexu: probably also add domain check here?

PR #412, opened 2018-02-23
clear marker state machine

yiminc:
Fix one more bug discovered during local activity stress test.
I run test workflow that receive 1000 signals, each signal generate 100 local activities. As the workflow making progress, the perf degraded that it takes longer and longer to process each signal. (initially, it took about 10ms to process 100 local activities.)

We use marker state machine to generate decision, we are not drive marker state machine to completed state by history events. Once marker decision is sent, the markerStateMachine is considered as completed.

For SideEffect/Version markers, when the history event is applied, there is no marker decision state machine yet because we preload those marker events.
For local activity, when we apply the history event, we use it to create the marker state machine, there is no other event to drive it to completed state. 

So for marker state machine, the sent state should be treated as the completed state and it should be removed once it is in sent state.

### DIFF
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -618,13 +618,15 @@ func (d *signalExternalWorkflowDecisionStateMachine) handleCompletionEvent() {
 	}
 }
 
-func (d *markerDecisionStateMachine) handleCompletionEvent() {
-	// Marker decision transit from SENT to COMPLETED on EventType_MarkerRecorded event
+func (d *markerDecisionStateMachine) handleDecisionSent() {
+	// Marker decision state machine is considered as completed once decision is sent.
+	// For SideEffect/Version markers, when the history event is applied, there is no marker decision state machine yet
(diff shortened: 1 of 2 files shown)


internal/internal_decision_state_machine.go
```diff
@@ -618,13 +618,15 @@ func (d *signalExternalWorkflowDecisionStateMachine) handleCompletionEvent() {
 	}
 }
 
-func (d *markerDecisionStateMachine) handleCompletionEvent() {
-	// Marker decision transit from SENT to COMPLETED on EventType_MarkerRecorded event
+func (d *markerDecisionStateMachine) handleDecisionSent() {
```
mfateev: Wouldn't introducing another type of marker state machine for this use case make it more clear?

yiminc: All 3 maker use cases (SideEffect/Version/LocalActivity) are all the same, and this change serves them well.  Would add new type when needed in future.

PR #414, opened 2018-03-06
Add tags to metrics

vancexu:


internal/internal_logging_tags.go
```diff
@@ -36,3 +36,7 @@ const (
 	tagSideEffectID    = "SideEffectID"
 	tagChildWorkflowID = "ChildWorkflowID"
 )
+
+func newTag(tagName, tagValue string) map[string]string {
```
yiminc: use tagScope() helper method.

internal/internal_worker.go
```diff
@@ -1020,6 +1020,7 @@ func newAggregatedWorker(
 
 	ensureRequiredParams(&workerParams)
 	workerParams.MetricsScope = tagScope(workerParams.MetricsScope, tagDomain, domain)
+	workerParams.MetricsScope = tagScope(workerParams.MetricsScope, tagTaskList, taskList)
```
yiminc: could you update tagScope to take variable inputs so we only call tagScope once and only create one map.

internal/internal_logging_tags.go
vancexu: Done

internal/internal_worker.go
vancexu: Done

internal/internal_workflow_client.go
```diff
@@ -190,7 +190,8 @@ func (wc *workflowClient) StartWorkflow(
 	}
 
 	if wc.metricsScope != nil {
-		wc.metricsScope.Counter(metrics.WorkflowStartCounter).Inc(1)
+		tagScope(wc.metricsScope, tagWorkflowType, workflowType.Name).
```
yiminc: this will only include that tag for this specific metric. We want to include those tags for all metrics. But if you do the same thing for every metric, you are going to create a log of tagged subscope, which is not a cheap operation.

internal/internal_workflow_client.go
```diff
@@ -48,10 +48,11 @@ const (
...
 	// workflowClient is the client for starting a workflow execution.
 	workflowClient struct {
-		workflowService workflowserviceclient.Interface
-		domain          string
-		metricsScope    tally.Scope
-		identity        string
+		workflowService   workflowserviceclient.Interface
+		domain            string
+		metricsScope      tally.Scope
+		tagToMetricsScope map[string]tally.Scope // to avoid repeated creation of subScope
```
yiminc: this need to handle multi-thread concurrent access issue.

internal/internal_task_handlers.go
```diff
@@ -1247,7 +1249,11 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 			ath.logger.Error("Activity panic.",
 				zap.String("PanicError", fmt.Sprintf("%v", p)),
 				zap.String("PanicStack", st))
-			ath.metricsScope.Counter(metrics.ActivityTaskPanicCounter).Inc(1)
+			activityTypeName := t.ActivityType.GetName()
+			if _, ok := ath.tagToMetricsScope[activityTypeName]; !ok {
```
yiminc: extract these code so we don't have to dup this logic in 3 places.

internal/internal_task_handlers.go
```diff
@@ -113,14 +113,15 @@ type (
...
-		metricsScope     tally.Scope
-		logger           *zap.Logger
-		userContext      context.Context
-		hostEnv          *hostEnvImpl
-		activityProvider activityProvider
+		taskListName      string
+		identity          string
+		service           workflowserviceclient.Interface
+		metricsScope      tally.Scope
+		tagToMetricsScope map[string]tally.Scope // to avoid repeated creation of subScope
```
yiminc: type scopeFactory struct {
  tagToScopes map[string]tally.Scope
}

func (f *scopeFactory) getScope(tag string) Scope {
 //...
}

vancexu:
@yiminc all comments addressed. let me know if you have more concerns thanks.

internal/internal_worker.go
```diff
@@ -1073,14 +1073,38 @@ func newAggregatedWorker(
...
+	}
 	return metricsScope.Tagged(tagsMap)
 }
 
+// getOrCreateTaggedScope is used to get or create single tag for metrics.
+// Note: This assume the map has no conflicts of tagValue for different tagName,
+// otherwise retrieved subScope will be corrupted. e.g you cannot getOrCreateTaggedScope(m, scope, "domain", "A name")
+// and then getOrCreateTaggedScope(m, scope, "workflow-type", "A name")
+func getOrCreateTaggedScope(m sync.Map, scope tally.Scope, tagName, tagValue string) tally.Scope {
+	taggedScope, ok := m.Load(tagValue)
```
yiminc: can you make the key be tagName+tagValue, then you don't have to worry about the case you put in your comment.

internal/client.go
```diff
@@ -326,7 +327,7 @@ func NewClient(service workflowserviceclient.Interface, domain string, options *
 	return &workflowClient{
 		workflowService: metrics.NewWorkflowServiceWrapper(service, metricScope),
 		domain:          domain,
-		metricsScope:    metricScope,
+		metricsScope:    &metrics.TaggedScope{Scope: metricScope, Map: &sync.Map{}},
```
yiminc: create a func to create this because who using this taggedScope should not need to know that it needs a sync.Map internally:
func newTaggedScope(scope tally.Scope) *TaggedScope {
  return &metrics.TaggedScope{Scope: scope, Map: &sync.Map{}}
}

internal/common/metrics/scope.go
```diff
@@ -186,3 +193,27 @@ func (s *replayAwareScope) SubScope(name string) tally.Scope {
...
 	return s.scope.Capabilities()
 }
+
+// GetTaggedScope return a scope with tag
+func (ts *TaggedScope) GetTaggedScope(tagName, tagValue string) tally.Scope {
+	if ts.Map == nil {
+		ts.Map = &sync.Map{}
+	}
+
+	key := tagName + tagValue // used to prevent collision of tagValue (map key) for different tagName
```
yiminc: add a separater here: key := tagName + ":" + tagValue.

internal/common/metrics/scope.go
```diff
@@ -186,3 +193,27 @@ func (s *replayAwareScope) SubScope(name string) tally.Scope {
...
+	}
+	if taggedScope == nil {
+		panic("metric scope cannot be tagged") // This should never happen
+	}
+
+	return taggedScope.(tally.Scope)
+}
+
+// NewTaggedScope create a new TaggedScope
+func NewTaggedScope() *TaggedScope {
```
yiminc: let this function take a scope as input parameter

internal/internal_task_handlers.go
```diff
@@ -1089,7 +1089,7 @@ func newActivityTaskHandlerWithCustomProvider(
 		identity:         params.Identity,
 		service:          service,
 		logger:           params.Logger,
-		metricsScope:     params.MetricsScope,
+		metricsScope:     &metrics.TaggedScope{Scope: params.MetricsScope, Map: &sync.Map{}},
```
yiminc: use that newTaggedScope(scope) method

internal/internal_workflow_testsuite.go
```diff
@@ -212,7 +213,7 @@ func newTestWorkflowEnvironmentImpl(s *WorkflowTestSuite) *testWorkflowEnvironme
 		env.logger = logger
 	}
 	if env.metricsScope == nil {
-		env.metricsScope = tally.NoopScope
+		env.metricsScope = &metrics.TaggedScope{Scope: s.scope, Map: &sync.Map{}}
```
yiminc: use that newTaggedScope(scope) method

internal/internal_workflow_client_test.go
```diff
@@ -190,7 +192,7 @@ func (s *workflowRunSuite) SetupTest() {
 	mockCtrl := gomock.NewController(s.T())
 	s.workflowServiceClient = workflowservicetest.NewMockClient(mockCtrl)
 
-	metricsScope := tally.NoopScope
+	metricsScope := &metrics.TaggedScope{Scope: tally.NoopScope, Map: &sync.Map{}}
```
yiminc: use the newTaggedScope(scope) function

PR #428, opened 2018-03-31
side effect with ID

yiminc:
Public API is UpdateSideEffect(). Not super happy about that name. Please comment if you have better idea. Also need help to have better doc for that public API.

Besides that, I have manually tested with local cadence server with workflow worker disable stickiness to make sure that it worked in replay mode and also verified history events that only record new marker when value is changed.

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/uber-go/tally"
 	m "go.uber.org/cadence/.gen/go/shared"
+	"go.uber.org/cadence/encoded"
 	"go.uber.org/cadence/internal/common"
 	"go.uber.org/cadence/internal/common/metrics"
 	"go.uber.org/zap"
@@ -86,6 +87,7 @@ type (
 		sideEffectResult map[int32][]byte
 		changeVersions   map[string]Version
 		pendingLaTasks   map[string]*localActivityTask
+		idedSideEffect   map[string][]byte
 		unstartedLaTasks map[string]struct{}
 
 		counterID         int32     // To generate sequence IDs for activity/timer etc.
@@ -162,6 +164,7 @@ func newWorkflowExecutionEventHandler(
 		workflowInfo:          workflowInfo,
 		decisionsHelper:       newDecisionsHelper(),
 		sideEffectResult:      make(map[int32][]byte),
+		idedSideEffect:        make(map[string][]byte),
 		changeVersions:        make(map[string]Version),
 		pendingLaTasks:        make(map[string]*localActivityTask),
 		unstartedLaTasks:      make(map[string]struct{}),
@@ -512,6 +515,46 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() ([]byte, error), callback
 	wc.logger.Debug("SideEffect Marker added", zap.Int32(tagSideEffectID, sideEffectID))
 }
 
+func (wc *workflowEnvironmentImpl) UpdateSideEffect(id string, f func() ([]byte, error), equals func(a, b encoded.Value) bool) encoded.Value {
+	if result, ok := wc.idedSideEffect[id]; ok {
+		encodedResult := EncodedValue(result)
+		if wc.isReplay {
+			return encodedResult
+		}
+
+		newValue, err := f()
+		if err != nil {
+			panic(err)
+		}
+		if equals(EncodedValue(newValue), encodedResult) {
+			return encodedResult
+		}
+
+		return wc.recordUpdatedSideEffect(id, newValue)
+	}
+
+	if wc.isReplay {
+		// this should not happen
+		panic("SideEffect result with given ID not found during replay")
+	}
+
+	result, err := f()
+	if err != nil {
+		panic(err)
+	}
+	return wc.recordUpdatedSideEffect(id, result)
+}
+
+func (wc *workflowEnvironmentImpl) recordUpdatedSideEffect(id string, data []byte) encoded.Value {
+	details, err := wc.hostEnv.encodeArgs([]interface{}{id, string(data)})
+	if err != nil {
+		panic(err)
+	}
+	wc.decisionsHelper.recordUpdatedSideEffectMarker(id, details)
+	wc.idedSideEffect[id] = data
+	return EncodedValue(data)
+}
+
 func (weh *workflowExecutionEventHandlerImpl) ProcessEvent(
 	event *m.HistoryEvent,
 	isReplay bool,
@@ -809,6 +852,12 @@ func (weh *workflowExecutionEventHandlerImpl) handleMarkerRecorded(
 		return nil
 	case localActivityMarkerName:
 		return weh.handleLocalActivityMarker(attributes.Details)
+	case idedSideEffectMarkerName:
+		var fixedID string
+		var result string
+		encodedValues.Get(&fixedID, &result)
+		weh.idedSideEffect[fixedID] = []byte(result)
+		return nil
 	default:
 		return fmt.Errorf("unknown marker name \"%v\" for eventID \"%v\"",
 			attributes.GetMarkerName(), eventID)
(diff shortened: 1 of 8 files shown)


workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+//   otherwise, it records the new value with the same id on history.
+//
+// Caution: do not use MutableSideEffect to modify closures, always retrieve result from MutableSideEffect's encoded
+// return value.
+//
+// The difference between MutableSideEffect() and SideEffect() is that every new SideEffect() call in non-replay will
+// result in a new marker been recorded on history. However, MutableSideEffect() only record a new marker if the value
+// changed. During replay, MutableSideEffect() will not execute the function again, but it will return the exact same
+// value as it was returning during the non-replay run.
+func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{}, equals func(a, b encoded.Value) bool) encoded.Value {
```
mfateev: I'm not sure if it is worth the optimization, but you could save the unnecessary  serialization of function result if equals function would have the following signature:

func(a interface{}, b encoded.Value) bool

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+//   if they are equal, it returns the value without recording a new one in history;
+//   otherwise, it records the new value with the same id on history.
+//
+// Caution: do not use MutableSideEffect to modify closures, always retrieve result from MutableSideEffect's encoded
+// return value.
+//
+// The difference between MutableSideEffect() and SideEffect() is that every new SideEffect() call in non-replay will
+// result in a new marker been recorded on history. However, MutableSideEffect() only record a new marker if the value
+// changed. During replay, MutableSideEffect() will not execute the function again, but it will return the exact same
+// value as it was returning during the non-replay run.
```
mfateev: May be add the explanation in which use cases (like watching dynamic config changes) this is useful.

workflow/workflow.go
mfateev: Actually I think you can even have func(a, b interface{}) as a has a type and it can be used to decode b.

internal/internal_event_handlers.go
```diff
@@ -512,6 +515,46 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() ([]byte, error), callback
...
+		}
+		if equals(EncodedValue(newValue), encodedResult) {
+			return encodedResult
+		}
+
+		return wc.recordMutableSideEffect(id, newValue)
+	}
+
+	if wc.isReplay {
+		// this should not happen
```
purplesusan: this should not happen -> This should not happen.

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
 	return encoded
 }
 
+// MutableSideEffect executes provided function once, then it look up history for value with the given id,
```
purplesusan: executes provided function -> executes the provided function
then it look up history for value -> then it looks up the history for the value
given id, if there -> given id. If there

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// if there is no existing value, then it record the function result as value with the given id on history;
```
purplesusan: record -> records
as value -> as a value

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// otherwise, it compare if the existing value from history has changed from the new function result by calling the
```
purplesusan: it compare if -> it compares whether

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// provided equals function,
```
purplesusan: equals function, if they -> equals function. If they

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+//   if they are equal, it returns the value without recording a new one in history;
+//   otherwise, it records the new value with the same id on history.
+//
+// Caution: do not use MutableSideEffect to modify closures, always retrieve result from MutableSideEffect's encoded
```
purplesusan: closures, always -> closures. Always
retrieve result -> retrieve the result

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// return value.
+//
+// The difference between MutableSideEffect() and SideEffect() is that every new SideEffect() call in non-replay will
+// result in a new marker been recorded on history. However, MutableSideEffect() only record a new marker if the value
```
purplesusan: been -> being
record -> records

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
 	return internal.SideEffect(ctx, f)
 }
 
+// MutableSideEffect executes provided function once, then it look up history for value with the given id,
```
purplesusan: executes provided function -> executes the provided function
look up history -> looks up the history
for value -> for the value
given id, if there -> given id. If there

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// if there is no existing value, then it record the function result as value with the given id on history;
```
purplesusan: record -> records
as value -> as a value

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// otherwise, it compare if the existing value from history has changed from the new function result by calling the
```
purplesusan: compare if -> compares whether

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// provided equals function,
```
purplesusan: equals function, if they are -> equals function. If they are

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+//   if they are equal, it returns the value without recording a new one in history;
+//   otherwise, it records the new value with the same id on history.
+//
+// Caution: do not use MutableSideEffect to modify closures, always retrieve result from MutableSideEffect's encoded
```
purplesusan: closures, always -> closures. Always
retrieve result -> retrieve the result

workflow/workflow.go
```diff
@@ -262,6 +262,24 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+// return value.
+//
+// The difference between MutableSideEffect() and SideEffect() is that every new SideEffect() call in non-replay will
+// result in a new marker been recorded on history. However, MutableSideEffect() only record a new marker if the value
```
purplesusan: been -> being
record -> records

internal/workflow.go
```diff
@@ -804,6 +804,28 @@ func SideEffect(ctx Context, f func(ctx Context) interface{}) encoded.Value {
...
+//
+// Caution: do not use MutableSideEffect to modify closures. Always retrieve result from MutableSideEffect's encoded
+// return value.
+//
+// The difference between MutableSideEffect() and SideEffect() is that every new SideEffect() call in non-replay will
+// result in a new marker being recorded on history. However, MutableSideEffect() only records a new marker if the value
+// changed. During replay, MutableSideEffect() will not execute the function again, but it will return the exact same
+// value as it was returning during the non-replay run.
+//
+// One good use case of MutableSideEffect() is to dynamic change config without breaking determinism.
```
mfateev: is to access dynamically changing config

PR #430, opened 2018-04-03
Actually detect the existing lint failures

Groxx:
Currently, `make lint` doesn't detect any problems, apparently because
`golint internal/activity.go` doesn't report any issues.
However, if you `golint internal` you get multiple results, and you get
the same if you `golint internal/activity.go internal/other_files.go`...

... as long as you have enough files.
Or whatever triggers it.  `golint $(ls internal/*.go)` catches the same
results as this set of changes (because it's doing the same thing).

Why `golint` fails like this, I have no idea.

The fix though is to call `golint` with a whitelisted set of go files,
per directory (on the assumption that every directory is a package).
Make doesn't make this particularly easy, but it's at least possible.

```
$ make lint
./internal/activity.go:186:9: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:48:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:66:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:85:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:103:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:121:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:136:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:149:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:179:8: should not use basic type untyped string as key in context.WithValue
./internal/internal_task_pollers.go:352:9: should not use basic type untyped string as key in context.WithValue
Found 10 lint suggestions; failing.
make: *** [lint] Error 1
```

It even retains the same "run everything, fail if any fail" behavior:
```diff
$ git diff
diff --git a/client/client.go b/client/client.go
index d2d4183..8240de8 100644
--- a/client/client.go
+++ b/client/client.go
@@ -53,8 +53,6 @@ type (
        // WorkflowIDReusePolicy defines workflow ID reuse behavior.
        WorkflowIDReusePolicy = internal.WorkflowIDReusePolicy

-       // Client is the client for starting and getting information about a workflow executions as well as
-       // completing activities asynchronously.
        Client interface {
                // StartWorkflow starts a workflow execution
                // The user can use this to start using a function or workflow type name.
```
results in
```
$ make lint
./client/client.go:56:2: exported type Client should have comment or be unexported
Found 1 lint suggestions; failing.
./internal/activity.go:186:9: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:48:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:66:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:85:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:103:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:121:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:136:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:149:8: should not use basic type untyped string as key in context.WithValue
./internal/activity_test.go:179:8: should not use basic type untyped string as key in context.WithValue
./internal/internal_task_pollers.go:352:9: should not use basic type untyped string as key in context.WithValue
Found 10 lint suggestions; failing.
make: *** [lint] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,13 +91,25 @@ cover: cover_profile
 cover_ci: cover_profile
 	goveralls -coverprofile=$(BUILD)/cover.out -service=travis-ci || echo -e "\x1b[31mCoveralls failed\x1b[m";
 
+# golint fails to report many lint failures if it is only given a single file
+# to work on at a time.  and we can't exclude files from its checks, so for
+# best results we need to give it a whitelist of every file in every package
+# that we want linted.
+#
+# so lint + this golint func works like:
+# - iterate over all dirs (outputs "./folder/")
+# - find .go files in a dir (via wildcard, so not recursively)
+# - filter to only files in LINT_SRC
+# - if it's not empty, run golint against the list
+define lint_if_present
+test -n "$1" && golint -set_exit_status $1
+endef
 
 lint:
-	@lintFail=0; for file in $(LINT_SRC); do \
-		golint -set_exit_status "$$file"; \
-		if [ $$? -eq 1 ]; then lintFail=1; fi; \
-	done; \
-	if [ $$lintFail -eq 1 ]; then exit 1; fi;
+	@$(foreach pkg,\
+		$(sort $(dir $(LINT_SRC))), \
+		$(call lint_if_present,$(filter $(wildcard $(pkg)*.go),$(LINT_SRC))) || ERR=1; \
+	) test -z "$$ERR" || exit 1
 	@OUTPUT=`gofmt -l $(ALL_SRC) 2>&1`; \
 	if [ "$$OUTPUT" ]; then \
 		echo "Run 'make fmt'. gofmt must be run on the following files:"; \


Groxx:
I'm not sure at the moment what should be done for the lint failures, but figured this is an effective way to display them :)

vancexu:
Hi @Groxx, thank you for the contribution!
You patch looks good, would you like to sign the CLA above, so that we can merge it?
BTW, I provide a fix of these lint issue https://github.com/uber-go/cadence-client/pull/431

Groxx:
Yep, signed :)  What should I do to pull in your lint fixes?  I'm not sure how to "stack" pull requests on top of each other.

PR #431, opened 2018-04-03
Fix lint issue from context.WithValue

vancexu:


internal/internal_activity.go
```diff
@@ -105,12 +105,15 @@ type (
...
+
+	// context.WithValue need this type instead of basic type string to avoid lint error
+	contextKey string
 )
 
 const (
-	activityEnvContextKey          = "activityEnv"
-	activityOptionsContextKey      = "activityOptions"
-	localActivityOptionsContextKey = "localActivityOptions"
+	activityEnvContextKey          contextKey = "activityEnv"
```
mfateev: Does this syntax makes activity and localActivity keys of contextKey type as well?

vancexu: nop, only activityEnvContextKey is of contextKey

vancexu: I changed their order to make it looks even safer

internal/internal_activity.go
```diff
@@ -105,12 +105,15 @@ type (
...
+	contextKey string
 )
 
 const (
-	activityEnvContextKey          = "activityEnv"
-	activityOptionsContextKey      = "activityOptions"
-	localActivityOptionsContextKey = "localActivityOptions"
+	activityOptionsContextKey                 = "activityOptions"
+	localActivityOptionsContextKey            = "localActivityOptions"
+	activityEnvContextKey          contextKey = "activityEnv"
```
yiminc: why not change for the other 2 as well?

Groxx:
why not change them all?  lint is warning about using "normal" key types for context because they're able to collide with other code (including by accident).

Issue #423, opened 2018-03-21
no way to read heartbeat details from DescribeWorkflowExecution


yiminc:
if pending activity has heartbeat, the result from DescribeWorkflowExecution will include that heartbeat as blob. There is no way to decode that blob into actual object type.

vancexu:
`client.NewValue()` and `NewValues()` can be used to read this.

 NewValue creates a new encoded.Value which can be used to decode binary data returned by Cadence.  For example:
 User had Activity.RecordHeartbeat(ctx, "my-heartbeat") and then got response from calling Client.DescribeWorkflowExecution.
 The response contains binary field PendingActivityInfo.HeartbeatDetails,
 which can be decoded by using:
   var result string // This need to be same type as the one passed to RecordHeartbeat
   NewValue(data).Get(&result)

PR #435, opened 2018-04-12
Add Decoder API in client

vancexu:
Related issue: https://github.com/uber-go/cadence-client/issues/423


encoded/encoded.go
```diff
@@ -38,4 +38,10 @@ type (
 		// Get extract the encoded values into strong typed value pointers.
 		Get(valuePtr ...interface{}) error
 	}
+
+	// Decoder is used to decode data
```
mfateev: I would not create a new interface as we already have Value and Values.
I would add
NewValue(data []byte) Value
NewValues(data[]) Values

vancexu: Good suggestion, changed.

internal/client.go
```diff
@@ -377,3 +377,15 @@ func (p WorkflowIDReusePolicy) toThriftPtr() *s.WorkflowIdReusePolicy {
 	}
 	return &policy
 }
+
+// NewValue creates a new encoded.Value
```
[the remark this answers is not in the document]
vancexu: Added

PR #437, opened 2018-04-16
Dump all registered activities when activity not found

### LINKED ISSUE #418: Cadence should dump all known activity types if receive a task with unknown activity type.
Cadence should dump all known activity types if receive a task with unknown activity type.
This can help customer who accidentally change the activity function name.

Similar functionality is already implemented on the workflow side

arthurguber:
This commit modifies task handler in the client SDK to dump all activities registered when a workflow executes an activity whose name is not found, for example, due to a typo when specifying the activity by name.

Fixes #418

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1240,7 +1240,8 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 	activityImplementation := ath.getActivity(activityType.GetName())
 	if activityImplementation == nil {
 		// Couldn't find the activity implementation.
-		return nil, fmt.Errorf("unable to find activityType=%v", activityType.GetName())
+		supported := strings.Join(ath.getRegisteredActivityNames(), ", ")
+		return nil, fmt.Errorf("unable to find activityType=%v. Supported types: [%v]", activityType.GetName(), supported)
 	}
 
 	// panic handler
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -1240,7 +1240,8 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 	activityImplementation := ath.getActivity(activityType.GetName())
 	if activityImplementation == nil {
 		// Couldn't find the activity implementation.
-		return nil, fmt.Errorf("unable to find activityType=%v", activityType.GetName())
+		supported := strings.Join(ath.hostEnv.getRegisteredAcvitityNames(), ", ")
```
mfateev: It is kind of weird that above ath.getActivity is called and getRegisteredActivityNames added to hostEnv instead of activityTaskHandlerImpl.

internal/internal_worker.go
```diff
@@ -574,6 +574,14 @@ func (th *hostEnvImpl) getRegisteredActivities() []activity {
 	return activities
 }
 
+func (th *hostEnvImpl) getRegisteredAcvitityNames() []string {
+	activityNames := make([]string, 0, len(th.activityFuncMap))
```
mfateev: As you are using slices, the more idiomatic way would be:

var activityNames []string
...
activityNames = append(activityNames, a.ActivityType().GetName())

return activityNames

or even:

func (th *hostEnvImpl) getRegisteredAcvitityNames() (activityNames []string) {
...
activityNames = append(activityNames, a.ActivityType().GetName())
...
}

internal/internal_task_handlers.go
arthurguber: fixed.

internal/internal_worker.go
arthurguber: fixed

PR #438, opened 2018-04-16
Un-pin zap, bump to a newer version to pull in zaptest logger

Groxx:
Mostly motivated by the addition of a testing.T.Logf-compatible logger
provided by zap 1.8.0, which allows quite a lot cleaner test-logs, especially
when they are run in parallel.
But generally: there's no reason for zap to be pinned to a *specific*
version like it was.  Ranges are far nicer for users of libraries.

---

I've been running through the tests, and discovered that logs are currently
either `zap.Noop`ed or just printed, which can make it hard to understand
which tests are producing which logs, especially if you run them in parallel.

So, this is the start of a hopefully long sequence of changes to tests.
The end goal is to parallelize as many as possible so the race detector
actually has something to detect (I've already found one), and to move
the tests off `testify.Suite`.  It's a terrible abstraction now that we have
`t.Run`, and it can't even handle running things in parallel.


Groxx:
bleh.  can't submit individual requests without blending the whole branch.
any recommendations / requests for how to structure these, to make them easy to review / merge?

PR #449, opened 2018-04-26
Refactor domain client update and describe for crossdc

vancexu:
**This is a break change**
I need to add `ReplicationConfiguration` to domain client update and describe API. 
Changed signature to high level request and respond, to ensure forward compatible.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -259,23 +259,22 @@ type (
 		//	- InternalServiceError
 		Register(ctx context.Context, request *s.RegisterDomainRequest) error
 
-		// Describe a domain. The domain has two part of information.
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -259,23 +259,22 @@ type (
 		//	- InternalServiceError
 		Register(ctx context.Context, request *s.RegisterDomainRequest) error
 
-		// Describe a domain. The domain has two part of information.
-		// DomainInfo - Which has Name, Status, Description, Owner Email.
+		// Describe a domain. The domain has 3 part of information
```
mfateev: I think you want to add dots at the end of the sentences.

PR #450, opened 2018-05-01
check nil before stop workflow worker

yiminc:
check nil before stop workflow worker


internal/internal_worker.go
```diff
@@ -962,7 +962,9 @@ func (aw *aggregatedWorker) Start() error {
 		}
```
mfateev: Why do we not skip worker start if there are not activities. And above when there are no workflows as well.

yiminc: In theory, it is possible to register workflow/activities after worker started.

PR #455, opened 2018-05-04
expose method to set sticky cache size

yiminc:
expose method to set sticky cache size

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -318,15 +318,33 @@ func newWorkflowTaskHandler(
 }
 
 // TODO: need a better eviction policy based on memory usage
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -348,15 +348,33 @@ func newWorkflowTaskHandler(
...
 
 // TODO: need a better eviction policy based on memory usage
-var workflowCache = cache.New(defaultStickyCacheSize, &cache.Options{
-	RemovedFunc: func(cachedEntity interface{}) {
-		wc := cachedEntity.(*workflowExecutionContext)
-		wc.onEviction()
-	},
-})
+var workflowCache cache.Cache
+var stickyCacheSize = defaultStickyCacheSize
```
mfateev: Doesn't it need locking as it is written and accessed from different threads?

PR #457, opened 2018-05-04
do not keep reference for the whole history during decision process

yiminc:
we page through the history events during decision processing, but all the events are kept in memory during this time window. 
This diff fix that.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -240,7 +240,7 @@ func (eh *history) getMoreEvents() (*s.History, error) {
 	return eh.workflowTask.historyIterator.GetNextPage()
 }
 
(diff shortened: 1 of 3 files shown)


internal/internal_task_handlers.go
```diff
@@ -274,51 +266,29 @@ OrderEvents:
 		switch event.GetEventType() {
 		case s.EventTypeDecisionTaskStarted:
 			if !eh.IsNextDecisionFailed() {
-				eh.currentIndex++ // Since we already processed the current event
-				decisionStartedEvent = event
+				eh.currentIndex++
+				reorderedEvents = append(reorderedEvents, event)
```
mfateev: May be rename "reorderedEvents" to something else as there is no more reordering.

PR #460, opened 2018-05-07
fix closure leak

yiminc:
For every ExecuteLocalActivity (same for activity/childWorkflow/timer), we use a new coroutine to handle cancellation of the context. The coroutine creates a new selector, adds receiver to the context's done channel and the future, then select on it. The Select() calls receiveAsyncImpl() on the context's done channel which adds a callback to the done channel's blockedReceives collection. The callback is a closure function created in Select() function, that closure has access to another closure function for the select case which is created in ExecuteLocalActivity. Even though the coroutine exits, the objects created within the coroutine as well as all objects created within ExecuteLocalActivity won't be re-claimed. Because the done channel is on the root context, which only get GCed when that workflow is closed, and the done channel holds reference to the closure functions which pins all the objects that those closure functions have access to. This causes the memory leak.


internal/internal_workflow.go
```diff
@@ -567,17 +580,35 @@ func (c *channelImpl) receiveAsyncImpl(callback receiveCallback) (v interface{},
 	return nil, false, true
 }
 
+func (c *channelImpl) removeReceiveCallback(callback *receiveCallback) {
```
mfateev: See the last update at https://vbauerster.github.io/2017/04/removing-items-from-a-slice-while-iterating-in-go/ for removing without allocating.

yiminc: we only remove one element, so the current way should already be the optimized one as it reuses the same underlying array.

PR #463, opened 2018-05-08
Add Data Converter to support custom serialization/deserialization

vancexu:
Users now can implement encoded.DataConverter and let Cadence use it to serialize/deserialize parameters that need to be sent over the wire.
DataConverter can be set in 3 places:

1. Worker.
When create worker, one can provide data converter in WorkerOptions. Then everything processed by that worker will use provided data converter.
2. Client.
When create client to perform operations, one can provide data converter in clientOptions. Then all operation will use provided converter to encode arguments.
3. Workflow/Activity. 
Inside workflow implementation, one can use WithWorkflowDataConverter to create context and pass to other API calls like ExecuteActivity.

### DIFF
--- a/internal/internal_activity.go
+++ b/internal/internal_activity.go
@@ -29,6 +29,7 @@ import (
 	"reflect"
 
 	"github.com/uber-go/tally"
+	"go.uber.org/cadence/encoded"
 	"go.uber.org/zap"
 	"time"
 )
@@ -68,6 +69,7 @@ type (
 		InputArgs                     []interface{}
 		ScheduleToCloseTimeoutSeconds int32
 		WorkflowInfo                  *WorkflowInfo
+		DataConverter                 encoded.DataConverter
 	}
 
 	// asyncActivityClient for requesting activity execution
@@ -104,6 +106,7 @@ type (
 		scheduledTimestamp time.Time
 		startedTimestamp   time.Time
 		taskList           string
+		dataConverter      encoded.DataConverter
 	}
 
 	// context.WithValue need this type instead of basic type string to avoid lint error
@@ -178,6 +181,9 @@ func getValidatedLocalActivityOptions(ctx Context) (*executeLocalActivityParams,
 	if p.ScheduleToCloseTimeoutSeconds <= 0 {
 		return nil, errors.New("missing or negative ScheduleToCloseTimeoutSeconds")
 	}
+	if p.DataConverter == nil {
+		p.DataConverter = newDefaultDataConverter()
+	}
 
 	return p, nil
 }
@@ -221,7 +227,7 @@ func validateFunctionArgs(f interface{}, args []interface{}, isWorkflow bool) er
 	return nil
 }
 
-func getValidatedActivityFunction(f interface{}, args []interface{}) (*ActivityType, []byte, error) {
+func getValidatedActivityFunction(f interface{}, args []interface{}, dataConverter encoded.DataConverter) (*ActivityType, []byte, error) {
 	fnName := ""
 	fType := reflect.TypeOf(f)
 	switch getKind(fType) {
@@ -242,7 +248,7 @@ func getValidatedActivityFunction(f interface{}, args []interface{}) (*ActivityT
 			"Invalid type 'f' parameter provided, it can be either activity function or name of the activity: %v", f)
 	}
 
-	input, err := getHostEnvironment().encodeArgs(args)
+	input, err := encodeArgs(dataConverter, args)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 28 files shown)


internal/internal_worker.go
```diff
@@ -786,24 +678,30 @@ argsLoop:
...
+		return newDefaultDataConverter().ToData(arg)
+	}
+	return dc.ToData(arg)
 }
 
 // decode single value(like return parameter).
-func (th *hostEnvImpl) decodeArg(data []byte, to interface{}) error {
-	return th.decode(data, []interface{}{to})
+func decodeArg(dc encoded.DataConverter, data []byte, to interface{}) error {
+	if dc == nil {
```
yiminc: could you have unit test to cover input with multiple parameters case

internal/workflow_testsuite.go
```diff
@@ -64,17 +71,40 @@ type (
...
-	return b != nil
+	return b.values != nil
+}
+
+// Get extract data from encoded data to desired value type. valuePtr is pointer to the actual value type.
+func (b ErrorDetailsValues) Get(valuePtr ...interface{}) error {
+	if !b.HasValues() {
+		return ErrNoData
+	}
+	for i, item := range b {
```
yiminc: maybe do a sanity check to make sure len(valuePtr) <= len(b) and then loop through valuePtr instead of b.

internal/internal_activity.go
```diff
@@ -311,32 +317,35 @@ func deSerializeFnResultFromFnType(fnType reflect.Type, result []byte, to interf
...
 			return err
 		}
 	}
 	return nil
 }
 
-func deSerializeFunctionResult(f interface{}, result []byte, to interface{}) error {
+func deSerializeFunctionResult(f interface{}, result []byte, to interface{}, dataConverter encoded.DataConverter) error {
 	fType := reflect.TypeOf(f)
+	if dataConverter == nil {
```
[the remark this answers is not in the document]
vancexu: This is actually on purpose, for test code that use this, I want people just pass nil instead of figuration out how to deal with dc. So prefer not change

internal/internal_worker.go
vancexu: added.

internal/internal_activity.go
```diff
@@ -68,6 +69,7 @@ type (
 		InputArgs                     []interface{}
 		ScheduleToCloseTimeoutSeconds int32
 		WorkflowInfo                  *WorkflowInfo
+		DataConverter                 encoded.DataConverter
```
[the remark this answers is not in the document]
vancexu: ok, thanks for remind here

internal/workflow_testsuite.go
vancexu: changed

internal/internal_activity.go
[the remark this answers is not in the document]
vancexu: Understand, that's what I already did. But to be in defensive way, I feel it is still better to have this check, especially in this function.

internal/internal_workflow_testsuite.go
```diff
@@ -282,6 +286,7 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(param
 	childEnv.parentEnv = env
 	childEnv.startedHandler = startedHandler
 	childEnv.testWorkflowEnvironmentShared = env.testWorkflowEnvironmentShared
+	childEnv.workerOptions = WorkerOptions{DataConverter: params.dataConverter}
```
yiminc: you need to copy workerOptions from parent to child, no just for DataConverter.

internal/internal_workflow_testsuite_test.go
```diff
@@ -392,6 +413,30 @@ func testWorkflowHello(ctx Context) (string, error) {
 	return result, nil
 }
 
+func testWorkflowHelloWithDataConverter(ctx Context) (string, error) {
```
yiminc: can you move this function as workflowFn := func() ... into the Test_ActivityMockFunction_WithDataConverter

internal/internal_worker.go
```diff
@@ -1267,3 +1170,42 @@ func getTestTags(ctx context.Context) map[string]map[string]string {
...
+	return &defaultDataConverter{}
+}
+
+func (dc *defaultDataConverter) ToData(r ...interface{}) ([]byte, error) {
+	if len(r) == 1 && isTypeByteSlice(reflect.TypeOf(r[0])) {
+		return r[0].([]byte), nil
+	}
+
+	var encoder encoding
+	if isUseThriftEncoding(r) {
```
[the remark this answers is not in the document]
vancexu: created https://github.com/uber-go/cadence-client/issues/479

internal/internal_event_handlers.go
```diff
@@ -504,7 +511,7 @@ func (wc *workflowEnvironmentImpl) SideEffect(f func() ([]byte, error), callback
 			callback(result, err)
 			return
 		}
-		details, err = wc.hostEnv.encodeArgs([]interface{}{sideEffectID, result})
+		details, err = encodeArgs(wc.GetDataConverter(), []interface{}{sideEffectID, result})
```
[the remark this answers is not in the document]
vancexu: yes, changed all to use GetDataConverter()

PR #468, opened 2018-05-14
runID optional for DescribeWorkflowExecution

arthurguber:
This commit documents that runID is an optional parameter for DescribeWorkflowExecution function.


client/client.go
```diff
@@ -234,6 +234,8 @@ type (
 		QueryWorkflow(ctx context.Context, workflowID string, runID string, queryType string, args ...interface{}) (encoded.Value, error)
 
 		// DescribeWorkflowExecution returns information about the specified workflow execution.
+		// runID parameter is optional.
```
mfateev: // - runID can be default(empty string). if empty string then it will pick the last execution of that workflow ID.
And update the comment above of QueryWorkflow to say "last execution"

PR #475, opened 2018-05-22
Add non-deterministic workflow option

### LINKED ISSUE #333: Ability to fail workflow on non-deterministic error
Currently we have some validation on the client to detect when workflow implementation is modified in non-deterministic manner.  Client behavior is to through a panic in the decider which should cause decision to fail with non-deterministic error.  The workflow stays stuck in this state until the workflow implementation is fixed.  
Customers are asking for capability to have an option to fail such workflow executions immediately rather than stuck in this loop.  This is useful for executions which are short lived and it is ok to fail them incase of non-deterministic errors.

arthurguber:
This commit adds an option to configure how worker execution should
handle detection of a non-deterministic workflow.

Currently, when non-determinism is detected in execution history, the
task handler on the client simply logs the error to console *without*
replying back to the cadence server, which leads the server to think
the decision task has timed out and thus reschedules a new decsion task.

The currently behavior is less than satisfactory for 2 reasons:
1. From server side there's no indication of the non-determinism error.
   It simply looks like a timeout.
2. From the client side, there's no recover option in that the client
   will just pull off the new rescheduled decision task and fail again
   and again.

This commit adds a new option to cadence workers to explicit fail the
workflow where a mismatched history has been detected. This should
yield a more robust debugging experience for customers whose workflows
fail due to non-determinism.

Resolves #333

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -98,15 +98,16 @@ type (
 
 	// workflowTaskHandlerImpl is the implementation of WorkflowTaskHandler
 	workflowTaskHandlerImpl struct {
-		domain                 string
-		metricsScope           tally.Scope
-		ppMgr                  pressurePointMgr
-		logger                 *zap.Logger
-		identity               string
-		enableLoggingInReplay  bool
-		disableStickyExecution bool
-		hostEnv                *hostEnvImpl
-		laTunnel               *localActivityTunnel
+		domain                         string
+		metricsScope                   tally.Scope
+		ppMgr                          pressurePointMgr
+		logger                         *zap.Logger
+		identity                       string
+		enableLoggingInReplay          bool
+		disableStickyExecution         bool
+		hostEnv                        *hostEnvImpl
+		laTunnel                       *localActivityTunnel
+		nonDeterministicWorkflowPolicy NonDeterministicWorkflowPolicy
 	}
 
 	activityProvider func(name string) activity
@@ -307,6 +308,7 @@ func newWorkflowTaskHandler(
 		enableLoggingInReplay:  params.EnableLoggingInReplay,
 		disableStickyExecution: params.DisableStickyExecution,
 		hostEnv:                hostEnv,
+		nonDeterministicWorkflowPolicy: params.NonDeterministicWorkflowPolicy,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


internal/internal_worker.go
```diff
@@ -141,6 +141,10 @@ type (
 		DisableStickyExecution bool
 
 		StickyScheduleToStartTimeout time.Duration
+
+		// MismatchedWorkflowPolicy is used for configuring how client's decision task handler deals with
+		// mismatched history events (presumably arising from non-deterministic workflow definitions).
+		MismatchedWorkflowPolicy MismatchedWorkflowHistoryPolicy
```
yiminc: rename to NonDeterministicWorkflowPolicy

internal/internal_worker.go
```diff
@@ -141,6 +141,10 @@ type (
 		DisableStickyExecution bool
 
 		StickyScheduleToStartTimeout time.Duration
+
+		// MismatchedWorkflowPolicy is used for configuring how client's decision task handler deals with
```
yiminc: is used for configuring how decision worker deals with non-deterministic history events (presumably arising from non-deterministic workflow definitions or non-backward compatible workflow definition changes).

internal/worker.go
```diff
@@ -137,9 +137,32 @@ type (
...
+type MismatchedWorkflowHistoryPolicy int
+
+const (
+	// MismatchedWorkflowHistoryPolicyIgnore is the default policy for handling detected non-determinism.
+	// This option simply logs to console with an error message that non-determinism is detected, but
+	// does *NOT* reply anything back to the server.
+	// It is chosen as default for backward compatibility reasons because it preserves the old behavior
+	// for handling non-determinism that we had before MismatchedWorkflowHistoryPolicy type was added to
+	// allow more configurability.
+	MismatchedWorkflowHistoryPolicyIgnore MismatchedWorkflowHistoryPolicy = iota
```
yiminc: Ignore is not clear, it is more like PauseWorkflow to user.

internal/internal_task_handlers.go
```diff
@@ -653,7 +655,26 @@ ProcessEvents:
 		// check if decisions from reply matches to the history events
 		if err := matchReplayWithHistory(replayDecisions, respondEvents); err != nil {
```
yiminc: I think there is another place where non-deterministic error could happen, which is in the state machine transition. I think we need to handle that as well.

arthurguber: As discussed, this will be fixed separately as there's some unknown corner cases.

internal/internal_task_handlers.go
```diff
@@ -653,7 +655,26 @@ ProcessEvents:
...
+			// will pick up the error and correctly wrap it in the response request we sent back
+			// to the server, which in this case will contain a request to fail the workflow.
+			//
+			// If we simply return the error, the decision task completion code path will not
+			// execute at all, therefore, no response is sent back to the server and we will
+			// look like a decision task time out.
+			switch wth.nonDeterministicWorkflowPolicy {
+			case NonDeterministicWorkflowPolicyFailWorkflow:
+				eventHandler.Complete(nil, NewCustomError("nondeterministic workflow", err.Error()))
+			case NonDeterministicWorkflowPolicyPauseWorkflow:
```
yiminc: Now I think block or stuck is better than Pause.

PR #482, opened 2018-05-25
Add tag ClientImpl to all metrics

vancexu:
We have multiple clients(GO+Java) now. We need to emit the client tag for us to distinguish calls from GO vs Java.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -334,7 +334,7 @@ func NewClient(service workflowserviceclient.Interface, domain string, options *
 	if options != nil {
 		metricScope = options.MetricsScope
 	}
-	metricScope = tagScope(metricScope, tagDomain, domain)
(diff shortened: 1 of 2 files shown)


internal/internal_logging_tags.go
```diff
@@ -36,3 +36,8 @@ const (
 	tagSideEffectID    = "SideEffectID"
 	tagChildWorkflowID = "ChildWorkflowID"
 )
+
+const (
+	tagClientImpl = "ClientImpl" // metrics tag for cadence client implementation
```
yiminc: use
 	// header that contains the client implementation
	clientImplHeaderName  = "cadence-client-name"
	clientImplHeaderValue = "uber-go"

yiminc: these const already defined. do not define new ones.

PR #496, opened 2018-06-04
Reset workflow stickiness on eviction

arthurguber:
This commit improves the workflow execution cache by having it
notify the server to reset a workflow's stickiness when that
workflow's execution context is evicted from the cache due to the
cache being full.

### DIFF
--- a/evictiontest/workflow_cache_eviction_test.go
+++ b/evictiontest/workflow_cache_eviction_test.go
@@ -1,0 +1,183 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// This test must be its own package because workflow execution cache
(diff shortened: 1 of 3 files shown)


evictiontest/workflow_cache_eviction_test.go
```diff
@@ -0,0 +1,178 @@
...
+		workflowID := common.StringPtr("testID" + strconv.Itoa(int(taskID)))
+		runID := common.StringPtr("runID" + strconv.Itoa(int(taskID)))
+		// how we initialize the response here is the result of a series of trial and error
+		// the goal is we want to fabricate a response that looks real enough to our worker
+		// that it will actually go along with processing it instead of just tossing it out
+		// after polling it or giving an error
+		ret := &m.PollForDecisionTaskResponse{
+			TaskToken:              make([]byte, 5),
+			WorkflowExecution:      &m.WorkflowExecution{WorkflowId: workflowID, RunId: runID},
+			WorkflowType:           &m.WorkflowType{Name: common.StringPtr("go.uber.org/cadence/internal.testReplayWorkflow")},
```
yiminc: the name should include package evictiontest? Another way is to register workflow with custom name so you don't have to use full name.

evictiontest/workflow_cache_eviction_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	resetStickyAPICalled := make(chan struct{})
+	mockResetStickyTaskList := func(ctx context.Context, _ResetRequest *m.ResetStickyTaskListRequest, opts ...yarpc.CallOption,
+	) (success *m.ResetStickyTaskListResponse, err error) {
+		resetStickyAPICalled <- struct{}{}
+		return &m.ResetStickyTaskListResponse{}, nil
+	}
+	// pick a random size between [5, 15), there's no significance to the range here
+	// other than it's not 0 or 1 and it's sufficiently small for the test to run reasonably quickly
+	cacheSize := rand.Intn(10) + 5
```
yiminc: let's just give a fixed number to avoid randomization.

evictiontest/workflow_cache_eviction_test.go
arthurguber: So actually in this instance it doesn't matter because workflow lookup happens after creating & inserting workflow cache and apparently we don't evict when workflow execution fails due to this.

evictiontest/workflow_cache_eviction_test.go
```diff
@@ -0,0 +1,178 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// This test must be its own package because workflow execution cache
+// is package level global, so any tests against it should belong to
+// its own package to avoid inter-test interference.
```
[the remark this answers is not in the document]
arthurguber: yep.

evictiontest/workflow_cache_eviction_test.go
yiminc: please file an issue to get this fixed.

PR #498, opened 2018-06-05
handle long running decision

yiminc:
add ability to handle long running decision task

### DIFF
--- a/internal/internal_public.go
+++ b/internal/internal_public.go
@@ -45,12 +45,13 @@ type (
 	// WorkflowTaskHandler represents decision task handlers.
 	WorkflowTaskHandler interface {
 		// Processes the workflow task
-		// The response is either
-		// RespondDecisionTaskCompletedRequest or RespondDecisionTaskFailedRequest
+		// The response could be:
+		// - RespondDecisionTaskCompletedRequest
+		// - RespondDecisionTaskFailedRequest
+		// - RespondQueryTaskCompletedRequest
 		ProcessWorkflowTask(
 			task *s.PollForDecisionTaskResponse,
-			historyIterator HistoryIterator,
-			emitStack bool) (response interface{}, stackTrace string, err error)
+			historyIterator HistoryIterator) (response interface{}, context *WorkflowExecutionContext, err error)
 	}
 
 	// ActivityTaskHandler represents activity task handlers.
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -226,24 +226,31 @@ func (wtp *workflowTaskPoller) processWorkflowTask(workflowTask *workflowTask) e
 		return nil
 	}
 
-	executionStartTime := time.Now()
+	startTime := time.Now()
 	// Process the task.
-	completedRequest, _, err := wtp.taskHandler.ProcessWorkflowTask(workflowTask.task, workflowTask.historyIterator, false)
+	completedRequest, wc, err := wtp.taskHandler.ProcessWorkflowTask(workflowTask.task, workflowTask.historyIterator)
 	if err != nil {
 		wtp.metricsScope.Counter(metrics.DecisionExecutionFailedCounter).Inc(1)
 		return err
 	}
 	if completedRequest == nil {
+		if wc != nil {
+			// decision task cannot complete, we need a timer
+			wtp.scheduleRespondDecisionTaskCompleted(wc, workflowTask, startTime)
+		}
+
 		return nil
 	}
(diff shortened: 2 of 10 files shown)


internal/internal_task_pollers.go
```diff
@@ -253,6 +260,90 @@ func (wtp *workflowTaskPoller) processResetStickinessTask(rst *resetStickinessTa
...
+	if response == nil || response.DecisionTask == nil {
+		return nil
+	}
+
+	startTime := time.Now()
+	newTask := wtp.toWorkflowTask(response.DecisionTask)
+
+	if newTask.task.History.Events[0].GetEventId() != w.previousStartedEventID+1 {
+		// sanity check for new task, this should not happen.
+		wtp.logger.Info("New task from RespondDecisionTaskCompleted has unexpected events",
```
mfateev: Why is it info then? Should be an error.

internal/internal_task_pollers.go
```diff
@@ -253,6 +260,90 @@ func (wtp *workflowTaskPoller) processResetStickinessTask(rst *resetStickinessTa
...
+		wtp.scheduleRespondDecisionTaskCompleted(w, newTask, startTime)
+		return nil
+	}
+
+	response, err = wtp.RespondTaskCompletedWithMetrics(completedRequest, newTask.task, startTime)
+	if err != nil {
+		return err
+	}
+
+	return w.processCompleteDecisionResponseWithLock(response, wtp)
```
mfateev: Do we have any limit on number of such recursive calls? Could it be that we run out of stack?

internal/internal_task_pollers.go
```diff
@@ -326,10 +430,9 @@ func (wtp *workflowTaskPoller) RespondTaskCompleted(completedRequest interface{}
 
 	if err != nil {
 		wtp.metricsScope.Counter(metrics.DecisionResponseFailedCounter).Inc(1)
```
mfateev: May be move into RespondTaskCompletedWithMetrics as well?

internal/internal_public.go
```diff
@@ -45,12 +45,13 @@ type (
...
-		// RespondDecisionTaskCompletedRequest or RespondDecisionTaskFailedRequest
+		// The response could be:
+		// - RespondDecisionTaskCompletedRequest
+		// - RespondDecisionTaskFailedRequest
+		// - RespondQueryTaskCompletedRequest
 		ProcessWorkflowTask(
 			task *s.PollForDecisionTaskResponse,
-			historyIterator HistoryIterator,
-			emitStack bool) (response interface{}, stackTrace string, err error)
+			historyIterator HistoryIterator) (response interface{}, context *WorkflowExecutionContext, err error)
```
mfateev: After this change WorkflowExecutionContext plays dual role. One for workflow implementation code, another for poller. I think returning an interface from this call with well defined set of methods would make it cleaner. It is OK if the current context does implement this interface. Then workflowExecutionContext might stay private.

internal/internal_task_pollers.go
```diff
@@ -253,6 +260,90 @@ func (wtp *workflowTaskPoller) processResetStickinessTask(rst *resetStickinessTa
...
+	response, err := wtp.RespondTaskCompletedWithMetrics(completeRequest, workflowTask.task, startTime)
+	if err != nil {
+		return
+	}
+
+	wc.processCompleteDecisionResponseWithLock(response, wtp)
+	return
+}
+
+func (w *WorkflowExecutionContext) processCompleteDecisionResponseWithLock(response *s.RespondDecisionTaskCompletedResponse, wtp *workflowTaskPoller) error {
```
mfateev: At google they had standard to call such functions ....Locked. It would make it processCompleteDecisionResponseLocked

internal/internal_task_pollers.go
yiminc: It only do recursive call when there is new events received on server while worker try to complete decision task. Go stack size limit is 1GB, I think it would be safe here.

internal/internal_task_pollers.go
yiminc: good to know, will update

internal/internal_task_pollers.go
yiminc: will update

internal/internal_task_pollers.go
yiminc: will update

internal/internal_public.go
yiminc: I can do a follow up refactoring change on this.

PR #499, opened 2018-06-05
Fix typo in NewCustomError

pupapaik:


pupapaik:
This helps people to find right way to complete activity with error. It took me several minutes until I reliazed right function.

PR #501, opened 2018-06-07
Refactor workflowExecutionContext

yiminc:
Define interface WorkflowExecutionContext and its implementation workflowExecutionContextImpl.

### DIFF
--- a/internal/internal_public.go
+++ b/internal/internal_public.go
@@ -28,6 +28,9 @@ package internal
 // point of view only to access them from other packages.
 
 import (
+	"sync"
+	"time"
+
 	s "go.uber.org/cadence/.gen/go/shared"
 )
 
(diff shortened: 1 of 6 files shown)


internal/internal_public.go
```diff
@@ -42,6 +45,18 @@ type (
 		HasNextPage() bool
 	}
 
+	// WorkflowExecutionContext represents one instance of workflow execution state in memory. Lock must be obtained before
+	// calling into any methods.
+	WorkflowExecutionContext interface {
+		sync.Locker
+		ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error)
+		ProcessLocalActivityResult(lar *localActivityResult) (interface{}, error)
+		CompleteDecisionTask(waitLocalActivity bool) interface{}
```
mfateev: Document waitLocalActivity parameter semantic

internal/internal_public.go
```diff
@@ -42,6 +45,18 @@ type (
...
+		GetDelayReportDuration(startTime time.Time) time.Duration
```
mfateev: Document semantic of the GetDelayReportDuration

internal/internal_task_handlers.go
```diff
@@ -621,24 +614,28 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
...
 
-	workflowContext, skipReplayCheck, err := wth.getOrCreateWorkflowContext(task, historyIterator)
+	workflowContext, err := wth.getOrCreateWorkflowContext(task, historyIterator)
 	if err != nil {
 		return nil, nil, err
 	}
-	defer workflowContext.release()
-
-	workflowTask := &workflowTask{task: task, historyIterator: historyIterator}
+	defer workflowContext.Unlock()
```
mfateev: Where is it locked in this method? Can code be refactored to make locking/unlocking pare visible?

internal/internal_public.go
yiminc: When i put doc for this method, I think I would rather expose a different method, say GetDecisionTimeout() and then it is the poller's decision to decide when to respond.

internal/internal_task_handlers.go
yiminc: The lock is obtained in the getOrCreate() method because there are some operations that happen in that method that need to be done with lock.

PR #513, opened 2018-06-26
Log metrics and do not panic when the channel receives a corrupted signal

halakaraki:


internal/internal_workflow.go
```diff
@@ -532,7 +547,11 @@ func (c *channelImpl) ReceiveAsync(valuePtr interface{}) (ok bool) {
 
 func (c *channelImpl) ReceiveAsyncWithMoreFlag(valuePtr interface{}) (ok bool, more bool) {
 	v, ok, more := c.receiveAsyncImpl(nil)
-	c.assignValue(v, valuePtr)
+	err:=c.assignValue(v, valuePtr)
+	if err != nil {
+		// failed so serialize corrupt signal. Dropping signal
+		return false, more
```
yiminc: instead of returning here, we should continue go get next signal until there is no more signal.

internal/internal_workflow.go
```diff
@@ -651,12 +670,14 @@ func (c *channelImpl) Close() {
...
 
-// Takes a value and assigns that 'to' value.
-func (c *channelImpl) assignValue(from interface{}, to interface{}) {
+// Takes a value and assigns that 'to' value. logs a metric if it is unable to deserialize
+func (c *channelImpl) assignValue(from interface{}, to interface{}) (error) {
 	err := decodeAndAssignValue(c.dataConverter, from, to)
+	//add to metrics
 	if err != nil {
-		panic(err)
+		c.scope.Counter(metrics.CorruptedSignalsCounter).Inc(1)
```
yiminc: ditto

internal/internal_workflow_test.go
```diff
@@ -348,8 +352,8 @@ func sayGreetingActivity(input *sayGreetingActivityRequest) (string, error) {
 func greetingsWorkflow(ctx Context) (result string, err error) {
 	// Get Greeting.
 	ao := ActivityOptions{
-		ScheduleToStartTimeout: 10 * time.Second,
-		StartToCloseTimeout:    5 * time.Second,
+		ScheduleToStartTimeout: 10000 * time.Second,
```
yiminc: why changing this?

internal/internal_workflow.go
halakaraki: This is in receiveAsync. I think it's not meant to be blocking

internal/internal_workflow.go
halakaraki: The waiting happens in Receive method. I think assignValue should not be concerned with blocking and should only return an error

internal/internal_workflow_test.go
halakaraki: Oops, will revert

internal/internal_workflow.go
halakaraki: resolved. Will keep consuming until channel is drained or a valid signal is received

internal/internal_workflow.go
halakaraki: resolved

Issue #497, opened 2018-06-05
Workflow definition not found should evict execution context from cache


arthurguber:
This bug is 100% internal(i.e. invisible to customers), nor does it have any functional incorrectness.

The only drawback of this "bug" is a tiny performance loss.

When a workflow task(for running a new workflow) is received, the task handler on the client side will create an execution context and put this context in the workflow cache.

Then, the task handler will attempt to start execution of this workflow. It does so by first looking up the workflow definition in the registered workflow map.

If the workflow definition is not found, the task handler will note the "unrecognized workflow" error and reply the error to the server. However, it does NOT evict out of the cache the execution context it created.

This leads to a stale cache entry wasting a cache slot, that will eventually be cleared out when the cache is full by virtue of LRU evicting old entries.

We'd like the task handler to evict this context from the cache pro-activively, i.e. as soon as an unfound workflow error is set.

PR #517, opened 2018-07-03
Evict execution context from cache when there is no workflow definition found

lihannan99:
Previously if there is error during **getOrCreateWorkflowContext**, the  local cache is not cleared. Once **set workflowContext.err** and **Unlock()** will evict the cache proactively.  

Issue:
https://github.com/uber-go/cadence-client/issues/497

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -616,6 +616,10 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 
 	workflowContext, err := wth.getOrCreateWorkflowContext(task, historyIterator)
 	if err != nil {
+		// The WorkflowContext will be removed from the cache if an error is found
+		// in workflowContext.err when calling Unlock() method.
+		workflowContext.err = err
+		defer workflowContext.Unlock()
 		return nil, nil, err
 	}
 	defer workflowContext.Unlock()
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -624,12 +624,11 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 		workflowContext.Unlock(err)
 	}()
 
-
 	response, err := workflowContext.ProcessWorkflowTask(task, historyIterator)
 	return response, workflowContext, err
 }
 
-func (w *workflowExecutionContextImpl)  ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error) {
+func (w *workflowExecutionContextImpl) ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error) {
 	if err = w.ResetIfStale(task, historyIterator); err != nil {
 		return
 	}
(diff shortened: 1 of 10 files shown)


internal/internal_task_handlers.go
```diff
@@ -616,6 +616,10 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 
 	workflowContext, err := wth.getOrCreateWorkflowContext(task, historyIterator)
 	if err != nil {
+		// The WorkflowContext will be removed from the cache if an error is found
+		// in workflowContext.err when calling Unlock() method.
+		workflowContext.err = err
```
yiminc: Have you tested this? I don't think this would work. Even if the workflow type is not defined, the getOrCreateWorkflowContext() won't return error. The error happen when it try to process the WorkflowExecutionStartedEvent.
Also, we should not try to set workflowContext.err directly.

internal/internal_task_handlers.go
```diff
@@ -616,6 +616,10 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
...
+		defer workflowContext.Unlock()
```
yiminc: I think here is what we need to update. Change it to:
defer func() { workflowContext.Unlock(err); }
and evict the context within the Unlock method when err is not nil. You also need to change the signature of Unlock().

internal/internal_task_handlers.go
lihannan99: yeah, I was somehow thought the error comes from getOrCreateWorkflowContext and mocked an error and tested on that. 

Gotcha, I won't set the workflowContext.err directly.

internal/internal_task_handlers.go
lihannan99: will do. 
Thanks!

internal/internal_task_handlers.go
```diff
@@ -618,13 +619,17 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
...
+		workflowContext.Unlock(err)
+	}()
+
 
 	response, err := workflowContext.ProcessWorkflowTask(task, historyIterator)
 	return response, workflowContext, err
 }
 
-func (w *workflowExecutionContextImpl) ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error) {
+func (w *workflowExecutionContextImpl)  ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error) {
```
vancexu: Could you run `make fmt`? This is not formatted.

internal/internal_task_handlers_test.go
```diff
@@ -389,6 +390,8 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_NondeterministicDetection() {
 		NonDeterministicWorkflowPolicy: NonDeterministicWorkflowPolicyBlockWorkflow,
 	}
 	taskHandler := newWorkflowTaskHandler(testDomain, params, nil, getHostEnvironment())
+	newWorkflowTaskWorkerInternal(taskHandler, t.service,testDomain, params)
```
vancexu: Will this do something? Could you add a unit test for your change?

internal/internal_task_handlers.go
lihannan99: sure. Will do

internal/internal_task_handlers_test.go
lihannan99: This will set the laTunnel for the taskHandler, without it, the tests will end up panic as taskHandler.laTunnel is nil 

I will try to see if I can add some unit tests for it.

PR #521, opened 2018-07-10
Add an AfterFn mock-delay-returning function

Groxx:
When combined with a func provided to Return, this can be used to
simulate dynamic / open-ended respons and delay sequences on activity
and workflow functions.

This is particularly useful for running tests with randomized behavior,
as now another important piece of behavior can be varied.  Previously
only the number of calls and the return value could be changed easily.

---

Since the MockCallWrapper (and testify's mock.Call) don't appear to validate
that e.g. "Return" isn't called multiple times, I didn't add an error / panic to
After or AfterFn to prevent combining them.
I'd be happy to panic on misuse / make other changes if you'd prefer!
Tests are also fairly simple at this point, in part because there are few helpers,
and to do a lot better would mean a lot of copypasta.  I can also do that though,
and mirror all the After tests / add extras if you'd like.

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -66,7 +66,7 @@ type (
 		env  *TestWorkflowEnvironment
 
 		runFn        func(args mock.Arguments)
-		waitDuration time.Duration
+		waitDuration func() time.Duration
 	}
 )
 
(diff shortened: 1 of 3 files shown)


internal/workflow_testsuite.go
```diff
@@ -65,8 +65,9 @@ type (
 		call *mock.Call
 		env  *TestWorkflowEnvironment
 
-		runFn        func(args mock.Arguments)
-		waitDuration time.Duration
+		runFn          func(args mock.Arguments)
+		waitDuration   time.Duration
+		waitDurationFn func() time.Duration
```
mfateev: A possible code simplification is to store only the Fn and make After set the Fn that returns a passed value.

Groxx: Yeah, I'll do that.  Other stuff has func-wrappers in here, and it'd be good to get rid of some branches.

Groxx:
can't figure why coverage would be decreasing, unless it's comparing between branches (instead of against master)...

Issue #530, opened 2018-07-27
Non-Deterministic error when StartTimer and CancelTimer in one decision


yiminc:
Another repro that might need a different fix:

func init() {
	workflow.Register(Workflow)
}

// Workflow workflow decider
func Workflow(ctx workflow.Context, name string) error {
	logger := workflow.GetLogger(ctx)

	cancelTimerCtx, cancelTimerF := workflow.WithCancel(ctx)

	lao := workflow.LocalActivityOptions{
		// use short timeout as local activity is execute as function locally.
		ScheduleToCloseTimeout: time.Second,
	}
	ctx = workflow.WithLocalActivityOptions(ctx, lao)

	logger.Info("Setting up timer.")
	timerF := workflow.NewTimer(cancelTimerCtx, 10*time.Minute)

	var helloResult string
	err := workflow.ExecuteLocalActivity(ctx, helloworldActivity, name).Get(ctx, &helloResult)
	if err != nil {
		logger.Error("Hello activity failed.", zap.Error(err))
		return err
	}

	cancelTimerF()
	err = timerF.Get(ctx, nil)
	if err != nil {
		logger.Sugar().Warnf("Timer Canceled. Err: %v", err)
	}

	logger.Info("Sleeping for 20 seconds.  Kill worker.")
	workflow.Sleep(ctx, 20*time.Second)

	logger.Info("Done.")
	return nil
}

[...]

PR #532, opened 2018-07-28
Fix decision state machine

yiminc:
I think this is a better fix for #530 
Once we have this landed, I will revert #531 
Here are the changes:
1) Do not return decisions from ProcessEvent(). Previously, we get decisions after apply every event. Now, we only get decisions after apply each batch of events (one batch for one DecisionTasStarted) during replay, and also get last new decisions before we try to complete the decision task.

2) Preload LocalActivity marker event but apply them after applied the decision task started event. This is needed because we need to unblock the completed local activities. This reordering events is similar to what we already done for SideEffect/Version markers. This will not cause non-deterministic error because it only move LocalActivity marker ahead of other DecisionEvents and we know that all decision events only drive decision state machine to Initiated state and does not unblock any future/channel so it won't affect new decisions.

3) Remove decision state machine from helper once it state is completed. This is needed because previously we would get decisions after every event which will remove completed state machines but now we only get decisions for each batch of events. So if user code try to reuse ActivityID immediately after one activity is completed, after apply the first activity's completed event, we need to remove the state machine for that activity. Otherwise, the next activity as it reuse the duplicate ActivityID will have the same ID for the state machine and will have issue.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -54,7 +54,7 @@ type (
 	workflowExecutionEventHandler interface {
 		// Process a single event and return the assosciated decisions.
 		// Return List of decisions made, any error.
-		ProcessEvent(event *s.HistoryEvent, isReplay bool, isLast bool) ([]*s.Decision, error)
+		ProcessEvent(event *s.HistoryEvent, isReplay bool, isLast bool) error
 		// ProcessQuery process a query request.
 		ProcessQuery(queryType string, queryArgs []byte) ([]byte, error)
 		StackTrace() string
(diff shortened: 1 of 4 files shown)


internal/internal_task_handlers.go
```diff
@@ -672,23 +674,26 @@ ProcessEvents:
...
+					return nil, err
 				}
 			}
-
-			if w.isWorkflowCompleted {
-				// If workflow is already completed then we can break from processing
-				// further decisions.
-				break ProcessEvents
+		}
+		isReplay := len(reorderedEvents) > 0 && reorderedHistory.IsReplayEvent(reorderedEvents[len(reorderedEvents)-1])
```
mfateev: In Java I create a struct that was return of NextDecisionEvents. This way the isReplay flag and replayDecisions can be part of that struct simplifying this method.

PR #535, opened 2018-08-03
Fix query for unit test framework

yiminc:
this is a regression from data converter change.


internal/internal_workflow_testsuite_test.go
```diff
@@ -1260,8 +1260,8 @@ func (s *WorkflowTestSuiteUnitTest) Test_QueryWorkflow() {
 	stateWaitSignal, stateWaitActivity, stateDone := "wait for signal", "wait for activity", "done"
 	workflowFn := func(ctx Context) error {
 		var state string
-		err := SetQueryHandler(ctx, queryType, func() (string, error) {
-			return state, nil
+		err := SetQueryHandler(ctx, queryType, func(queryInput string) (string, error) {
```
mfateev: May be test more than one arg just to be sure?

yiminc: i have tested 2 args as well

PR #544, opened 2018-08-15
Fail decision task when process task return error

yiminc:
If it failed to process decision task, we would want to fail the decision task instead of silently timeout the task.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -628,6 +628,20 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 }
 
 func (w *workflowExecutionContextImpl) ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error) {
+	defer func() {
+		if err != nil {
+			// unexpected error happened, decision task will timeout if we don't handle this err here.
+			// convert error to DecisionTaskFailed.
+			w.wth.logger.Info("Failed to process decision task.",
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -628,6 +628,20 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 }
 
 func (w *workflowExecutionContextImpl) ProcessWorkflowTask(task *s.PollForDecisionTaskResponse, historyIterator HistoryIterator) (completeRequest interface{}, err error) {
+	defer func() {
+		if err != nil {
+			// unexpected error happened, decision task will timeout if we don't handle this err here.
+			// convert error to DecisionTaskFailed.
+			w.wth.logger.Info("Failed to process decision task.",
```
vancexu: Maybe use logger.Error

internal/internal_task_handlers.go
```diff
@@ -1187,6 +1180,17 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
 	}
 }
 
+func errorToFailDecisionTask(taskToken []byte, err error, identity string) *s.RespondDecisionTaskFailedRequest {
+	failedCause := s.DecisionTaskFailedCauseWorkflowWorkerUnhandledFailure
+	_, details := getErrorDetails(err, nil)
```
vancexu: Should we use wth.dataConverter instead of nil here?

vancexu: Ok, then it make sense. 

Would be good to also add a test for this change. 
Other than that LGTM

PR #545, opened 2018-08-15
Handle cancel timer failed event

yiminc:
CancelTimer/CancelActivity could fail. We need to handle that case.


internal/internal_task_handlers.go
```diff
@@ -186,9 +186,9 @@ func isDecisionEvent(eventType s.EventType) bool {
 		s.EventTypeWorkflowExecutionCanceled,
 		s.EventTypeWorkflowExecutionContinuedAsNew,
 		s.EventTypeActivityTaskScheduled,
-		s.EventTypeActivityTaskCancelRequested,
+		s.EventTypeActivityTaskCancelRequested, s.EventTypeRequestCancelActivityTaskFailed,
```
mfateev: Why is it on the same line?

internal/internal_task_handlers.go
```diff
@@ -186,9 +186,9 @@ func isDecisionEvent(eventType s.EventType) bool {
...
 		s.EventTypeTimerStarted,
-		s.EventTypeTimerCanceled,
+		s.EventTypeTimerCanceled, s.EventTypeCancelTimerFailed,
```
mfateev: ditto

internal/internal_task_handlers.go
yiminc: because they are for the same decision.

yiminc: Just realize that for cancel activity, the respond decision event will always be EventTypeActivityTaskCancelRequested.

PR #551, opened 2018-08-24
Add metrics for non-deterministic error.

yiminc:


internal/internal_task_handlers.go
```diff
@@ -707,6 +707,7 @@ ProcessEvents:
 	if !skipReplayCheck {
 		// check if decisions from reply matches to the history events
 		if err := matchReplayWithHistory(replayDecisions, respondEvents); err != nil {
+			w.wth.metricsScope.Counter(metrics.NonDeterministicError).Inc(1)
```
vancexu: Do we need some kind of tag here like workflowType?

PR #557, opened 2018-08-31
Add support for RetryPolicy in test framework

yiminc:
Add support for RetryPolicy in test framework.

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -208,12 +208,7 @@ func newTestWorkflowEnvironmentImpl(s *WorkflowTestSuite) *testWorkflowEnvironme
 	}
 
 	// move forward the mock clock to start time.
-	startTime := env.startTime
-	if startTime == time.Unix(0, 0) {
-		// if start time not set, use current clock time
-		startTime = env.wallClock.Now()
-	}
-	env.mockClock.Add(startTime.Sub(env.mockClock.Now()))
+	env.setStartTime(time.Now())
 
 	// put current workflow as a running workflow so child can send signal to parent
 	env.runningWorkflows[env.workflowInfo.WorkflowExecution.ID] = &testWorkflowHandle{env: env, callback: func(result []byte, err error) {}}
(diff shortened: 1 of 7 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -684,6 +697,35 @@ func (env *testWorkflowEnvironmentImpl) Complete(result []byte, err error) {
...
+		var expireTime time.Time
+		if params.retryPolicy.GetExpirationIntervalInSeconds() > 0 {
+			expireTime = params.scheduledTime.Add(time.Second * time.Duration(params.retryPolicy.GetExpirationIntervalInSeconds()))
+		}
+		backoff := getRetryBackoffFromThriftRetryPolicy(params.retryPolicy, env.workflowInfo.Attempt, errReason, env.Now(), expireTime)
+		if backoff > 0 {
+			env.registerDelayedCallback(func() {
+				// remove the current child workflow from the pending child workflow map because
+				// the childWorkflowID will be the same for retry run.
+				delete(env.runningWorkflows, env.workflowInfo.WorkflowExecution.ID)
```
mfateev: What's going to happen if signal is sent to a child while waiting for a retry? I think in real life it should be delivered to the next run.

internal/internal_workflow_testsuite_test.go
```diff
@@ -1792,3 +1792,137 @@ func (s *WorkflowTestSuiteUnitTest) Test_ContextMisuse() {
...
+}
+
+func (s *WorkflowTestSuiteUnitTest) Test_LocalActivityRetry() {
+
+	localActivityFn := func(ctx context.Context) (string, error) {
+		info := GetActivityInfo(ctx)
+		if info.Attempt < 2 {
+			return "", NewCustomError("bad-luck")
+		}
+		return "retry-done", nil
```
mfateev: May be return the non retryable error? Or have a test just for it?

internal/internal_workflow_testsuite.go
yiminc: Just realize that currently, there is no way to send signal to child workflow in the unit test environment. https://github.com/uber-go/cadence-client/issues/564 will address that separately.

internal/internal_workflow_testsuite_test.go
yiminc: will update to cover non-retryable error case.

internal/internal_workflow_testsuite_test.go
```diff
@@ -1808,18 +1816,23 @@ func (s *WorkflowTestSuiteUnitTest) Test_ActivityRetry() {
...
 				BackoffCoefficient:       2,
-				NonRetriableErrorReasons: []string{"bad-error"},
+				NonRetriableErrorReasons: []string{"bad-bug"},
 				ExpirationInterval:       time.Minute,
 			},
 		}
 		ctx = WithActivityOptions(ctx, ao)
 
+		err := ExecuteActivity(ctx, activityFailedFn).Get(ctx, nil)
+		if err == nil {
```
mfateev: May be add the check that it is indeed "bad-bug"

PR #566, opened 2018-09-06
Enable signal to child workflow in test framework

yiminc:
Add TestWorkflowEnvironment.SignalChildWorlfow(). Also modify the retry for child workflow to make sure the signal send during backoff time is delivered to the next attempt.

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -511,9 +511,9 @@ func (t *TestWorkflowEnvironment) SignalWorkflow(name string, input interface{})
 	t.impl.signalWorkflow(name, input)
 }
 
-// SignalWorkflow sends signal to the currently running test workflow.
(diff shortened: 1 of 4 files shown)


internal/workflow_testsuite.go
```diff
@@ -511,6 +511,11 @@ func (t *TestWorkflowEnvironment) SignalWorkflow(name string, input interface{})
 	t.impl.signalWorkflow(name, input)
 }
 
+// SignalWorkflow sends signal to the currently running test workflow.
+func (t *TestWorkflowEnvironment) SignalChildWorkflow(workflowID, signalName string, input interface{}) error {
```
mfateev: Will it signal the parent? Or it doesn't have an ID?
If it does shouldn't it be called SignalByID?

yiminc: Yes, it could signal parent if you pass in parent workflow ID. So rename to SignalWorkflowByID sounds good.

PR #570, opened 2018-09-10
Retry with heartbeat details

yiminc:
Heartbeat support in test framework will be added shortly in a separate diff.


yiminc:
https://github.com/uber-go/cadence-client/issues/571

PR #572, opened 2018-09-10
Add heartbeat support to test framework

yiminc:
Add heartbeat support to test framework.

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -63,8 +63,9 @@ type (
 	}
 
 	testActivityHandle struct {
-		callback     resultHandler
-		activityType string
+		callback         resultHandler
+		activityType     string
+		heartbeatDetails []byte
 	}
 
 	testWorkflowHandle struct {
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -840,11 +842,15 @@ func (env *testWorkflowEnvironmentImpl) executeActivityWithRetryForTest(
 			backoff := getRetryBackoffWithNowTime(p, task.GetAttempt(), *request.Reason, env.Now(), expireTime)
 			if backoff > 0 {
 				// need a retry
-				task.Attempt = common.Int32Ptr(task.GetAttempt() + 1)
 				waitCh := make(chan struct{})
 				env.postCallback(func() { env.runningCount-- }, false)
 				env.registerDelayedCallback(func() {
 					env.runningCount++
+					task.Attempt = common.Int32Ptr(task.GetAttempt() + 1)
+					activityID := string(task.TaskToken)
```
mfateev: Does such task token work if the same activity is invoked from multiple child workflows?

yiminc: Good catch! It won't work, and it is a bug. :)

PR #573, opened 2018-09-11
Fix test framework to allow same ActivityID from different child workflow

yiminc:


internal/internal_workflow_testsuite.go
```diff
@@ -819,6 +819,25 @@ func (env *testWorkflowEnvironmentImpl) ExecuteActivity(parameters executeActivi
...
+}
+
+func (env *testWorkflowEnvironmentImpl) deleteHandle(activityID string) {
+	delete(env.activities, env.makeUniqueID(activityID))
+}
+
+func (env *testWorkflowEnvironmentImpl) makeUniqueID(id string) string {
+	// ActivityID is unique per workflow, but different workflow could have same activityID.
+	// Make the key unique globally as we share the same collection for all running workflows in test.
+	return fmt.Sprintf("%v_%v", env.WorkflowInfo().WorkflowExecution.ID, id)
```
mfateev: Shouldn't it be the run id?

yiminc: Changed to RunID.

PR #575, opened 2018-09-13
Fix max concurrent decision task

yiminc:
The assumption was that when the ProcessTask() method returns the work is done, so the scheduler will trigger another poll to get more work. This was true until the local activity come into place. The ProcessTask() could return with nothing when there is local activity running. 
This fix makes ProcessTask() block until all work is done so the max concurrent execution control could work and it won't poll down more work than it could handle.

From implementation, major change is that, when local activity is done by local activity worker, we need a way to send back the result to the ProcessTask() to unblock it. The way that is implemented is that the ProcessWorkflowTask() now take workflowTask struct which has a channel, so when we create local activity task before send it to local activity worker, we set that result channel to the task. So when local activity worker finishes process, it can send the result to that result channel which the ProcessTask() is waiting on. It also need a done channel, so that if the ProcessTask() return earlier, the local activity worker won't block for ever trying to send over the result.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -69,6 +69,8 @@ type (
 	workflowTask struct {
 		task            *s.PollForDecisionTaskResponse
 		historyIterator HistoryIterator
+		doneCh          chan struct{}
+		laResultCh      chan *localActivityResult
 	}
 
 	// activityTask wraps a activity task.
@@ -424,7 +426,7 @@ func (w *workflowExecutionContextImpl) onEviction() {
 	w.mutex.Unlock()
 }
 
-func (w *workflowExecutionContextImpl) isDestroyed() bool {
+func (w *workflowExecutionContextImpl) IsDestroyed() bool {
 	return w.eventHandler == nil
 }
 
@@ -578,7 +580,7 @@ func (w *workflowExecutionContextImpl) resetStateIfDestroyed(task *s.PollForDeci
 	// It is possible that 2 threads (one for decision task and one for query task) that both are getting this same
 	// cached workflowContext. If one task finished with err, it would destroy the cached state. In that case, the
 	// second task needs to reset the cache state and start from beginning of the history.
-	if w.isDestroyed() {
+	if w.IsDestroyed() {
 		w.createEventHandler()
 		// reset history events if necessary
 		if !isFullHistory(task.History) {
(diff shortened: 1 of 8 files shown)


internal/internal_task_handlers.go
```diff
@@ -424,7 +426,7 @@ func (w *workflowExecutionContextImpl) onEviction() {
 	w.mutex.Unlock()
 }
 
-func (w *workflowExecutionContextImpl) isDestroyed() bool {
+func (w *workflowExecutionContextImpl) IsDestroyed() bool {
```
mfateev: Is it safe to check without locking?

internal/internal_task_pollers.go
```diff
@@ -227,26 +221,77 @@ func (wtp *workflowTaskPoller) processWorkflowTask(workflowTask *workflowTask) e
...
+	defer close(doneCh)
+
+process_WorkflowTask_Loop:
+	for {
+		startTime := time.Now()
+		workflowTask.doneCh = doneCh
+		workflowTask.laResultCh = laResultCh
+		completedRequest, wc, err := wtp.taskHandler.ProcessWorkflowTask(workflowTask)
+		if err == nil && completedRequest == nil {
+			// decision task cannot complete, we need a timer to force complete it
```
mfateev: Confusing. May be explain that it is waiting on local activity?

internal/internal_task_pollers.go
```diff
@@ -227,26 +221,77 @@ func (wtp *workflowTaskPoller) processWorkflowTask(workflowTask *workflowTask) e
...
+		wait_LocalActivity_Loop:
+			for {
+				deadlineToTrigger := time.Duration(float32(0.8) * float32(wc.GetDecisionTimeout()))
```
mfateev: Where does this 0.8 coming from? May be make it a constant with self describing name?

internal/internal_task_pollers.go
```diff
@@ -467,7 +464,12 @@ func (latp *localActivityTaskPoller) PollTask() (interface{}, error) {
 
 func (latp *localActivityTaskPoller) ProcessTask(task interface{}) error {
 	result := latp.handler.executeLocalActivityTask(task.(*localActivityTask))
-	latp.laTunnel.deliverResult(result)
+	select {
```
mfateev: Add comment explaining as this code is confusing. The way I understood is that laResultCh is synchronous and there is guarantee that doneCh is closed if nobody is listening on laResultCh.

internal/internal_task_handlers.go
yiminc: No. But they are all called with lock obtained.

internal/internal_task_pollers.go
yiminc: will add more comments

internal/internal_task_pollers.go
yiminc: added.

internal/internal_task_pollers.go
yiminc: your understanding is correct. Will add more comments.

PR #577, opened 2018-09-14
Flush buffered heartbeat if activity failed

yiminc:
We buffer heartbeat and only send the latest to server periodically. The interval is 0.8 of configured heartbeat timeout. This works fine, except that we lost the unreported progress when activity failed.
This fix try to flush buffered heartbeat in case activity failed, before we respond activity task failed.


yiminc:
Issue filed: https://github.com/uber/cadence/issues/1116

PR #578, opened 2018-09-14
Return WorkflowAlreadyStarted error from child workflow start failed

yiminc:
Return WorkflowAlreadyStarted if start child workflow failed.

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -1005,7 +1005,9 @@ func (weh *workflowExecutionEventHandlerImpl) handleStartChildWorkflowExecutionF
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


internal/internal_event_handlers.go
```diff
@@ -1005,7 +1005,9 @@ func (weh *workflowExecutionEventHandlerImpl) handleStartChildWorkflowExecutionF
 		return nil
 	}
 
-	err := fmt.Errorf("start child workflow failed: %v", attributes.GetCause())
+	err := &m.WorkflowExecutionAlreadyStartedError{
```
vancexu: Why we don't care about detail failure cause now?

mfateev: AlreadyStarted is the only cause.

PR #587, opened 2018-10-10
Add executable checksum to DecisionTaskCompleted

yiminc:
Add executable binary md5 checksum to complete decision task as ExecutionContext.


internal/internal_worker.go
```diff
@@ -900,6 +908,47 @@ func (aw *aggregatedWorker) Start() error {
...
+		return err
+	}
+
+	checksum := h.Sum(nil)
+	executableChecksum = hex.EncodeToString(checksum[:])
+
+	return nil
+}
+
+func getExecutableChecksum() string {
```
vancexu: Why we don't need lock when read?

internal/internal_worker.go
```diff
@@ -900,6 +908,47 @@ func (aw *aggregatedWorker) Start() error {
 	return nil
 }
 
+var executableChecksum string
+var executableChecksumLock sync.Mutex
+
+func setExecutableChecksum() error {
```
vancexu: might be better rename `initExecutableChecksum()`

vancexu:
may also need to update `FeatureVersion` for server compatibility

PR #588, opened 2018-10-11
Include LocalActivityType into the result marker

yiminc:
Add ActivityType to the local activity result marker, and check ActivityType during replay.

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -125,13 +125,14 @@ type (
 	}
 
 	localActivityMarkerData struct {
-		ActivityID string        `json:"activityId,omitempty"`
-		ErrReason  string        `json:"errReason,omitempty"`
-		ErrJSON    string        `json:"errJson,omitempty"` // string instead of []byte so the encoded blob is human readable
-		ResultJSON string        `json:"resultJson,omitempty"`
-		ReplayTime time.Time     `json:"replayTime,omitempty"`
-		Attempt    int32         `json:"attempt,omitempty"` // record attempt, starting from 0.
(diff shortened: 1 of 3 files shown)


internal/workflow.go
```diff
@@ -415,6 +415,7 @@ func ExecuteLocalActivity(ctx Context, activity interface{}, args ...interface{}
 	params := &executeLocalActivityParams{
 		localActivityOptions: *options,
 		ActivityFn:           activity,
+		ActivityType:         lastPartOfName(getFunctionName(activity)),
```
samarabbas: Can we register local activity like we do for regular one?
I'm little worried that using last part of the name is might not be enough as it will still break replay if someone renames a function.

yiminc: We do the same for regular activity. The intention is that in case user move code around (from one package to another package), we don't want to break their workflow.

yiminc: Register would not help, unless your enforce that every local activity should be registered with a friendly name. Even that would not prevent problem because user could still update the registered name.

internal/internal_event_handlers.go
```diff
@@ -933,6 +934,11 @@ func (weh *workflowExecutionEventHandlerImpl) handleLocalActivityMarker(markerDa
 	}
 
 	if la, ok := weh.pendingLaTasks[lamd.ActivityID]; ok {
+		if len(lamd.ActivityType) > 0 && lamd.ActivityType != la.params.ActivityType {
```
mfateev: We could have a separate counter that includes al events including local activity scheduling events and pass that counter down to localActivityResult to ensure determinism.

PR #590, opened 2018-10-13
Plumb through workflow type into client (#1170)

andrewjdawson2016:
This is the client side change needed to plumb the workflow type into the activity info. 

The following image shows the code the activity author writes to access these fields 
screen shot 2018-10-15 at 10 18 19 pm

As a follow up diff I will update metrics and logger to include these tags.


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -942,6 +942,7 @@ struct PollForActivityTaskResponse {
   120: optional i32 attempt
   130: optional i64 (js.type = "Long") scheduledTimestampOfThisAttempt
   140: optional binary heartbeatDetails
+  150: optional WorkflowType workflowType
```
mfateev: Let's add workflowDomain field as well.

internal/activity.go
```diff
@@ -49,6 +49,8 @@ type (
 		StartedTimestamp   time.Time     // Time of activity start
 		Deadline           time.Time     // Time of activity timeout
 		Attempt            int32         // Attempt starts from 0, and increased by 1 for every retry if retry policy is specified.
+		WorkflowType       *WorkflowType
```
mfateev: I would move them just after TaskToken in this struct.

PR #591, opened 2018-10-17
Do not run local activity in query task

yiminc:
If query request comes in while the decision task is started but not completed yet, in that case the query task will contains last DecisionTaskStarted event and it is not in replay. However, we don't want to run local activity in query task.


yiminc:
For query task, ideally yes it should stop on last decision task started. But it won't prevent pending local activity from running. The problem is one decision task could generate say 10 local activities, and 5 of them finished before timeout, so worker will respond back with that 5 marker, and immediately get a new decision task. Now if a query task is dispatched at this time, even if we stop after previous decision task started, the other 5 local activities are still pending and could be scheduled if we don't have this fix.

PR #597, opened 2018-10-24
Handle illegal StateMachine transition as Non-Deterministic error

yiminc:
Handle illegal state machine panic as non-deterministic error so it will be handled according to NonDeterministicWorkflowPolicy.
This change is needed to hook up the Reset API once it is ready on server side.

### DIFF
--- a/internal/internal_decision_state_machine.go
+++ b/internal/internal_decision_state_machine.go
@@ -111,6 +111,11 @@ type (
 		scheduledEventIDToCancellationID map[int64]string
 		scheduledEventIDToSignalID       map[int64]string
 	}
+
+	// panic when decision state machine is in illegal state
+	stateMachineIllegalStatePanic struct {
+		message string
+	}
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/internal_decision_state_machine.go
```diff
@@ -306,9 +311,17 @@ func (d *decisionStateMachineBase) moveState(newState decisionState, event strin
...
+}
+
+func panicIllegalState(message string) {
+	panic(stateMachineIllegalStatePanic{message: message})
+}
+
 func (d *decisionStateMachineBase) failStateTransition(event string) {
 	// this is when we detect illegal state transition, likely due to ill history sequence or nondeterministic decider code
-	panic(fmt.Sprintf("invalid state transition: attempt to %v, %v", event, d))
+	panicIllegalState(fmt.Sprintf("invalid state transition: attempt to %v, %v", event, d))
```
vancexu: I wonder what is the benefits to wrap panic to stateMachineIllegalStatePanic?

PR #599, opened 2018-10-26
update workflow.GoNamed() doc

mikolajb:


workflow/deterministic_wrappers.go
```diff
@@ -84,7 +84,7 @@ func Go(ctx Context, f func(ctx Context)) {
 
 // GoNamed creates a new coroutine with a given human readable name.
 // It has similar semantic to goroutine in a context of the workflow.
-// Name appears in stack traces that are blocked on this Channel.
+// Name appears in stack traces that are blocked on this coroutine.
```
mfateev: good catch. I would rephrase it to something like: 
Name appears in stack traces that include this coroutine.

mikolajb: Done

PR #609, opened 2018-11-21
Return error instead of panic on duplicate childWorkflowID/ActivityID

yiminc:


internal/internal_activity.go
```diff
@@ -93,7 +93,7 @@ type (
 		// The ExecuteActivity schedules an activity with a callback handler.
 		// If the activity failed to complete the callback error would indicate the failure
 		// and it can be one of ActivityTaskFailedError, ActivityTaskTimeoutError, ActivityTaskCanceledError
-		ExecuteActivity(parameters executeActivityParams, callback resultHandler) *activityInfo
+		ExecuteActivity(parameters executeActivityParams, callback resultHandler) (*activityInfo, error)
```
mfateev: Could we just call the callback with the error synchronously?

mfateev:
Add the unit tests for the duplicated ids.

yiminc:
Discussed offline, will keep existing behavior to panic so the workflow will be stuck. Because most of the time, when user are using duplicate IDs (for activity or child workflow), it indicate logic bug in their workflow code. So panic and stuck the workflow is a desired behavior so user could fix the code and it will resume from there. If we return error, workflow would make unintended progress.

PR #612, opened 2018-11-27
Use correct long poll timeout

yiminc:


yiminc:
Looks like this is not an issue. Client side uses 25s timeout while server side would return empty result at 20s, and client side would retry on empty result.

PR #613, opened 2018-11-28
Changed to not require retryPolicy expirationInterval when maxRetries…

mfateev:
… is present

### DIFF
--- a/internal/error_test.go
+++ b/internal/error_test.go
@@ -266,6 +266,7 @@ func Test_CustomError_Pointer(t *testing.T) {
 		return err2 // pointer in details
 	}
 	RegisterWorkflow(errorWorkflowFn2)
+	wfEnv = s.NewTestWorkflowEnvironment()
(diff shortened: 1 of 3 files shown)


internal/internal_activity.go
```diff
@@ -227,8 +227,8 @@ func validateRetryPolicy(p *shared.RetryPolicy) error {
 	if p.GetMaximumAttempts() < 0 {
 		return errors.New("negative MaximumAttempts on retry policy is invalid")
 	}
-	if p.GetExpirationIntervalInSeconds() <= 0 {
-		return errors.New("ExpirationIntervalInSeconds on retry policy must be greater than 0")
+	if p.GetExpirationIntervalInSeconds() < 0 {
```
yiminc: should add check to make sure max attempt and expiration are not 0 at same time.

mfateev: It is already there on line 236

Issue #616, opened 2018-12-04
Make TestWorkflowEnvironment support test multiple workflows


yiminc-zz:
Currently, the TestWorkflowEnvironment can run multiple workflows but only for parent-child relationships. It would be good to be able to use TestWorkflowEnvironment to test 2 workflows that are not parent-child.

PR #617, opened 2018-12-04
Panic test if attempt to run multiple workflows with same test env

yiminc:
Current TestWorkflowEnvironment only support to run one workflow.
Created task to support testing multiple workflows with one env instancehttps://github.com/uber-go/cadence-client/issues/616
This PR will detect the case and throw explicit message so we don't run into wired error (like data race issues).

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -50,6 +50,7 @@ const (
 	defaultTestRunID            = "default-test-run-id"
 	defaultTestWorkflowTypeName = "default-test-workflow-type-name"
 	defaultTestDomainName       = "default-test-domain-name"
+	workflowTypeNotSpecified    = "workflow-type-not-specified"
 )
 
 type (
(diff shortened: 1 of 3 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -371,7 +372,15 @@ func (env *testWorkflowEnvironmentImpl) executeWorkflow(workflowFn interface{},
 }
 
 func (env *testWorkflowEnvironmentImpl) executeWorkflowInternal(delayStart time.Duration, workflowType string, input []byte) {
+	env.locker.Lock()
+	if env.workflowInfo.WorkflowType.Name != workflowTypeNotSpecified {
+		// Current TestWorkflowEnvironment only support to run one workflow.
+		// Created task to support testing multiple workflows with one env instancehttps://github.com/uber-go/cadence-client/issues/616
```
vancexu: Maybe worth to put a TODO mark here.

PR #618, opened 2018-12-07
Update shared idl to include archival config

andrewjdawson2016:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -781,9 +781,17 @@ struct DomainInfo {
   50: optional map<string,string> data
 }
 
+const i32 ARCHIVAL_RETENTION_DAYS_INFINITE = -1
+
+struct ArchivalConfiguration {
+  10: optional string bucketName
+  20: optional i32 retentionDays
```
mfateev: I believe we use 0 or not present for infinite in other structs. For example expiration on a retry policy.

PR #623, opened 2018-12-19
Add cron support

yiminc:


internal/client.go
```diff
@@ -280,6 +280,19 @@ type (
...
+		// as a cron based on the schedule. Cron workflow never stop until it is terminated or cancelled (by returning
+		// cadence.CanceledError). The cron spec is as following:
+		// ┌───────────── minute (0 - 59)
+		// │ ┌───────────── hour (0 - 23)
+		// │ │ ┌───────────── day of the month (1 - 31)
+		// │ │ │ ┌───────────── month (1 - 12)
+		// │ │ │ │ ┌───────────── day of the week (0 - 6) (Sunday to Saturday)
+		// │ │ │ │ │
+		// │ │ │ │ │
+		// * * * * *
```
mfateev: Add information about which timezone is used.

mfateev: Also add information about what happens when workflow that has CronSchedule fails. Also what is behavior if workflow is still running while the next schedule is due. How does the workflow execution timeout relat to the schedule.

yiminc: It will use cadence server's timezone. In Uber's environment, it seems all servers in both SJC and DCA are all using UTC as timezone. I don't understand what you mean by "workflow has CronSchedule fails"? If a cron schedule is invalid, it will fail when the start is requested. If workflow is still running while next schedule is due, it will miss that schedule. When it completed, it will schedule next run using the complete time.

yiminc: Will update the doc to make those clear.

PR #641, opened 2019-02-06
Test framework: add support to mock GetVersion

yiminc:
adding support to mock GetVersion in unit test.

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -1622,28 +1622,14 @@ func (env *testWorkflowEnvironmentImpl) SideEffect(f func() ([]byte, error), cal
 	callback(f())
 }
 
-func (env *testWorkflowEnvironmentImpl) GetVersion(changeID string, minSupported, maxSupported Version) (retVerion Version) {
-	if _, ok := env.expectedMockCalls[mockMethodForGetVersion]; ok {
-		// GetVersion is mocked
-		args := []interface{}{changeID, minSupported, maxSupported}
-		// below call will panic if mock is not properly setup.
-		mockRet := env.mock.MethodCalled(mockMethodForGetVersion, args...)
-		m := &mockWrapper{name: mockMethodForGetVersion, fn: mockFnGetVersion}
-		if mockFn := m.getMockFn(mockRet); mockFn != nil {
-			executor := &activityExecutor{name: mockMethodForGetVersion, fn: mockFn}
-			reflectValues := executor.executeWithActualArgsWithoutParseResult(nil, args)
-			if len(reflectValues) != 1 || !reflect.TypeOf(reflectValues[0].Interface()).AssignableTo(reflect.TypeOf(retVerion)) {
-				panic(fmt.Sprintf("mock of GetVersion has incorrect return type, expected workflow.Version, but actual is %T (%v)",
-					reflectValues[0].Interface(), reflectValues[0].Interface()))
-			}
-			return reflectValues[0].Interface().(Version)
-		} else {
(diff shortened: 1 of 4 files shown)


internal/workflow_testsuite.go
```diff
@@ -306,6 +307,17 @@ func (t *TestWorkflowEnvironment) OnRequestCancelExternalWorkflow(domainName, wo
 	return t.wrapCall(call)
 }
 
+// OnGetVersion setup a mock for workflow.GetVersion() call. By default, if mock is not setup, the GetVersion call from
+// workflow code will always return the maxSupported version. Make it not possible to test old version branch. With this
+// mock support, it is possible to test code branch for different versions.
+// Note: if mock for GetVersion is setup, the test framework expect all calls to GetVersion from the tested workflow are
+// all mocked. It is not supported to have one GetVersion call mocked and the second GetVersion call not mocked. You need
+// to either not mock any GetVersion call, or mock all GetVersion calls.
```
mfateev: I don't think it is acceptable. The problem is that the workflow might reference some library that calls GetVersion but the implementer of the workflow might not even know about it. I believe it should be possible to mock each "changeId" individually.

internal/internal_workflow_testsuite.go
```diff
@@ -1622,7 +1622,31 @@ func (env *testWorkflowEnvironmentImpl) SideEffect(f func() ([]byte, error), cal
...
+		if mockFn := m.getMockFn(mockRet); mockFn != nil {
+			executor := &activityExecutor{name: mockMethodForGetVersion, fn: mockFn}
+			reflectValues := executor.executeWithActualArgsWithoutParseResult(nil, args)
+			if len(reflectValues) != 1 || !reflect.TypeOf(reflectValues[0].Interface()).AssignableTo(reflect.TypeOf(retVerion)) {
+				panic(fmt.Sprintf("mock of GetVersion has incorrect return type, expected workflow.Version, but actual is %T (%v)",
+					reflectValues[0].Interface(), reflectValues[0].Interface()))
+			}
+			return reflectValues[0].Interface().(Version)
+		} else {
+			if len(mockRet) != 1 || !reflect.TypeOf(mockRet[0]).AssignableTo(reflect.TypeOf(retVerion)) {
```
dmetzgar: Could this be `reflect.TypeOf((*Version)(nil))` instead?

Also, fix spelling of `retVerion`.

internal/workflow_testsuite.go
yiminc: good point, let me update

internal/internal_workflow_testsuite.go
yiminc: remove retVersion and use DefaultVersion instead.

internal/workflow_testsuite.go
```diff
@@ -310,11 +310,11 @@ func (t *TestWorkflowEnvironment) OnRequestCancelExternalWorkflow(domainName, wo
 // OnGetVersion setup a mock for workflow.GetVersion() call. By default, if mock is not setup, the GetVersion call from
 // workflow code will always return the maxSupported version. Make it not possible to test old version branch. With this
 // mock support, it is possible to test code branch for different versions.
-// Note: if mock for GetVersion is setup, the test framework expect all calls to GetVersion from the tested workflow are
-// all mocked. It is not supported to have one GetVersion call mocked and the second GetVersion call not mocked. You need
-// to either not mock any GetVersion call, or mock all GetVersion calls.
+//
+// Note: mock can be setup for a specific changeID. Or if mock.Anything is used as changeID then all calls to GetVersion
+// will be mocked.
```
mfateev: You can add that mock for a specific changeID has priority over Anything mock.

PR #645, opened 2019-02-15
Graceful handling of goexit inside an activity while testing (#644)

### LINKED ISSUE #644: Activity doesn't fail if something inside the activity calls Goexit
For example, if test instrumentation replaces a dependency somewhere and something calls `t.FailNow()` inside the activity. `FailNow` calls `runtime.Goexit`, which means that the test doesn't propagate the result correctly.

This can be verified by just calling `runtime.Goexit()` inside an activity and run it through a test suite.

recht:
Fixes bad handling of Goexit by moving `postCallback` to a defer that checks if result is empty. In case of `Goexit`, `defer` is still called, but `recover()` returns `nil`.
Previous to this, Goexit would bubble up, and `postCallback` wouldn't be invoked, which eventually would trigger a test timeout without any indication of the actual error.

Fixes #644

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -863,14 +863,23 @@ func (env *testWorkflowEnvironmentImpl) ExecuteActivity(parameters executeActivi
 	env.setActivityHandle(activityInfo.activityID, activityHandle)
 	env.runningCount++
 	// activity runs in separate goroutinue outside of workflow dispatcher
+	// do callback in a defer to handle calls to runtime.Goexit inside the activity (which is done by t.FailNow)
 	go func() {
-		result := env.executeActivityWithRetryForTest(taskHandler, parameters, task)
-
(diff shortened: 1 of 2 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -863,14 +863,23 @@ func (env *testWorkflowEnvironmentImpl) ExecuteActivity(parameters executeActivi
...
-
-		// post activity result to workflow dispatcher
-		env.postCallback(func() {
-			env.handleActivityResult(activityInfo.activityID, result, parameters.ActivityType.Name, parameters.DataConverter)
-			env.runningCount--
-		}, false /* do not auto schedule decision task, because activity might be still pending */)
+		var result interface{}
+		defer func() {
+			if result == nil && recover() == nil {
+				reason := "activity exited"
```
mfateev: Would you make it more explicit? Something like "activity called runtime.Goexit"?

PR #646, opened 2019-02-20
Update doc for list API

vancexu:
Also fix some typos

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -289,7 +289,7 @@ type (
 const (
 	// WorkflowIDReusePolicyAllowDuplicateFailedOnly allow start a workflow execution
 	// when workflow not running, and the last execution close state is in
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -201,6 +201,9 @@ type (
 
 		// ListClosedWorkflow gets closed workflow executions based on request filters.
 		// Retrieved workflow executions are sorted by start time in descending order.
+		// (Retrieved workflow executions could also be sorted by closed time in descending order,
+		// if cadence server side config EnableReadFromClosedExecutionV2 is set to true.)
```
mfateev: Shouldn't this flag be set to true by default?

vancexu: I personally also like to default true. But for backward compatibility I leave it default false.

Issue #563, opened 2018-09-06
Feature request: cadence-instance-based registration rather than global


Groxx:
Most likely this is a relatively large change, but it'd help in a few ways.

In an abstract nutshell: I'd like to be able to `cadence.NewInstance().RegisterWorkflow(...)` and run multiple instances side-by-side.

Currently, mostly, I'm interested in running separate instances in tests so they can't impact each other, and so I can register as many funcs as I wish without risk of duplication or polluting the registered-names list.  E.g. currently I can't do this:
```bash
go test -count 10 .
```
with
```go
func TestThing(t *testing.T) {
  f := func(ctx Context) error { ... }
  workflow.RegisterWorkflow(f)
  NewTestWorkflowEnvironment().ExecuteWorkflow(f)
}
```
because the same-(inferred-)named func will be registered twice.  I'm working around this with name-based test registration that generates a random name, which works, but it's very much a *workaround* and requires being careful to generate unique names.

---

It's also kinda hard to work with e.g. [Fx](https://github.com/uber-go/fx) and bound methods at the moment.  I haven't come up with a good way to register them that also works in tests (especially [FxTest](https://github.com/uber-go/fx/blob/master/fxtest/fxtest.go) instances, which are useful for integration tests, as they re-run their `Provide`d funcs with each test):
```go
type activity struct {
    repositories
    gateways
    etc
}
func New(repositories, gateways, etc) activity {
    act := &activity{...}
    // do I activity.Register(act.run) here?
    // then I can't call New more than once, ever, per process.  including in tests.
    // same for other fx lifecycle methods, like OnStart hooks.
    // ... but where else do I do it?
}
func (a activity) run(ctx Context, argsFromWorkflow...) {
    a.gateway.DoThing(ctx, arg)
    // etc
}
```
This same difficulty applies to *any* instance-based pattern, be it Fx or some other dependency injector or just doing things by hand without globals.  A single global instance of Cadence makes any non-global system harder to build, and most systems / patterns intentionally avoid globals.

---

Currently, to get access to run-time objects like dependency-injection-constructed things (or anything non-global) the recommendation is to inject it into the worker's context and then extract it in the activity's `Context`.

While this does function as a workaround, it has a couple issues:
- it's only able to do a single instance / be associated with whatever could be constructed *before* starting Cadence.  Instances created afterward need to be accessed somehow, requiring a global or (likely per-Context) registry of some sort, which sorta goes back to "but we're trying to avoid globals"
[...]

PR #650, opened 2019-02-25
Adds ability to unregister workflows

algobardo:


internal/workflow.go
```diff
@@ -232,6 +232,19 @@ func RegisterWorkflow(workflowFunc interface{}) {
 	RegisterWorkflowWithOptions(workflowFunc, RegisterWorkflowOptions{})
 }
 
+// UnRegisterWorkflow - un-registers a workflow from the framework.
+// The supplied workflow can be a workflow function or a string.
+// In case a string is passed, that should be the alias of the workflow.
+// In case a function is passed, all the aliases referring to the workflow (if any) will be unregistered as well.
+// This method calls panic if workflowFunc doesn't comply with the expected format.
+func UnRegisterWorkflow(wf interface{}) {
```
mfateev: Would you add comment in which situation this function is needed. Otherwise it creates false impression that it should be called on shutdown. My understanding that it is needed only for test scenarios, correct?

algobardo: Hi Maxim, 
you are right, this function is only for testing. If you have a better idea on how to expose this functionality I can move it around, otherwise I will just add a comment.
Please note that it would be nice if cadencefx unregisters all workflows OnStop, see T2502463

mfateev: I see. So unregistering is not really what you want. You want ability to register workflows not globally, but to a specific Worker.

mfateev: I believe https://github.com/uber-go/cadence-client/issues/563 is what you want.

algobardo: I think #563 is more than I want. There are two problems that I want to solve, and this diff is only the first step to solve them, possibly the wrong one.

- Whenever there are two tests in the same package that attempt to do `fxtest.New(Module).Start()` only the first can succeed, since the second one will fail to start due to the attempt to register the same workflows twice. The same also applies for activities registration. Cadencefx should unregister all workflows and activities OnStop. (This is T2502463)

- When testing a workflow that calls a child workflow we need to:
1. Register the child workflow (mandatory)
2. Mock it with OnWorkflow (optional)
In the case we only want to do 1. and run the actual workflow, the service will necessarily use a mock of its downstream dependencies. The ability to unregister and re-register workflows should enable us to register the same workflow with different mocks of its dependencies in different tests.

mfateev: I'm confused here.
If workflows are registered in init() I thought fxtest.New(Module).Start() wouldn't register them again.

I don't understand what "downstream dependencies" are in this case.

mfateev: I thought that mocks are linked to a specific instance of a TestWorkflowEnvironment and are not related to registrations.

algobardo: We don't register workflows nor activities in init(), and I don't fully understand how that is supposed to work within an fx service. Our workflows and our activities are methods of types that are "provided" (read instantiated) by fx https://github.com/uber-go/fx/blob/master/app.go#L89, and as such they are not available at `init` time. Such types have their own dependencies that are other types instantiated by fx.

algobardo: Note that #563 also mentions the same pattern:
```
func New(repositories, gateways, etc) activity { .  <---- fx constructor and dependencies
    act := &activity{...}
    // do I activity.Register(act.run) here?  <------  registration not in init
```

mfateev: I see. You kind of fight against the original Cadence design that workflows and activities are functions, not methods.
The original design was that fx service instantiates a background activity context which activity uses to get all the external dependencies from. And registration is global as all it does is to map a function name to a specific function implementation. And it is overridable through a TestWorkflowEnvironment for unit testing.

mfateev: I agree that some users are OK calling activities by string name instead of function pointer and want to use method pointers to register them. But until #563 is implemented I would recommend to stick to the supported way of using background context to pass dependencies to activities.

zawadzkidiana (commented):
See below

zawadzkidiana (commented):
see below

zawadzkidiana:
Thanks for this, @algobardo, and I'm sorry it's been sitting unreviewed for so long — nearly seven years is a long time to wait on feedback.

**On the change itself**

The approach is sound: `UnRegisterWorkflow` accepts either a workflow function or its registered alias, resolves the correct registration name via `getWorkflowAlias`, then removes both the function-map entry and any reverse alias mapping. Locking is consistent with the surrounding `RegisterWorkflow*` methods, and the tests are genuinely thorough — they cover re-registration after unregistering, unregistering by alias vs. by function, and (nicely) verify that unregistering one workflow doesn't disturb another's registration.

Two things worth flagging on the design, in case this gets revived:

1. `workflowAliasMap` is keyed by function name — one entry per function. The "selective un-registration" test (registering `wf2` under `alias1` then `alias2`, then unregistering `"alias1"` by name) passes because it always unregisters by the alias **string**. But if a function is registered under two different aliases and you call `UnRegisterWorkflow(wf2)` by function reference rather than by string, `getWorkflowAlias(functionName)` can only ever return the most recently registered alias — the map has room for one entry per function name. That path would silently leave the first alias's function-map entry in place. Might be worth documenting that "unregister by function reference only clears the last-registered alias," or teaching the map to hold multiple aliases per function.
2. Minor Go naming nit: idiomatic naming here would be `UnregisterWorkflow` (lowercase second syllable, matching `Unmarshal`, `Unlock`, etc.) rather than `UnRegisterWorkflow`.

**Where the code stands today**

I checked this against current `master`, and the registration internals have been substantially rewritten since 2019 — `hostEnvImpl` (the type this PR extends) no longer exists; registration now lives in a `registry` struct in `internal/registry.go` with a similar but not identical shape (it also chains registries via a `next` pointer for test isolation, which this PR's implementation doesn't account for). So this specific diff won't apply as-is; it would need to be rewritten against `registry.RegisterWorkflow`/`RegisterWorkflowWithOptions`. That said, `UnregisterWorkflow` still doesn't exist on `master`, so the feature itself is still a real gap — if you're still interested, a fresh PR against the current `registry` type, addressing the multi-alias edge case above, would be very welcome.

Thanks again for the contribution, and sorry for the long silence.

---
_Generated by [Claude Code](https://claude.ai/code)_

zawadzkidiana (commented):
UnRegisterWorkflow is a reasonable API gap to fill, and the test coverage here (re-registration, alias vs non-alias, partial unregistration when one func has two aliases) is solid work.

It doesn't apply to current code though. This targets the old hostEnvImpl global singleton, which was replaced by a per-worker registry type years ago. The diff won't merge as-is.

Recommend closing this and opening a fresh PR against internal/registry.go, porting the test cases over. Happy to review that version.

Separately: GitHub shows this branch has merge conflicts, which is also why none of the required checks (DCO, Golang integration/unit tests, lint) have ever run, they're stuck waiting on conflict resolution. Given the underlying code no longer applies, a rebase wouldn't really fix this, it needs the rewrite above instead.

PR #651, opened 2019-02-27
Gracefully handle panic in child workflow mock

mfateev:
Currently if mock panics (for example due to wrong number of return arguments) the `ChildWorkflowFuture.GetChildWorkflowExecution().Get(ctx, ...) `call never returns. As the original panic is not even logged it is not possible to get to the root cause without debugger. 

This PR ensures that even in case of panic in a mock the future returned from `GetChildWorkflowExecution` is unblocked.


mfateev:
I don't think it is needed for executeMockWithActualArgs which is used for activity invocation only. And the problem is strictly child workflow related.

PR #653, opened 2019-03-05
Improved panic message when MutableSideEffect not found during replay

algobardo:


internal/internal_event_handlers.go
```diff
@@ -569,7 +569,7 @@ func (wc *workflowEnvironmentImpl) MutableSideEffect(id string, f func() interfa
 
 	if wc.isReplay {
 		// This should not happen
-		panic("MutableSideEffect with given ID not found during replay")
+		panic(fmt.Sprintf("MutableSideEffect with given ID not found during replay, ID: %s", id))
```
mfateev: Would you change it to something like:
"Non deterministic workflow code change detected. MutableSideEffect API call doesn't have a correspondent event in the workflow history. MutableSideEffect ID: %s"

PR #655, opened 2019-03-07
Added missing Cron and parent related fields to WorkflowInfo

mfateev:


venkat1109:
Is there a unit test to verify if workflowContext is populated correctly from poll response ?

internal/internal_task_handlers.go
```diff
@@ -502,6 +509,10 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *s.PollForDecisio
 		TaskStartToCloseTimeoutSeconds:      attributes.GetTaskStartToCloseTimeoutSeconds(),
 		Domain:                              wth.domain,
 		Attempt:                             attributes.GetAttempt(),
+		CronSchedule:                        attributes.CronSchedule,
+		ContinuedExecutionRunID:             attributes.ContinuedExecutionRunId,
+		ParentWorkflowDomain:                attributes.ParentWorkflowDomain,
+		ParentWorkflowExecution:             parentWorkflowExecution,
```
vancexu: why not just use `attributes.ParentWorkflowExecution` here?

PR #656, opened 2019-03-08
Improved panic message when MutableSideEffect not found during replay

algobardo:


algobardo:
@mfateev Addressed your comments

PR #666, opened 2019-03-15
Fix buffered channel bug

yycptt:
Check if there's any blocked sends when getting values from buffer.

### DIFF
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -25,13 +25,14 @@ package internal
 import (
 	"errors"
 	"fmt"
-	"github.com/robfig/cron"
 	"reflect"
 	"runtime"
 	"strings"
 	"sync"
 	"time"
 	"unicode"
+
+	"github.com/robfig/cron"
 
 	"github.com/uber-go/tally"
 	"go.uber.org/atomic"
@@ -584,6 +585,17 @@ func (c *channelImpl) receiveAsyncImpl(callback *receiveCallback) (v interface{}
 		r := c.buffer[0]
 		c.buffer[0] = nil
 		c.buffer = c.buffer[1:]
+
+		// Move blocked sends into buffer
+		if len(c.blockedSends) > 0 {
+			b := c.blockedSends[0]
+			c.blockedSends[0] = nil
+			c.blockedSends = c.blockedSends[1:]
+			if b.fn() {
+				c.buffer = append(c.buffer, b.value)
+			}
+		}
+
 		return r, true, true
 	}
 	if c.closed {
--- a/internal/internal_workflow.go
+++ b/internal/internal_workflow.go
@@ -587,12 +587,13 @@ func (c *channelImpl) receiveAsyncImpl(callback *receiveCallback) (v interface{}
 		c.buffer = c.buffer[1:]
 
 		// Move blocked sends into buffer
-		if len(c.blockedSends) > 0 {
+		for len(c.blockedSends) > 0 {
 			b := c.blockedSends[0]
 			c.blockedSends[0] = nil
 			c.blockedSends = c.blockedSends[1:]
 			if b.fn() {
 				c.buffer = append(c.buffer, b.value)
+				break
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


yycptt:
Close the PR for now for adding a unit test.

internal/internal_workflow.go
```diff
@@ -584,6 +585,17 @@ func (c *channelImpl) receiveAsyncImpl(callback *receiveCallback) (v interface{}
 		r := c.buffer[0]
 		c.buffer[0] = nil
 		c.buffer = c.buffer[1:]
+
+		// Move blocked sends into buffer
+		if len(c.blockedSends) > 0 {
+			b := c.blockedSends[0]
+			c.blockedSends[0] = nil
+			c.blockedSends = c.blockedSends[1:]
```
yycptt: Create a popBlockedSends function for these three lines?

internal/internal_workflow.go
```diff
@@ -584,6 +585,17 @@ func (c *channelImpl) receiveAsyncImpl(callback *receiveCallback) (v interface{}
 		r := c.buffer[0]
 		c.buffer[0] = nil
 		c.buffer = c.buffer[1:]
+
+		// Move blocked sends into buffer
+		if len(c.blockedSends) > 0 {
+			b := c.blockedSends[0]
+			c.blockedSends[0] = nil
```
andrewjdawson2016: Does this line do anything given the next line?

internal/internal_workflow_test.go
```diff
@@ -757,6 +758,32 @@ func (s *WorkflowUnitTest) Test_CorruptedSignalOnClosedChannelWorkflow_Receive_S
 	s.EqualValues(0, len(result))
 }
 
+func bufferedChanWorkflowTest(ctx Context) error {
```
andrewjdawson2016: I know this was the customer's repro but there are two things I find odd about this repro
1. Calling receive inside the for loop that is doing the sending is odd. 
2. I do not think we need the // Refill the buffer loop in order to repro this. 

Can we use a simpler repro here so its clear that the bug occurs more generally and not just in this case.

andrewjdawson2016: Can we add a parameter to this workflow that includes buffer size? That way you can easily write another test for buffer size greater than one.

internal/internal_workflow_test.go
```diff
@@ -757,6 +758,32 @@ func (s *WorkflowUnitTest) Test_CorruptedSignalOnClosedChannelWorkflow_Receive_S
...
+}
+
+func (s *WorkflowUnitTest) Test_BufferedChanWorkflow() {
+	env := s.NewTestWorkflowEnvironment()
+
+	env.ExecuteWorkflow(bufferedChanWorkflowTest)
+	s.True(env.IsWorkflowCompleted())
+	s.NoError(env.GetWorkflowError())
+}
+
```
andrewjdawson2016: Can we add a test or maybe even in the same test that passes buffer size > 1

internal/internal_workflow_test.go
yycptt: Yeah I can change the bufferSize to a parameter.

internal/internal_workflow_test.go
yycptt: Once bufferSize is a parameter, I can use a loop to test more cases.

internal/internal_workflow.go
```diff
@@ -584,6 +585,17 @@ func (c *channelImpl) receiveAsyncImpl(callback *receiveCallback) (v interface{}
...
+			c.blockedSends = c.blockedSends[1:]
+			if b.fn() {
```
mfateev: I think it should iterate through blockedSends until b.fn() returns true. In this case if it returns false nothing is added to the buffer.

mfateev: I think you have to use Select that matches a different branch than send to reproduce this edge case.

internal/internal_workflow.go
yycptt: No actually. I am just following existing pattern in the code. See Line 606.

internal/internal_workflow_test.go
yycptt: Sure. I can create a simple one.

internal/internal_workflow.go
mfateev: I believe it would retain memory without this assignment.

internal/internal_workflow.go
yycptt: Updated and created a test case for it.

Issue #637, opened 2019-02-01
Add activity fields to Logger returned by activity.GetLogger


mfateev:
Customer report:

`I thought activity.GetLogger(ctx) would return the same zap logger that cadence uses from workflow.GetLogger() ... with all the default fields such as wf id, type, runid, etc eg {"log":"{\"Domain\":\"clm-cadence-phx3\",\"RunID\":\"0ffba59e-ce35-4400-81fd-01c728a110b9\",\"TaskList\":\"clmTaskList\",\"WorkerID\":\"1@infradocker03-phx2@clmTaskList\",\"WorkflowID\":\"ProvisionWorkflow-u113715-phx3-1548982163\",\
"WorkflowType\":\"code.uber.internal/infra/clm/cadence/workflows.ProvisionWorkflow\",\"activityName\":\"DisableHostActivity\",\"issueID\":\"compute::issue:090340c46f0955fdbbbecda859cb52f3\",\"level\":\"info\",\"msg\":\"zapify.go:68 S
leeping before retrying activity\",\"retryNum\":0,\"retryNumMax\":2,\"sleep\":30000000000,\"sleepTime\":300000000000,\"target\":\"u113715-phx3\",\"time\":\"2019-02-01T01:04:24Z\",\"timeout\":600000000000}\n","stream":"stderr","time":
"2019-02-01T01:04:24.128733984Z"}
5:37 PM
But it looks like it lacks workflow id/type, eg, just with the fields I add, eg {"log":"{\"Domain\":\"clm-cadence-phx3\",\"TaskList\":\"clmTaskList\",\"WorkerID\":\"1@infradocker03-phx2@clmTaskList\",\"level\":\"debug\",\"msg\":\"zapify.go:66 Waiting for blacksmith to complete disable request\",\"requestID\":\"9
f7989f8-cae9-4273-80c0-d6c1c31bddce\",\"target\":\"agent1525-phx3\",\"time\":\"2019-02-01T01:05:04Z\",\"timeout\":600000000000}\n","stream":"stderr","time":"2019-02-01T01:05:04.824595464Z"}`

iainbecker:
Thanks, actually title should be "Add workflow fields to activity logger" ... looking at the example I posted the missing fields seem to be:

RunID
WorkflowID
WorkflowType

It would also be good to add an ActivityType field to activity logger like WorkflowType for workflow, that lists the full path to the activity type in question. Thanks!

[...]

PR #669, opened 2019-03-21
allow activity logger to have workflow id and run id

lihannan99:
This should fix https://github.com/uber-go/cadence-client/issues/637

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -272,6 +272,8 @@ func WithActivityTask(
 	}
 
 	logger.With(
+		zapcore.Field{Key: tagActivityID, Type: zapcore.StringType, String: *task.ActivityId},
+		zapcore.Field{Key: tagActivityType, Type: zapcore.StringType, String: *task.ActivityType.Name},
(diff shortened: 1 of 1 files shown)


internal/activity.go
```diff
@@ -269,6 +270,13 @@ func WithActivityTask(
 	} else {
 		deadline = startToCloseDeadline
 	}
+
+	logger.With(
+		zapcore.Field{Key: tagWorkflowType, Type: zapcore.StringType, String: *task.WorkflowType.Name},
+		zapcore.Field{Key: tagWorkflowID, Type: zapcore.StringType, String: *task.WorkflowExecution.WorkflowId},
+		zapcore.Field{Key: tagRunID, Type: zapcore.StringType, String: *task.WorkflowExecution.RunId},
+	)
```
mfateev: Does it already include activity type?

lihannan99: Good point, I will add that as well

lihannan99: Do we also want to add activity ID ?

lihannan99: I added activity ID and activity type to it.

mfateev: Is ActivityID ok? As the number of them is potentially unbounded.

PR #677, opened 2019-03-29
enable worker graceful shutdown v2

yux0:
This is the activity worker graceful shutdown.

Shutdown steps:
1. Stop polling all task
2. Terminate outstanding polling
3. Close a worker stop channel (Provide a new activity API so users can handle worker shutdown from activity).
[3.1] Flush out the batch heartbeat event
4. Wait for a specific timeout (User can provide it from worker options)
5. Cancel the top-level context.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1496,7 +1496,7 @@ func newServiceInvoker(
 	service workflowserviceclient.Interface,
 	cancelHandler func(),
 	heartBeatTimeoutInSec int32,
-	workerStopChannel <- chan struct{},
+	workerStopChannel <-chan struct{},
 ) ServiceInvoker {
 	return &cadenceInvoker{
 		taskToken:             taskToken,
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1375,7 +1375,6 @@ type cadenceInvoker struct {
 	service               workflowserviceclient.Interface
 	taskToken             []byte
 	cancelHandler         func()
-	retryPolicy           backoff.RetryPolicy
 	heartBeatTimeoutInSec int32       // The heart beat interval configured for this activity.
 	hbBatchEndTimer       *time.Timer // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
 	lastDetailsToReport   *[]byte
@@ -1449,7 +1448,14 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 
 func (i *cadenceInvoker) internalHeartBeat(details []byte) (bool, error) {
 	isActivityCancelled := false
-	err := recordActivityHeartbeat(context.Background(), i.service, i.identity, i.taskToken, details, i.retryPolicy)
+	timeout := time.Duration(i.heartBeatTimeoutInSec) * time.Second
+	if timeout <= 0 {
+		timeout = time.Duration(defaultHeartBeatIntervalInSec) * time.Second
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	err := recordActivityHeartbeat(ctx, i.service, i.identity, i.taskToken, details)
 
 	switch err.(type) {
 	case *CanceledError:
@@ -1495,7 +1501,6 @@ func newServiceInvoker(
 		identity:              identity,
 		service:               service,
 		cancelHandler:         cancelHandler,
-		retryPolicy:           serviceOperationRetryPolicy,
 		heartBeatTimeoutInSec: heartBeatTimeoutInSec,
 		closeCh:               make(chan struct{}),
 		workerStopChannel:     workerStopChannel,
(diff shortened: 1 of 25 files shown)


internal/internal_worker.go
```diff
@@ -1122,6 +1144,27 @@ func isInterfaceNil(i interface{}) bool {
...
+func getReadOnlyChannel(c chan struct{}) <-chan struct{} {
+	return c
+}
+
+func withWorkerStopChannel(parent context.Context, c <-chan struct{}) (context.Context, context.CancelFunc) {
+	if parent == nil {
+		parent = context.Background()
+	}
+
+	ctx := context.WithValue(parent, workerShutdownChannelContextKey, &c)
```
mfateev: Why is the shutdown channel is not part of the activityEnvironment structure?

internal/internal_task_handlers.go
```diff
@@ -1379,6 +1379,8 @@ type cadenceInvoker struct {
 	hbBatchEndTimer       *time.Timer // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
 	lastDetailsToReport   *[]byte
 	closeCh               chan struct{}
+	workerStopChannel     <-chan struct{}
+
```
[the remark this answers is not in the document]
yux0: Nit. Removed.

internal/internal_worker.go
yux0: The activity environment is generated in the Execute method in the activity task handler. So I use context to pass the channel the task handler. Another option is to put the channel in workerExecutionParameters and then pass it to the activity environment. Any concern to use context?

internal/internal_task_handlers.go
```diff
@@ -1417,6 +1419,13 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 			case <-i.closeCh:
 				// We got closed.
 				return
+			case <-i.workerStopChannel:
```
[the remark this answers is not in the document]
yux0: This line does flush the buffered heartbeats when the worker stop is called. So to be more proactive, do you mean to send all the heartbeat (after worker stop is called) without using buffered heartbeat?

internal/internal_worker.go
mfateev: I think workerExecutionParams makes more sense. I would prefer having all of the activity related data to be part of the activityEnvironment instead of multiple context keys.

yux0: Ok. Then I will move it to workerExecutionParameters.

internal/workflow_testsuite.go
```diff
@@ -170,6 +170,11 @@ func (t *TestActivityEnvironment) SetHeartbeatDetails(details interface{}) {
 	t.impl.setHeartbeatDetails(details)
 }
 
+// SetWorkerStopChannel sets the worker stop channel to be returned from activity.GetWorkerShutdownChannel()
+func (t *TestActivityEnvironment) SetWorkerStopChannel(c chan struct{}) {
```
mfateev: Inconsistent: SetWorkerStopChannel and GetWorkerShutdownChannel. Rename one or another.

internal/workflow_testsuite.go
```diff
@@ -385,6 +390,11 @@ func (t *TestWorkflowEnvironment) SetWorkerOptions(options WorkerOptions) *TestW
 	return t
 }
 
+// SetWorkerStopChannel sets the worker stop channel to be returned from activity.GetWorkerShutdownChannel()
+func (t *TestWorkflowEnvironment) SetWorkerStopChannel(c chan struct{}) {
```
mfateev: Why do we need this? I thought worklows do not have access to it.

internal/internal_workers_test.go
```diff
@@ -101,12 +121,67 @@ func (s *WorkersTestSuite) TestActivityWorker() {
...
+	}
+	overrides := &workerOverrides{activityTaskHandler: newNoResponseActivityTaskHandler()}
+	a := &greeterActivity{}
+	hostEnv := getHostEnvironment()
+	hostEnv.addActivity(a.ActivityType().Name, a)
+	activityWorker := newActivityWorker(
+		s.service, domain, executionParameters, overrides, hostEnv, stopChannel,
+	)
+	activityWorker.Start()
+	time.Sleep(1 * time.Second)
```
mfateev: Could you refactor the test to not rely on Sleep? It adds 2 seconds to the overall test suite execution.

internal/workflow_testsuite.go
yux0: Make sense. Updated all naming to WokerStopChannel

internal/workflow_testsuite.go
yux0: Yes. Workflows do not require it. The reason why I added in workflow test suite is that the workflow test suite runs activities as well. To provide this method so that users can run activities with TestWorkflowEnvironment.

internal/internal_workers_test.go
yux0: Good point. Updated this test to get rid of sleep.

internal/workflow_testsuite.go
mfateev: I see. Then update the comment to explain how this channel should be used, please.

PR #685, opened 2019-04-10
Add partial documentation to Selector interface

uberbo:
Please fill in the blanks.


internal/workflow.go
```diff
@@ -73,10 +73,22 @@ type (
...
-		AddReceive(c Channel, f func(c Channel, more bool)) Selector
-		AddSend(c Channel, v interface{}, f func()) Selector
-		AddFuture(future Future, f func(f Future)) Selector
-		AddDefault(f func())
+		AddReceive(c Channel, handlerFunc func(c Channel, more bool)) Selector
+		AddSend(c Channel, v interface{}, handlerFunc func()) Selector
+		// AddFuture adds a future to the selector,
+		// calling Get on the provided future inside the handlerFunc is guaranteed not to block,
+		// hence IsReady will return true.
+		// If AddFuture is called after Select, then ???.
```
yux0, line 81: If AddFuture is called after Select, the selector will not execute the handler function.
If AddFuture is called with the same future twice, the selector will be executed twice

internal/workflow.go
```diff
@@ -73,10 +73,22 @@ type (
...
+		// If AddFuture is called with the same future twice, then ???.
+		AddFuture(future Future, handlerFunc func(Future)) Selector
+		AddDefault(handlerFunc func())
+		// Select blocks until a Future or Channel is ready, it then calls the handlerFunc provided;
+		// unless a default is provided, then ???.
```
yux0, line 86: The selector should work the same as go select

yux0:
As you already touch this part, could you also add some doc on the other Selector interface?

yux0 (approved):

Issue #687, opened 2019-04-15
Domain is hardcoded in ReplayWorkflowHistory and ReplayWorkflowHistoryFromJSONFile #1645


lihannan99:
https://github.com/uber/cadence/issues/1645

lihannan99:
tried coconut package with 

```
package workflow_test

import (
	"code.uber.internal/infra/coconut/coconut-api/lifecycleworker"
	"github.com/stretchr/testify/suite"
	"go.uber.org/cadence/testsuite"
	"go.uber.org/cadence/worker"
	"go.uber.org/zap"
	"testing"
)

type UnitTestSuite struct {
	suite.Suite
	testsuite.WorkflowTestSuite
}

func TestUnitTestSuite(t *testing.T) {
	suite.Run(t, new(UnitTestSuite))
	// lifecycleworker.RegisterWorkflows()
}


func (s *UnitTestSuite) Test_DSLWorkflow_1() {
	lifecycleworker.RegisterWorkflows()
	logger, err := zap.NewDevelopment()
	s.NoError(err)
	err = worker.ReplayWorkflowHistoryFromJSONFile(logger, "dsl.json")
	s.NoError(err)
}

```

The test passed without any error.

PR #689, opened 2019-04-19
allow customizable domain for replay-domain

lihannan99:


lihannan99:
This should fix https://github.com/uber-go/cadence-client/issues/687

mfateev:
I think we need to fix the replay without requiring additional domain argument.

PR #690, opened 2019-04-22
fix invalid memory address or nil pointer dereference 

lihannan99:
When an error occurred in replayworkflow or the workflow is not completed 

user will get 
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x181ed7f]

goroutine 45 [running]:
code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal.(*workflowExecutionContextImpl).queueResetStickinessTask(...)
	/Users/scotthornberger/gocode/src/code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal/internal_task_handlers.go:442
code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal.(*workflowExecutionContextImpl).onEviction(0xc0000e3550)
	/Users/scotthornberger/gocode/src/code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal/internal_task_handlers.go:422 +0x1cf
code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal.getWorkflowCache.func1.1(0x1bc8ee0, 0xc0000e3550)
	/Users/scotthornberger/gocode/src/code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal/internal_task_handlers.go:355 +0x3c
created by code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal/common/cache.(*lru).Delete
	/Users/scotthornberger/gocode/src/code.uber.internal/infra/coconut/vendor/go.uber.org/cadence/internal/common/cache/lru.go:139 +0x175

```
This is due to missing `laChannel` in WorkflowTaskHandler when in replay. This will fix the issue. 

related to https://github.com/uber-go/cadence-client/issues/687

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -325,7 +325,7 @@ func newWorkflowTaskHandler(
 		hostEnv:                        hostEnv,
 		nonDeterministicWorkflowPolicy: params.NonDeterministicWorkflowPolicy,
 		dataConverter:                  params.DataConverter,
-		laTunnel: 						laTunnel,
+		laTunnel:                       laTunnel,
 	}
 }
 
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -309,6 +309,11 @@ func newWorkflowTaskHandler(
 	hostEnv *hostEnvImpl,
 ) WorkflowTaskHandler {
 	ensureRequiredParams(&params)
+	// laTunnel is needed for queueResetStickinessTask
+	laTunnel := &localActivityTunnel{
+		taskCh:   make(chan *localActivityTask, 1000),
+		resultCh: make(chan interface{}),
+	}
 	return &workflowTaskHandlerImpl{
 		domain:                         domain,
 		logger:                         params.Logger,
@@ -320,6 +325,7 @@ func newWorkflowTaskHandler(
 		hostEnv:                        hostEnv,
 		nonDeterministicWorkflowPolicy: params.NonDeterministicWorkflowPolicy,
 		dataConverter:                  params.DataConverter,
+		laTunnel: 						laTunnel,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


internal/internal_task_handlers.go
```diff
@@ -309,6 +309,11 @@ func newWorkflowTaskHandler(
 	hostEnv *hostEnvImpl,
 ) WorkflowTaskHandler {
 	ensureRequiredParams(&params)
+	// laTunnel is needed for queueResetStickinessTask
+	laTunnel := &localActivityTunnel{
+		taskCh:   make(chan *localActivityTask, 1000),
```
mfateev: What happens if someone runs 1500 local activities?

lihannan99: The exceeded activities will be blocked until there is more room in the buffered channel. 
There is a go routines in Baseworker's localActivityTaskPoller continues to pull the `taskCh` and then dispatch these tasks to localActivityWorker.

`newWorkflowTaskHandler` is only called in two places.  One is in `newWorkflowWorkerInternal` https://github.com/uber-go/cadence-client/blob/4cc749b964d61fd441fa67810d43683bb2455c03/internal/internal_worker.go#L244-L261

The other one is in `replayWorkflowHistory` 
https://github.com/uber-go/cadence-client/blob/4cc749b964d61fd441fa67810d43683bb2455c03/internal/worker.go#L265

In a "`normal`" case, `newWorkflowWorkerInternal` will create a `localActivityTunnel` with a buffered channel with 1000 slots. 

https://github.com/uber-go/cadence-client/blob/4cc749b964d61fd441fa67810d43683bb2455c03/internal/internal_worker.go#L291-L294

However, I saw the comment in here

https://github.com/uber-go/cadence-client/blob/d0cf9e09a98f1a9f6fc09fab3872923f4f8c8418/internal/internal_task_handlers.go#L888-L889

I think instead of tried to initiate a laTunnel here. we should  just check if `laTunnel` is `nil`.

https://github.com/uber-go/cadence-client/blob/d0cf9e09a98f1a9f6fc09fab3872923f4f8c8418/internal/internal_task_handlers.go#L442

PR #693, opened 2019-04-25
Add header propagation logic to client

### LINKED ISSUE #729: Context propagation support
Make use of server `Header` field to propagate context across workflows.

shreyassrivatsan:
* Add ContextPropagator interface that allows propagating selected context information across a workflow
* Add context propagator to code on the StartWorkflow, Activity paths. 

Fixes #729


internal/context.go
```diff
@@ -419,3 +426,15 @@ func (c *valueCtx) Value(key interface{}) interface{} {
 	}
 	return c.Context.Value(key)
 }
+
+func spanFromContext(ctx Context) opentracing.Span {
+	val := ctx.Value(activeSpanKey)
+	if sp, ok := val.(opentracing.Span); ok {
+		return sp
+	}
+	return nil
```
mfateev: Shouldn't we return NOOP Span instead?

internal/worker.go
```diff
@@ -163,6 +164,10 @@ type (
 		// Optional: worker graceful shutdown timeout
 		// default: 0s
 		WorkerStopTimeout time.Duration
+
+		// Optional: Sets ContextPropagators that allows users to control the context information passed through a workflow
+		// default: no ContextPropagators
+		ContextPropagators []ContextPropagator
```
mfateev: may be make the jaeger or open tracing implementation (if exists) a default?

internal/internal_worker_base.go
```diff
@@ -81,11 +82,12 @@ type (
 		IsReplaying() bool
 		MutableSideEffect(id string, f func() interface{}, equals func(a, b interface{}) bool) encoded.Value
 		GetDataConverter() encoded.DataConverter
+		GetContextPropagators() []ContextPropagator
 	}
 
 	// WorkflowDefinition wraps the code that can execute a workflow.
 	workflowDefinition interface {
-		Execute(env workflowEnvironment, input []byte)
+		Execute(env workflowEnvironment, header *shared.Header, input []byte)
```
mfateev: I would add header to the workflowEnvironment

internal/tracer_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/stretchr/testify/assert"
+	jaeger_config "github.com/uber/jaeger-client-go/config"
+	"go.uber.org/cadence/.gen/go/shared"
+)
+
+func TestTracingContextPropagator(t *testing.T) {
+	config := jaeger_config.Configuration{}
```
mfateev: So open tracing doesn't have its own implementation?

internal/workflow.go
```diff
@@ -846,6 +847,13 @@ func WithDataConverter(ctx Context, dc encoded.DataConverter) Context {
 	return ctx1
 }
 
+// WithContextPropagators adds ContextPropagators to the context.
+func WithContextPropagators(ctx Context, contextPropagators []ContextPropagator) Context {
```
mfateev: I don't think it should be public

internal/tracer_test.go
shreyassrivatsan: Unfortunately not. Had to pull on a tracing implementation just for the purpose of a test. They only have a noop tracer that does not implement the Inject, Extract methods we are interested in.

internal/worker.go
shreyassrivatsan: Will do that in the next change when I actually add the tracing support. Enabling tracing will add the default tracing context propagator.

internal/context.go
shreyassrivatsan: This is a copy of the open tracing implementation. Returns a nil in absence of an opentracing Span.

shreyassrivatsan:
@mfateev added a bunch more tests and responded to comments..

internal/context.go
mfateev: Does it mean that all Span using code always checks for this nil before using a Span?

shreyassrivatsan: Yes, but generally code (even RPC) doesnt need to deal with spans directly. They do something like.. 
```
span = opentracing.SpanFromContext(ctx)
newSpan = tracer.CreateSpan(ChildOf(span))
defer newSpan.Finish()
```

So the tracer is expected to handle that span can actually be nil..

internal/internal_workflow_testsuite.go
```diff
@@ -404,11 +406,16 @@ func (env *testWorkflowEnvironmentImpl) executeWorkflowInternal(delayStart time.
 		panic(err)
 	}
 	env.workflowDef = workflowDefinition
+	header := &shared.Header{
+		Fields: map[string][]byte{
+			"test-header": []byte("test-data"),
+		},
+	}
```
mfateev: What is the purpose of this header for workflow testing?

internal/internal_workflow_testsuite.go
```diff
@@ -463,6 +470,11 @@ func (env *testWorkflowEnvironmentImpl) executeActivity(
 		},
 		ActivityType: *activityType,
 		Input:        input,
+		Header: &shared.Header{
+			Fields: map[string][]byte{
+				testHeader: []byte("test-data"),
+			},
+		},
```
mfateev: What is the purpose of this header for activity testing?

internal/internal_workflow_testsuite.go
```diff
@@ -52,6 +52,7 @@ const (
 	defaultTestWorkflowTypeName = "default-test-workflow-type-name"
 	defaultTestDomainName       = "default-test-domain-name"
 	workflowTypeNotSpecified    = "workflow-type-not-specified"
+	testHeader                  = "test-header"
```
mfateev: I think customers should be able to inject any headers into the workflow under the test.

workflow/context_propagator.go
```diff
@@ -0,0 +1,110 @@
...
+	// ContextPropagator is an interface that determines what information from
+	// context to pass along
+	ContextPropagator = internal.ContextPropagator
+)
+
+// ContextKey to be used for the StringMapPropagator
+type ContextKey string
+
+// stringMapPropagator propagates the list of keys across a workflow,
+// interpreting the payloads as strings. This is provided as a sample for other
```
mfateev: Why not interpreting payloads as bytes?

internal/workflow_testsuite.go
```diff
@@ -68,6 +68,11 @@ type (
 		runFn        func(args mock.Arguments)
 		waitDuration func() time.Duration
 	}
+
+	// TestContextPropagator propagates test keys across a workflow
```
mfateev: Why do we need this? Can we just use the public one?

internal/internal_workflow_testsuite.go
shreyassrivatsan: Removed, made this configurable so user can use on tests..

internal/internal_workflow_testsuite.go
shreyassrivatsan: Removed, made this configurable so user can use on tests..

internal/internal_workflow_testsuite.go
shreyassrivatsan: Removed, made this configurable so user can use on tests..

internal/workflow_testsuite.go
shreyassrivatsan: Done

workflow/context_propagator.go
shreyassrivatsan: It's just a sample. The tracing propagator already has a sample of marshalling payloads into bytes, so wanted this sample to deal with strings..

mfateev: I'm not sure that sample should be part of the core library. May be move it into a samples repo?

shreyassrivatsan: Would you prefer if I leave it as internal and use it just for test purposes?

shreyassrivatsan: I'll move this to be a local struct that is used only by my test..

PR #695, opened 2019-04-30
Client Side Session Support

yycptt:
Session allows user to execute several activities on a single host


internal/internal_worker_base.go
```diff
@@ -209,6 +211,10 @@ func (bw *baseWorker) runPoller() {
 		case <-bw.shutdownCh:
 			return
 		case <-bw.pollerRequestCh:
+			if bw.sessionTokenBucket != nil {
```
yycptt: The worker is holding a token from pollerRequestCh when doing the long poll for activity task. Also the long poll can not be cancelled, as when client side called cancel(), server side will not be notified, unless an explicit call is made. Right now the solution is using another token bucket for sessions. Before poll, poller will check if there's available token. If so, do the poll without grabbing a token from the bucket. Token is only consumed in the create session activity.

internal/internal_worker.go
```diff
@@ -354,23 +371,83 @@ func (ww *workflowWorker) Stop() {
...
+	overrides *workerOverrides,
+	env *hostEnvImpl,
+	maxConCurrentSessionExecutionSize int,
+) Worker {
+	if params.Identity == "" {
+		params.Identity = getWorkerIdentity(params.TaskList)
+	}
+	if params.SessionResourceID == "" {
+		// TODO: Should we generate a uuid for user?
+		params.SessionResourceID = "Resource_" + uuid.New()
```
mfateev: What about using worker identity (with appended uuid if it is not unique across restarts)?

internal/activity.go
```diff
@@ -252,6 +252,7 @@ type ServiceInvoker interface {
 	// Returns ActivityTaskCanceledError if activity is cancelled
 	Heartbeat(details []byte) error
 	Close(flushBufferedHeartbeat bool)
+	SignalWorkflow(domain, wid, rid, signalName string, data []byte) error
```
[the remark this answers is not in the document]
yycptt: Makes sense. Updated. Also added clientOptions parameter to the method.

internal/internal_worker.go
yycptt: Yeah, it's better than hardcoding the "Resource" prefix. Updated.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+//    err = ExecuteActivity(sessionCtx, someActivityFunc, activityInput).Get(sessionCtx, nil)
+//    if err != nil {
+//        // Session has failed or activity itself failed.
+//    }
+//    ... // execute more activities using sessionCtx
+//    err = CompleteSession(sessionCtx)
+//    if err != nil {
+//        // Wrong ctx is used or failed to release session resource.
+//    }
+func CreateSession(ctx Context) (Context, error) {
```
mfateev: Let's add a SessionOptions struct as a parameter. For now let's add creationTimeout (which maps to stcheduleToStart sessionCreationActivity timeout) to it.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+// RecreateSession recreate a session based on the sessionInfo passed in. Activities executed within
+// the recreated session will be executed by the same worker as the previous session. RecreateSession()
+// returns an error under the same situation as CreateSession() and has the same usage as CreateSession().
+// It will not check the state of the session described by the sessionInfo passed in, so user can recreate
+// a session based on a failed or completed session.
+//
+// The main usage of RecreateSession is for long sessions that are splited into multiple runs. At the end of
+// one run, complete the current session, get sessionInfo from the context and pass the information to the
+// next run. In the new run, the session can be recreated on the information.
+func RecreateSession(ctx Context, sessionInfo *SessionInfo) (Context, error) {
```
mfateev: Let's rename to CreateSessionForResourceID.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+
+// CompleteSession completes a session. It releases worker resources, so other sessions can be created.
+//
+// After a session is completed, user can continue to use the context, but the activities will be scheduled
+// on the normal taskList (as user specified in ActivityOptions) and may be picked up by another worker since
+// it's not in a session.
+//
+// This API will return an error if the session in the context has already completed, or the resource release
+// process has failed. No error will returned if user tries to complete a session that has already failed and
+// the API call won't do anything.
```
mfateev: Remove error output

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+func CompleteSession(ctx Context) error {
+	return internal.CompleteSession(ctx)
+}
+
+// GetSessionInfo returns the sessionInfo stored in the context. If there are multiple sessions in the context,
+// (for example, the same context is used to create, complete, create another session. Then user found that the
+// session has failed, and created a new one on it), the most recent sessionInfo will be returned.
+//
+// This API will return an error if there's no sessionInfo in the context.
```
mfateev: I would make the API return pointer and no error.

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+	RegisterActivityWithOptions(sessionCreationActivity, RegisterActivityOptions{
+		Name: sessionCreationActivityName,
+	})
+	RegisterActivityWithOptions(sessionCompletionActivity, RegisterActivityOptions{
+		Name: sessionCompletionActivityName,
+	})
+}
+
+type (
+	sessionState string
```
mfateev: Let's make it enum

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+	errNoOpenSession            = errors.New("no open session in the context")
+	errNoSessionInfo            = errors.New("no session information found in the context")
+	errFoundExistingOpenSession = errors.New("found exisiting open session in the context")
+	errSessionFailed            = errors.New("session has failed")
+)
+
+// Note: Worker should be configured to process session. To do this, set the following
+// fields in WorkerOptions:
+//     EnableSessionWorker: true
+//     SessionResourceID: The identifier of the resource consumed by sessions.
```
mfateev: Add more information on how resourceID is expected to be used. For example for per process cache vs host file system.

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+func getResourceSpecificTasklist(identity, resourceID string) string {
+	return resourceID + "@" + identity
+}
+
+func sessionCreationActivity(ctx context.Context, sessionID string) error {
+	sessionEnv, ok := ctx.Value(sessionEnvironmentContextKey).(*sessionEnvironment)
+	if !ok {
+		panic("no session environment in context")
+	}
+	sessionEnv.Lock()
```
mfateev: Would you move most of this activity code to a method of the sessionEnvironment?

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+	sessionEnv.doneChanMap[sessionID] = doneChan
+	sessionEnv.Unlock()
+
+	defer func() {
+		sessionEnv.sessionTokenBucket.addToken()
+	}()
+
+	activityEnv := getActivityEnv(ctx)
+
+	if sessionEnv.testEnv == nil {
```
mfateev: I would remove testEnv from the sessionEnvironment and would create a version of Client that GetClient returns that delegates Signal to testEnv.

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+}
+
+func getCreationTasklist(base string) string {
+	return base + "__internal_session_creation"
+}
+
+func getResourceSpecificTasklist(identity, resourceID string) string {
+	return resourceID + "@" + identity
+}
+
```
mfateev: I would add debug level logging of session creation and other status changes.

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+		MaximumAttempts:    5,
+	}
+	ao := ActivityOptions{
+		ScheduleToStartTimeout: time.Second * 10,
+		StartToCloseTimeout:    time.Second * 10,
+		RetryPolicy:            retryPolicy,
+	}
+	completionCtx := WithActivityOptions(ctx, ao)
+	// the tasklist will be overrided to use the one stored in sessionInfo
+	err := ExecuteActivity(completionCtx, sessionCompletionActivityName, sessionInfo.SessionID).Get(ctx, nil)
```
mfateev: I would first cancel sessionCreationActivity to ensure that even the sessionCompletion fails the creationActivity exits on the next heartbeat. 
Then I would not call Get on the Future to not block the workflow on the completion.

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+		return nil, err
+	}
+
+	tasklistChan := GetSignalChannel(ctx, sessionID) // use sessionID as channel name
+	// Retry is only needed when creating new session and the error returned is NewCustomError(errTooManySessionsMsg)
+	retryPolicy := &RetryPolicy{
+		InitialInterval:    time.Second,
+		BackoffCoefficient: 1.1,
+		MaximumInterval:    time.Second * 10,
+		MaximumAttempts:    10,
```
mfateev: Make retryExpiration configurable. Do not specify maxAttempts.

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+		NonRetriableErrorReasons: []string{
+			"cadenceInternal:Panic",
+			"cadenceInternal:Generic",
+			"cadenceInternal:Timeout START_TO_CLOSE",
+			"cadenceInternal:Timeout HEARTBEAT",
+		},
+	}
+	// Creation activity need an infinite StartToCloseTimeout.
```
mfateev: I would introduce sessionExecutionTimeout and make it configurable through sessionOptions

internal/session.go
```diff
@@ -0,0 +1,421 @@
...
+	scheduleToStartTimeout := time.Second * 10
```
mfateev: scheduleToStart should be configurable through sessionCreationTimeout

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package workflow
+
+import (
+	"go.uber.org/cadence/internal"
+)
+
```
mfateev: I would add a built in workflow query similar to __stack_trace one to return info of all outstanding sessions

internal/internal_worker.go
```diff
@@ -354,23 +371,83 @@ func (ww *workflowWorker) Stop() {
...
+	params workerExecutionParameters,
+	overrides *workerOverrides,
+	env *hostEnvImpl,
+	maxConCurrentSessionExecutionSize int,
+) Worker {
+	if params.Identity == "" {
+		params.Identity = getWorkerIdentity(params.TaskList)
+	}
+	if params.SessionResourceID == "" {
+		// TODO: Should we generate a uuid for user?
```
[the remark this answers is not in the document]
yycptt: Removed TODO.

internal/internal_worker.go
```diff
@@ -69,6 +71,8 @@ const (
 
 	defaultPollerRate = 1000
 
+	defaultMaxConcurrentSeesionExecutionSize = 1000 // Large concurrent session execution size (1k)
```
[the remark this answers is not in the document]
yycptt: Fixed.

internal/internal_worker.go
```diff
@@ -39,6 +39,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/pborman/uuid"
+
```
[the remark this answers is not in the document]
yycptt: Done.

internal/internal_worker.go
```diff
@@ -354,23 +371,83 @@ func (ww *workflowWorker) Stop() {
...
+	if params.SessionResourceID == "" {
+		// TODO: Should we generate a uuid for user?
+		params.SessionResourceID = params.Identity + "-" + uuid.New()
+	}
+	sessionEnvironment := newSessionEnvironment(params.Identity, params.SessionResourceID, maxConCurrentSessionExecutionSize)
+
+	params.UserContext = context.WithValue(params.UserContext, sessionEnvironmentContextKey, sessionEnvironment)
+	params.ConcurrentPollRoutineSize = 1
+	params.TaskList = getCreationTasklist(params.TaskList)
+	creationWorker := newActivityWorker(service, domain, params, nil, env, sessionEnvironment.sessionTokenBucket)
```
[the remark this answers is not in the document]
yycptt: Passed override to newActivityWorker.

internal/worker.go
```diff
@@ -27,6 +27,8 @@ import (
 	"time"
```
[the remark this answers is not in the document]
yycptt: Done.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+//        (not closed and failed).
+//     2. The number of sessions currently running on the worker has reached
+//        MaxConCurrentSessionExecutionSize, which is specified when starting the worker.
+//
+// If an activity is executed using the returned context, it's regarded as part of the
+// session and all activities within a session will be executed by the same worker.
+// User still needs to handle the error returned when executing an activity. Session will
+// not be marked as failed if an activity within it returns an error. Only when the worker
+// executing the session is down, that session will be marked as failed. Executing an activity
+// within a failed session will return an [TODO] error immediately without even scheduling the activity.
```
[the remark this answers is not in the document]
yycptt: Done.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+//    sessionCtx, err := CreateSession(ctx)
+//    if err != nil {
+//		    // Creation failed. Wrong ctx or too many outstanding sessions.
+//    }
+//    err = ExecuteActivity(sessionCtx, someActivityFunc, activityInput).Get(sessionCtx, nil)
+//    if err != nil {
+//        // Session has failed or activity itself failed.
+//    }
+//    ... // execute more activities using sessionCtx
+//    err = CompleteSession(sessionCtx)
```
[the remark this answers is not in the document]
yycptt: Updated API.

workflow/session.go
yycptt: Since we need to hide resourceID from user for now, is it ok to keep this name?

workflow/session.go
yycptt: Done. SessionOptions currently has two fields: creationTimeout and executionTimeout.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+
+// RecreateSession recreate a session based on the sessionInfo passed in. Activities executed within
+// the recreated session will be executed by the same worker as the previous session. RecreateSession()
+// returns an error under the same situation as CreateSession() and has the same usage as CreateSession().
+// It will not check the state of the session described by the sessionInfo passed in, so user can recreate
+// a session based on a failed or completed session.
+//
+// The main usage of RecreateSession is for long sessions that are splited into multiple runs. At the end of
+// one run, complete the current session, get sessionInfo from the context and pass the information to the
+// next run. In the new run, the session can be recreated on the information.
```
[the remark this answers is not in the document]
yycptt: Created a new exported structure called RecreateSessionParams and use that as the argument. The params can be get by calling GetRecreateParams() on SessionInfo.

workflow/session.go
```diff
@@ -0,0 +1,112 @@
...
+//         support. By default, 1000 is used.
+
+// CreateSession creates a session and return a new context which contains information
+// of the created session. The session will be created on the tasklist user specified in
+// ActivityOptions. If none is specified, the default one will be used.
+//
+// CreationSession will fail in the following situations:
+//     1. The context passed in already contains a session which is still open
+//        (not closed and failed).
+//     2. The number of sessions currently running on the worker has reached
```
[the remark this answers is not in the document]
yycptt: Updated.

workflow/session.go
yycptt: Done.

workflow/session.go
yycptt: Done.

internal/session.go
yycptt: Done.

internal/session.go
yycptt: Leave this as a future work as now we need to hide resourceID from user.

internal/session.go
yycptt: I split the code into multiple methods call on sessionEnvironment.

internal/session.go
yycptt: I changed sessionEnvironment into an interface and provided two implementations. The test implementation will perform signal using testEnv.

internal/session.go
yycptt: Done.

internal/session.go
yycptt: Done.

internal/session.go
yycptt: Done. Use creationTimeout as expirationInterval.

internal/session.go
yycptt: Done.

internal/session.go
yycptt: Done.

workflow/session.go
yycptt: Done.

internal/session.go
```diff
@@ -0,0 +1,497 @@
...
+// Note: Worker should be configured to process session. To do this, set the following
+// fields in WorkerOptions:
+//     EnableSessionWorker: true
+//     SessionResourceID: The identifier of the resource consumed by sessions.
+//         It's the user's responsibility to ensure there's only one worker using this resourceID.
+//         This option is not available for now as automatic session reestablishing is not implemented.
+//     MaxConCurrentSessionExecutionSize: the maximum number of concurrently sessions the resource
+//         support. By default, 1000 is used.
+
+// CreateSession creates a session and return a new context which contains information
```
mfateev: and returns

workflow/session.go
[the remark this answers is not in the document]
mfateev: Update the documentation sample as well and switch it to defer.

workflow/session.go
```diff
@@ -0,0 +1,127 @@
...
+// returns an error under the same situation as CreateSession() and has the same usage as CreateSession().
+// It will not check the state of the session described by the sessionInfo passed in, so user can recreate
+// a session based on a failed or completed session.
+//
+// The main usage of RecreateSession is for long sessions that are splited into multiple runs. At the end of
+// one run, complete the current session, get recreateSessionParams from sessionInfo and pass the parameter to
+// next run. In the new run, the session can be recreated using the parameter.
+func RecreateSession(ctx Context, params *RecreateSessionParams, sessionOptions *SessionOptions) (Context, error) {
+	return internal.RecreateSession(ctx, params, sessionOptions)
+}
```
mfateev: Hide RecreateSession as it is not possible to recreate yet using the public APIs.

yycptt: I think it's possible to recreate a session. The RecreateSessionParams can be got by GetSessionInfo(ctx).GetRecreateParams().

internal/session.go
```diff
@@ -0,0 +1,499 @@
...
+		CreationTimeout  time.Duration
+	}
+
+	// RecreateSessionParams contains information needed to recreate a session on the same worker.
+	// Use SessionInfo.GetRecreateParameter() and pass the returned value to RecreateSession().
+	// The field of this struct is exported so that it can be pass to new run of a workflow and user
+	// should NOT handcraft this object. Always use GetRecreateParameter() and pass the returned value
+	// to the new run.
+	RecreateSessionParams struct {
+		Tasklist string
```
yycptt: I have to export the field so that the struct can be pass to the new run. User should always use GetRecreateParameter() as written in the comment.

workflow/session.go
```diff
@@ -0,0 +1,121 @@
...
+
+// RecreateSession recreate a session based on the sessionInfo passed in. Activities executed within
+// the recreated session will be executed by the same worker as the previous session. RecreateSession()
+// returns an error under the same situation as CreateSession() or the token passed in is invalid.
+// It also has the same usage as CreateSession().
+//
+// The main usage of RecreateSession is for long sessions that are splited into multiple runs. At the end of
+// one run, complete the current session, get recreateToken from sessionInfo by calling SessionInfo.GetRecreateToken()
+// and pass the token to the next run. In the new run, session can be recreated using that token.
+func RecreateSession(ctx Context, recreateToken []byte, sessionOptions *SessionOptions) (Context, error) {
```
yycptt: Use to recreateToken when recreating session.

PR #706, opened 2019-05-07
Create GetWorkflow method allowing to wait on workflow completion

### LINKED ISSUE #621: Capability to create WorkflowRun for existing workflow
Cadence client only supports creation of WorkflowRun through ExecuteWorkflow API.  Although this supports creating a WorkflowRun which could be used to wait on completion of already existing workflow, but there is no way to only support creation of WorkflowRun without first attempting to start the Workflow.
Basically I need an API like below:
```
    GetExecution(ctx context.Context, workflowID string, runID string) (WorkflowRun, error)
```
runID should be optional.

shreyassrivatsan:
New GetWorkflow method that retrieves workflow execution and allows waiting on it.
Fixes #621


internal/internal_workflow_client.go
```diff
@@ -260,6 +260,40 @@ func (wc *workflowClient) ExecuteWorkflow(ctx context.Context, options StartWork
...
 }
 
+// GetWorkflow gets a workflow execution and returns a WorkflowRun that will allow you to wait until this workflow
+// reaches the end state, such as workflow finished successfully or timeout.
+// The current timeout resolution implementation is in seconds and uses math.Ceil(d.Seconds()) as the duration. But is
+// subjected to change in the future.
+func (wc *workflowClient) GetWorkflow(ctx context.Context, workflowID string, runID string) (WorkflowRun, error) {
+
+	// retrieve the workflow execution information
+	resp, err := wc.DescribeWorkflowExecution(ctx, workflowID, runID)
```
vancexu: I think we might only need to call DescribeWorkflowExecution when runID is empty, so that user will be able to get not only the current run.

internal/internal_workflow_client.go
```diff
@@ -260,6 +260,40 @@ func (wc *workflowClient) ExecuteWorkflow(ctx context.Context, options StartWork
...
+		return nil, err
+	}
+
+	// get the current run ID
+	if resp.WorkflowExecutionInfo != nil &&
+		resp.WorkflowExecutionInfo.Execution != nil {
+		workflowID = resp.WorkflowExecutionInfo.Execution.GetWorkflowId()
+		runID = resp.WorkflowExecutionInfo.Execution.GetRunId()
+	} else {
+		return nil, errors.New("describeWorkflowExecution did not return workflowID and runID")
```
mfateev: Is it even possible?

internal/internal_workflow_client.go
mfateev: I think there is no need to call it at all as GetWorkflowHistory should work with empty runID.

client/client.go
```diff
@@ -95,6 +95,25 @@ type (
...
 
+		// GetWorkfow retrieves a workflow execution and return a WorkflowRun instance and error
+		// - workflow ID of the workflow.
+		// - runID can be default(empty string). if empty string then it will pick the last running execution of that workflow ID.
+		// The errors it can return:
+		//	- EntityNotExistsError, if domain does not exists
+		//	- BadRequestError
+		//	- InternalServiceError
+		//
+		// WorkflowRun has three methods:
```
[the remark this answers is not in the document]
shreyassrivatsan: done

client/client.go
```diff
@@ -95,6 +95,25 @@ type (
 		// NOTE: DO NOT USE THIS API INSIDE A WORKFLOW, USE workflow.ExecuteChildWorkflow instead
 		ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error)
 
+		// GetWorkfow retrieves a workflow execution and return a WorkflowRun instance and error
+		// - workflow ID of the workflow.
+		// - runID can be default(empty string). if empty string then it will pick the last running execution of that workflow ID.
+		// The errors it can return:
+		//	- EntityNotExistsError, if domain does not exists
```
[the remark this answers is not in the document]
shreyassrivatsan: update the comment.

internal/internal_workflow_client.go
shreyassrivatsan: What is the correct way to check for existence of workflow? I could just pass back a WorkflowRun object based of the workflowID and runID passed in, but after some validation perferable.. .

internal/internal_workflow_client.go
shreyassrivatsan: From the current implementation of server no, but just from the RPC perspective yes.. I can drop that check and assume response always exists..

client/client.go
```diff
@@ -94,14 +94,10 @@ type (
 		// NOTE: DO NOT USE THIS API INSIDE A WORKFLOW, USE workflow.ExecuteChildWorkflow instead
 		ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error)
 
-		// GetWorkfow retrieves a workflow execution and return a WorkflowRun instance (described above) and error
+		// GetWorkfow retrieves a workflow execution and return a WorkflowRun instance (described above)
```
mfateev: I would provide the full description even if it duplicates some of the data in the Execute document.

internal/client.go
```diff
@@ -86,6 +85,25 @@ type (
 		// NOTE: DO NOT USE THIS API INSIDE A WORKFLOW, USE workflow.ExecuteChildWorkflow instead
 		ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error)
 
+		// GetWorkfow retrieves a workflow execution and return a WorkflowRun instance and error
```
mfateev: Update this comment as well

PR #707, opened 2019-05-08
Correctly pass through WorkflowIDReusePolicy on SignalWithStart API

samarabbas:
Workflow client implementation does not pass on user specified
WorkflowIDReusePolicy to Cadence server on SignalWithStart API call
resulting in default policy 'WorkflowIDReusePolicyAllowDuplicate'
to be used for all calls.

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -343,6 +343,7 @@ func (wc *workflowClient) SignalWithStartWorkflow(ctx context.Context, workflowI
 		RetryPolicy:                         convertRetryPolicy(options.RetryPolicy),
 		CronSchedule:                        common.StringPtr(options.CronSchedule),
 		Memo:                                memo,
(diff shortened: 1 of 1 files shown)


venkat1109:
Can you also double-check if java client has the same problem ?

internal/internal_workflow_client.go
```diff
@@ -343,6 +343,7 @@ func (wc *workflowClient) SignalWithStartWorkflow(ctx context.Context, workflowI
 		RetryPolicy:                         convertRetryPolicy(options.RetryPolicy),
 		CronSchedule:                        common.StringPtr(options.CronSchedule),
 		Memo:                                memo,
+		WorkflowIdReusePolicy:               options.WorkflowIDReusePolicy.toThriftPtr(),
```
venkat1109: what do you think about adding unit tests to assert that we populate requests with correct values from higher level apis (by mocking service client) ?

samarabbas: Good point.  Let me land this first and then I will look into adding a unit test in a separate PR.

PR #713, opened 2019-05-14
Fix channel close bug

yycptt:
Fix https://github.com/uber-go/cadence-client/issues/710


internal/internal_workflow.go
```diff
@@ -695,9 +696,6 @@ func (c *channelImpl) Close() {
 		callback.fn(nil, false)
 	}
 	// All blocked sends are going to panic
-	for _, callback := range c.blockedSends {
-		callback.fn()
```
yycptt: I don't quite understand why we want to consider the value for blocked sends consumed when closing a channel.

PR #714, opened 2019-05-14
Add directions to enable prom style metrics reporting

shreyassrivatsan:
* Add comments to worker options
* Verify metrics functionality with prom style metrics

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -110,7 +110,7 @@ type (
 		Identity string
 
 		// Optional: Metrics to be reported. Metrics emitted by the cadence client are not prometheus compatible by
-		// default. To ensure metrics are compatible with prometheus make sure to create tally scope with sanitzer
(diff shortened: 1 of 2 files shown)


internal/worker.go
```diff
@@ -109,7 +109,32 @@ type (
 		// default: default identity that include hostname, groupName and process ID.
 		Identity string
 
-		// Optional: Metrics to be reported.
+		// Optional: Metrics to be reported. Metrics emitted by the cadence client are not prometheus compatible by
+		// default. To ensure metrics are compatible with prometheus make sure to create tally scope with sanitzer
```
mfateev: s/sanitzer/sanitizer

shreyassrivatsan: done

shreyassrivatsan:
@mfateev at this time, the only other solution I can think of is, dual emission of metrics to be prom compatible. I feel passing in a sanitized scope is neater though..

PR #716, opened 2019-05-16
Add List/Scan/Count Workflow APIs

vancexu:
This PR add 3 visibility APIs for SQL like query with ElasticSearch.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -243,7 +243,7 @@ type (
 		//  - "CloseTime between '2019-08-27T15:04:05+00:00' and '2019-08-28T15:04:05+00:00'".
 		//  - to list only open workflow use "CloseTime = missing"
 		// Retrieved workflow executions are sorted by StartTime in descending order when list open workflow,
-		// are sorted by CloseTime in descending order for other query.
+		// and sorted by CloseTime in descending order for other queries.
 		// The errors it can return:
 		//  - BadRequestError
 		//  - InternalServiceError
@@ -252,8 +252,9 @@ type (
 		// ScanWorkflow gets workflow executions based on query. This API only works with ElasticSearch,
 		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause
 		// (see ListWorkflow for query examples).
-		// ScanWorkflow is similar to ListWorkflow, but it is more efficient to retrieve large amount of workflows,
-		// and the result is not sorted.
+		// ScanWorkflow should be used when retrieving large amount of workflows and order is not needed.
+		// It will use more ElasticSearch resources than ListWorkflow, but will be several times faster
+		// when retrieving millions of workflows.
 		// The errors it can return:
 		//  - BadRequestError
 		//  - InternalServiceError
--- a/client/client.go
+++ b/client/client.go
@@ -236,6 +236,37 @@ type (
 		//  - EntityNotExistError
 		ListOpenWorkflow(ctx context.Context, request *s.ListOpenWorkflowExecutionsRequest) (*s.ListOpenWorkflowExecutionsResponse, error)
 
+		// ListWorkflow gets workflow executions based on query. This API only works with ElasticSearch,
+		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause,
+		// examples:
+		//  - "(WorkflowID = 'wid1' or (WorkflowType = 'type2' and WorkflowID = 'wid2'))".
+		//  - "CloseTime between '2019-08-27T15:04:05+00:00' and '2019-08-28T15:04:05+00:00'".
+		//  - to list only open workflow use "CloseTime = missing"
+		// Retrieved workflow executions are sorted by StartTime in descending order when list open workflow,
+		// are sorted by CloseTime in descending order for other query.
+		// The errors it can return:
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -220,6 +220,37 @@ type (
 		//  - EntityNotExistError
 		ListOpenWorkflow(ctx context.Context, request *s.ListOpenWorkflowExecutionsRequest) (*s.ListOpenWorkflowExecutionsResponse, error)
 
+		// ListWorkflow gets workflow executions based on query. This API only works with ElasticSearch,
+		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause,
+		// examples:
+		//  - "(WorkflowID = 'wid1' or (WorkflowType = 'type2' and WorkflowID = 'wid2'))".
+		//  - "CloseTime between '2019-08-27T15:04:05+00:00' and '2019-08-28T15:04:05+00:00'".
+		//  - to list only open workflow use "CloseTime = missing"
+		// Retrieved workflow executions are sorted by StartTime in descending order when list open workflow,
```
shreyassrivatsan: `when listing open workflows, and sorted by CloseTime in descending order for other queries`

client/client.go
```diff
@@ -220,6 +220,37 @@ type (
...
+
+		// ScanWorkflow gets workflow executions based on query. This API only works with ElasticSearch,
+		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause
+		// (see ListWorkflow for query examples).
+		// ScanWorkflow is similar to ListWorkflow, but it is more efficient to retrieve large amount of workflows,
+		// and the result is not sorted.
+		// The errors it can return:
+		//  - BadRequestError
+		//  - InternalServiceError
+		ScanWorkflow(ctx context.Context, request *s.ListWorkflowExecutionsRequest) (*s.ListWorkflowExecutionsResponse, error)
```
shreyassrivatsan: How do users know when to use what? Can we just make sorting an option on the request so and say that sorting will be expensive.. Seems like we should do that in the backend and only expose one API actually..

client/client.go
```diff
@@ -220,6 +220,37 @@ type (
...
+
+		// CountWorkflow gets number of workflow executions based on query. This API only works with ElasticSearch,
+		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause
```
shreyassrivatsan: All 3 descriptions say the query is basically the `SQL WHERE clause`, but one is with sorting, one just a count and one a simple where.. I understand `COUNT` is more performant than doing a list and counting on the client side, but was there a specific request for count as an API?

client/client.go
vancexu: Updated.

client/client.go
vancexu: Good question, I update comment to be more specific. 

It is not recommended by ES to use scroll for all search requests without order, so I have separate scan API.

> Scrolling is not intended for real time user requests, but rather for processing large amounts of data, e.g. in order to reindex the contents of one index into a new index with a different configuration.
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html

client/client.go
vancexu: Yes, so far I know 2 use case:
- Cadence Report 
- CLI. Oncall asked me multiple times how many open workflows luna has during some time range.

internal/client.go
```diff
@@ -205,6 +205,37 @@ type (
...
 		ListOpenWorkflow(ctx context.Context, request *s.ListOpenWorkflowExecutionsRequest) (*s.ListOpenWorkflowExecutionsResponse, error)
 
+		// ListWorkflow gets workflow executions based on query. This API only works with ElasticSearch,
+		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause,
+		// examples:
+		//  - "(WorkflowID = 'wid1' or (WorkflowType = 'type2' and WorkflowID = 'wid2'))".
+		//  - "CloseTime between '2019-08-27T15:04:05+00:00' and '2019-08-28T15:04:05+00:00'".
+		//  - to list only open workflow use "CloseTime = missing"
+		// Retrieved workflow executions are sorted by StartTime in descending order when list open workflow,
+		// are sorted by CloseTime in descending order for other query.
```
shreyassrivatsan: same change as in other file.

internal/client.go
```diff
@@ -205,6 +205,37 @@ type (
...
+		// The errors it can return:
+		//  - BadRequestError
+		//  - InternalServiceError
+		ListWorkflow(ctx context.Context, request *s.ListWorkflowExecutionsRequest) (*s.ListWorkflowExecutionsResponse, error)
+
+		// ScanWorkflow gets workflow executions based on query. This API only works with ElasticSearch,
+		// and will return BadRequestError when using Cassandra or MySQL. The query is basically the SQL WHERE clause
+		// (see ListWorkflow for query examples).
+		// ScanWorkflow is similar to ListWorkflow, but it is more efficient to retrieve large amount of workflows,
+		// and the result is not sorted.
```
shreyassrivatsan: same as above file

shreyassrivatsan:
Looks good, just a couple of `comment` comments..

PR #717, opened 2019-05-17
Set isReplay to true when clearing state

yycptt:
https://github.com/uber-go/cadence-client/issues/715


internal/internal_task_handlers.go
```diff
@@ -454,6 +454,8 @@ func (w *workflowExecutionContextImpl) clearState() {
 	w.previousStartedEventID = 0
 	w.newDecisions = nil
 	if w.eventHandler != nil {
+		// Set isReplay to true to prevent user code in defer guarded by !isReplaying() from running
+		w.eventHandler.isReplay = true
```
mfateev: I don't think this is going to work as this flag should be set when a thread is being cleaned up. Would you add a unit test to prove that it is indeed working correctly?

yycptt: I have verified this works locally. When stickiness is cleared, the flag will be set.

mfateev: This should work even when stickiness is off. Anyway we need a unit test to prove it.

yycptt: Sure. I will work on that.

PR #724, opened 2019-05-21
Add doc for activity with retry

longquanzheng:


internal/client.go
```diff
@@ -350,6 +350,10 @@ type (
 	}
 
 	// RetryPolicy defines the retry policy
+	// Note that the history of activity with retry policy will be different: the started event will be written down into
```
dmetzgar: Add a blank (comment) line or put a period on the end of the first sentence. Godoc will mash the lines together in one paragraph.

internal/client.go
```diff
@@ -350,6 +350,10 @@ type (
...
+	// history only when the activity completes or "finally" timeouts/fails. And the started event only records the last
+	// started time. Because of that, to check an activity has started or not, you cannot rely on history events. Instead,
+	// you can use CLI to describe the workflow to see the status of the activity.
```
dmetzgar: Put the name of the CLI command for describe workflow.

internal/client.go
longquanzheng: Period added. Blank line will break the lint check as it requires comment starting with "ReplyPolicy"

PR #728, opened 2019-05-23
Support SearchAttributes on start workflow

vancexu:
This PR support SearchAttributes features. Overall process is:

1. Users use GetSearchAttributes API to get a list of available key-valueType for SearchAttributes.
2. Users provide valid search attributes when start workflow.
3. Users use search attributes as query in list/scan/count workflow.


internal/internal_workflow_client.go
```diff
@@ -947,3 +977,19 @@ func getWorkflowMemo(input map[string]interface{}, dc encoded.DataConverter) (*s
 	}
 	return &s.Memo{Fields: memo}, nil
 }
+
+func getSearchAttributes(input map[string]interface{}) (*s.SearchAttributes, error) {
```
samarabbas: can you rename this to serialize search attributes?

PR #736, opened 2019-06-03
Add tracing support to client

shreyassrivatsan:
- Add tracer to client options and worker options
- Create tracing span for activities and local activities

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -28,6 +28,7 @@ import (
 	"reflect"
 	"time"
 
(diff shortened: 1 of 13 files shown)


internal/internal_workflow_client.go
```diff
@@ -178,6 +180,11 @@ func (wc *workflowClient) StartWorkflow(
 	if err != nil {
 		return nil, err
 	}
+
+	// create a workflow start span and attach it to the context object. finish it immediately
```
mfateev: May be add explanation why it is finished immediately

PR #741, opened 2019-06-06
Add more validation for replay

longquanzheng:


internal/worker.go
```diff
@@ -361,7 +362,10 @@ func replayWorkflowHistory(logger *zap.Logger, service workflowserviceclient.Int
 		Logger:   logger,
 	}
 	taskHandler := newWorkflowTaskHandler(domain, params, nil, getHostEnvironment())
-	_, _, err := taskHandler.ProcessWorkflowTask(&workflowTask{task: task, historyIterator: iterator})
+	resp, _, err := taskHandler.ProcessWorkflowTask(&workflowTask{task: task, historyIterator: iterator})
```
meiliang86: Seems to me that we should throw nondeterministic error inside ProcessWorkflowTask.
Can you check if https://github.com/uber-go/cadence-client/pull/709 solves this issue?

internal/worker.go
```diff
@@ -361,7 +368,32 @@ func replayWorkflowHistory(logger *zap.Logger, service workflowserviceclient.Int
...
+	if err != nil {
+		return err
+	}
+	err = fmt.Errorf("replay workflow doesn't return the same result as the last event, resp: %v, last: %v", resp, last)
```
meiliang86: Move this down. Also we should return different error in other if branches...

PR #752, opened 2019-06-13
Bug fix: Backoff retry should default to 2.0 without returning error

longquanzheng:


internal/internal_activity.go
```diff
@@ -194,7 +195,9 @@ func getValidatedActivityOptions(ctx Context) (*activityOptions, error) {
 	if p.HeartbeatTimeoutSeconds < 0 {
 		return nil, errors.New("invalid negative HeartbeatTimeoutSeconds")
 	}
-
+	if p.RetryPolicy.GetBackoffCoefficient() == 0 {
```
meiliang86: This logic should be in RetryPolicy itself. Otherwise other places that use this can still have the incorrect behavior.

longquanzheng: RetryPolicy is gen code we can't change it.
Seems like there is no better place to set this default value. Another place is when this RetryPlicy struct is created, but is outside of internal, it doesn't have access to this default value since it is internal pkg.

Issue #379, opened 2018-01-30
expose progress in other timeout errors


yiminc:
currently only heartbeat timeout expose the progress, we need the progress exposed in other type of timeout errors.

Issue #168, opened 2017-06-21
test framework: proper handle timeouts


yiminc:
our current test suite does not handle any timeouts (activity's heartbeat timeout, ScheduleToStart, StartToClose, ScheduleToClose, and workflow's ExecutionStartToClose, TaskStartToClose).

We need to update our test suite to properly handle those timeouts.

PR #759, opened 2019-06-18
Test framework: Keep timeout error details

yycptt:
Right now in the test framework, if user returns an timeout error in the activity (or using mock to return a timeout error), details of the timeout will be thrown away. This pr fixes this problem.

Also currently at server side the error reason (for retry policy) for timeout error uses the format "cadenceInternal:Timeout TimeOutType" (e.g. cadenceInternal:Timeout START_TO_CLOSE). However, it's not the case in the client side test framework. The error reason is always "cadenceInternal:Timeout", which means all timeout errors are always retriable.

Related Issue:
https://github.com/uber-go/cadence-client/issues/168
https://github.com/uber-go/cadence-client/issues/379

### DIFF
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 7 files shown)


internal/internal_utils.go
```diff
@@ -203,11 +204,20 @@ func getErrorDetails(err error, dataConverter encoded.DataConverter) (string, []
...
+		case *EncodedValues:
+			data = details.values
+		default:
+			panic("unknown error type")
+		}
 		if err0 != nil {
 			panic(err0)
 		}
-		return errReasonTimeout, data
+		return fmt.Sprintf("%v %v", errReasonTimeout, err.timeoutType), data
```
yycptt: Use the same format as server side: https://github.com/uber/cadence/blob/master/service/history/retry.go#L117

PR #760, opened 2019-06-18
Wait for local activity while holding workflow context lock

### LINKED ISSUE #742: Race condition of query and signal causing decision task timeout

shreyassrivatsan:
Move the wait for local activity to while holding the workflow context lock, to ensure that decision processing is serialized in the client

Fixes #742

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"runtime/debug"
 	"strings"
 	"sync"
 	"time"
@@ -459,6 +460,7 @@ func (w *workflowExecutionContextImpl) queueResetStickinessTask() {
 			RunId:      common.StringPtr(w.workflowInfo.WorkflowExecution.RunID),
 		},
 	}
+	debug.PrintStack()
 	// w.laTunnel could be nil for worker.ReplayHistory() because there is no worker started, in that case we don't
 	// care about resetStickinessTask.
 	if w.laTunnel != nil && w.laTunnel.resultCh != nil {
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -400,10 +400,6 @@ func (w *workflowExecutionContextImpl) Lock() {
 }
 
 func (w *workflowExecutionContextImpl) Unlock(err error) {
-	if _, ok := err.(*localActivityTimedOutError); ok {
-		w.mutex.Unlock()
-		return
-	}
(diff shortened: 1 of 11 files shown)


internal/internal_task_handlers.go
```diff
@@ -148,6 +148,10 @@ type (
 		currentIndex  int
 		next          []*s.HistoryEvent
 	}
+
+	localActivityTimedOutError struct {
```
mfateev: I find this name confusing. It looks like it indicates that a local activity timed out and caused this error. But this indicates the forced decision completion.

internal/internal_task_pollers.go
```diff
@@ -255,76 +254,32 @@ func (wtp *workflowTaskPoller) processWorkflowTask(workflowTask *workflowTask) e
...
-						return nil
-					}
-
-					// we are getting new decision task, so reset the workflowTask and continue process the new one
-					workflowTask = wtp.toWorkflowTask(response.DecisionTask)
-					continue process_WorkflowTask_Loop
-				}
+		if _, ok := err.(*localActivityTimedOutError); ok {
+			// force complete
+			response, err = wtp.forceRespondDecisionTaskCompleted(wc, workflowTask, startTime)
```
mfateev: I might be missing something, but I don't see how after this change the "heartbeating" of the workflow task is going to work.

internal/internal_task_handlers.go
shreyassrivatsan: I'll call this decision heartbeat error.

internal/internal_task_handlers.go
```diff
@@ -148,6 +148,10 @@ type (
 		currentIndex  int
 		next          []*s.HistoryEvent
 	}
+
+	decisionHeartBeatError struct {
```
mfateev: I believe Heartbeat is used everywhere in the code.

internal/internal_workers_test.go
```diff
@@ -331,6 +331,134 @@ func (s *WorkersTestSuite) TestLongRunningDecisionTask() {
...
+		break
+	case <-time.After(time.Second * 4):
+	}
+	worker.Stop()
+
+	s.True(isWorkflowCompleted)
+	s.Equal(2, localActivityCalledCount)
+}
+
+func (s *WorkersTestSuite) TestMultipleLocalActivities() {
```
mfateev: I don't see how it tests the race condition with the query. May be use the newly added integration tests that use Cadence docker?

shreyassrivatsan: This does not test the race condition (and technically does not add coverage), but is a better exercising of the local activity logic than just waiting for one.. 

I have not been able to get reliable repro for that - but will try adding a test with docker where I can force that scenario..

PR #764, opened 2019-06-19
Skip domain check for replayDomain for decision checking

longquanzheng:


internal/internal_task_handlers.go
```diff
@@ -1184,7 +1184,7 @@ func isDecisionMatchEvent(d *s.Decision, e *s.HistoryEvent, strictMode bool) boo
 		}
 		eventAttributes := e.RequestCancelExternalWorkflowExecutionInitiatedEventAttributes
 		decisionAttributes := d.RequestCancelExternalWorkflowExecutionDecisionAttributes
-		if eventAttributes.GetDomain() != decisionAttributes.GetDomain() ||
+		if (eventAttributes.GetDomain() != decisionAttributes.GetDomain() && !IsReplayDomain(decisionAttributes.GetDomain())) ||
```
mfateev: I'm confused. Why is domain populated in the decisionAttributes for CancelExternalWorkflowExecution if it is the same as the original domain of the workflow?

longquanzheng: Addressed offline also created https://github.com/uber-go/cadence-client/issues/766
Let's do the removal separately. I added check here allow domainName in the decisionAttribute to be empty.

PR #765, opened 2019-06-19
Add logs when tracing information is not propagated

shreyassrivatsan:


internal/tracer.go
```diff
@@ -62,16 +58,23 @@ func (t tracingWriter) Set(key, val string) {
...
+}
+
 func (t *tracingContextPropagator) Inject(
 	ctx context.Context,
 	hw HeaderWriter,
 ) error {
 	// retrieve span from context object
 	span := opentracing.SpanFromContext(ctx)
 	if span == nil {
+		t.logger.Info("could not retrieve span from context")
```
meiliang86: Potential noisy logs?

shreyassrivatsan: Maybe, I am not sure. i am going to leave it around for now and remove if I think its too noisy. But if someone has created a tracer and passed it into the client these should not fire and indicate a bad configuration somewhere else on users part.

PR #767, opened 2019-06-20
Also read span context from internal context key

shreyassrivatsan:


internal/tracer.go
```diff
@@ -74,7 +74,6 @@ func (t *tracingContextPropagator) Inject(
 	// retrieve span from context object
 	span := opentracing.SpanFromContext(ctx)
 	if span == nil {
-		t.logger.Info("could not retrieve span from context")
```
yycptt: Why we don't need these logging information?

shreyassrivatsan: These logs are noisy. I added them to see how much they will spew.. But if someone enables a tracer but doesnt start traces - which will be common in the start workflow scenario this will flood logs..

yycptt: Makes sense. Thanks for the explanation.

PR #768, opened 2019-06-21
Cancel session activities when session failed or completed

yycptt:


internal/session.go
```diff
@@ -224,12 +229,24 @@ func CompleteSession(ctx Context) {
...
+		// even though the creation activity has been cancelled, the session worker doesn't know. The worker will wait until
+		// next heartbeat to figure out that the workflow is completed and then release the resource. We need to make sure the
+		// completion activity is executed before the workflow exits.
+		// also run the activity in another coroutine so it won't block user workflow.
+		err := ExecuteActivity(completionCtx, sessionCompletionActivityName, sessionInfo.SessionID).Get(completionCtx, nil)
+		if err != nil {
+			GetLogger(completionCtx).Error("Complete session activity failed", zap.Error(err))
+		}
+		// the session context can only be cancelled after the completion activity finishes, otherwise the completion activity
+		// will be cancelled as well.
```
mfateev: I think we can have a separate context for the completion activity that is not cancelled when session context is cancelled. This way sessionContext is cancelled immediately. I think it is a requirement that it is cancelled as soon as CompleteSession returns.

yycptt: Makes sense. Then the context for the completion activity should be created when the session is created and stored in sessionInfo. When CompleteSession() is called, that context needs to be retrieved from sessionInfo and be used to execute the completion activity.

mfateev: Sounds good. Would you add that change?

yycptt: Sure. Will update the pr soon.

yycptt:
Sure. I will create another pr for the integration test.

PR #769, opened 2019-06-21
Add context propagation on child workflow and continue as new

shreyassrivatsan:


shreyassrivatsan:
> May be it is a time to add a unit test that relies on docker?

Yes, opened an issue for this #771

PR #780, opened 2019-07-01
Sync IDLs and fix all lint

longquanzheng:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -895,9 +910,7 @@ struct DomainConfiguration {
   10: optional i32 workflowExecutionRetentionPeriodInDays
   20: optional bool emitMetric
   30: optional string archivalBucketName
-  40: optional i32 archivalRetentionPeriodInDays
```
wxing1292: why these 2 are deleted?

@andrewjdawson2016

wxing1292: synced with Andrew, this is intentional, and since there is no one using it, this should be safe

vancexu:
I actually will have a PR related upsert search attributes, if not urgent would be good to wait me for that since there are multiple changes including tests related.

PR #781, opened 2019-07-01
Add upsert search attributes

vancexu:


internal/internal_workflow_testsuite.go
```diff
@@ -181,6 +181,8 @@ type (
 
 		workerStopChannel  chan struct{}
 		sessionEnvironment *testSessionEnvironmentImpl
+
+		searchAttributes map[string]interface{}
```
yycptt: Is it used somewhere?

internal/internal_event_handlers.go
```diff
@@ -288,6 +288,29 @@ func (wc *workflowEnvironmentImpl) SignalExternalWorkflow(domainName, workflowID
...
+	if err != nil {
+		return err
+	}
+
+	upsertID := wc.GenerateSequenceID()
+	wc.decisionsHelper.upsertSearchAttributes(upsertID, attr)
+	return nil
+}
+
+// This has to be used in workflowEnvironment implementations instead of in Workflow for testsuite mock purpose.
```
yycptt: I think this comment should be moved to L292.

internal/workflow.go
```diff
@@ -805,6 +806,43 @@ func signalExternalWorkflow(ctx Context, workflowID, runID, signalName string, a
...
+//   	"CustomIntField": 2,
+//   	"CustomBoolField": true,
+//   	"CustomKeywordField": "seattle",
+//   }
+// This is only supported when using ElasticSearch.
+func UpsertSearchAttributes(ctx Context, attributes map[string]interface{}) error {
+	ctx1 := setWorkflowEnvOptionsIfNotExist(ctx)
+	options := getWorkflowEnvOptions(ctx1)
+
+	if options.domain == nil || *options.domain == "" {
```
yycptt: I don't quite understand why we need this check?

vancexu: Good catch, I copied from other method but found no actually needed

internal/internal_workflow_testsuite.go
```diff
@@ -227,6 +227,8 @@ func newTestWorkflowEnvironmentImpl(s *WorkflowTestSuite) *testWorkflowEnvironme
 
 		doneChannel:       make(chan struct{}),
 		workerStopChannel: make(chan struct{}),
+
+		searchAttributes: make(map[string]interface{}),
```
yycptt: Remove this too.

PR #784, opened 2019-07-09
Support get search attributes inside workflow

vancexu:
This PR support get search attributes in workflow using GetInfo. 
Also support related test in testsuite.


internal/internal_event_handlers.go
```diff
@@ -297,9 +297,22 @@ func (wc *workflowEnvironmentImpl) UpsertSearchAttributes(attributes map[string]
 
 	upsertID := wc.GenerateSequenceID()
 	wc.decisionsHelper.upsertSearchAttributes(upsertID, attr)
+	wc.workflowInfo.SearchAttributes = mergeSearchAttributes(wc.workflowInfo.SearchAttributes, attr)
```
shreyassrivatsan: I dont like the idea of overriding the search attributes one off like this.. As part of processing the event we should have a cleaner way to update `WorkflowInfo`.. I'll need to think whats the best way to handle that..

internal/internal_event_handlers_test.go
```diff
@@ -164,10 +165,44 @@ func Test_ValidateAndSerializeSearchAttributes(t *testing.T) {
...
 	}
 	err := env.UpsertSearchAttributes(nil)
 	require.Error(t, err)
 
 	err = env.UpsertSearchAttributes(map[string]interface{}{"key": 1})
 	require.NoError(t, err)
 }
+
+func Test_MergeSearchAttributes(t *testing.T) {
+	var current *s.SearchAttributes
```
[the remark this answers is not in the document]
vancexu: done

internal/internal_event_handlers.go
vancexu: Ok, let me know if better option

shreyassrivatsan: Actually is this even the right place to have this? The user calls upsertSearchAttributes and that should cause a call to the server. The value on the client should only get updated while processing the event - not prior to the call and the server making the change.. Am I missing something?

vancexu: You are right. Changed to update in process event.

PR #792, opened 2019-07-18
Add default data converter

vancexu:


testsuite/testsuite.go
```diff
@@ -41,3 +42,8 @@ type (
 
 // ErrMockStartChildWorkflowFailed is special error used to indicate the mocked child workflow should fail to start.
 var ErrMockStartChildWorkflowFailed = internal.ErrMockStartChildWorkflowFailed
+
+// GetDefaultDataConverter return default data converter used by Cadence worker
```
mfateev: Why is it in testsuite and not encoded package?

mfateev:
I think we should allow setting defaultDataConverter like:

`encoded.SetDefaultDataConverter`

internal/encoded.go
```diff
@@ -0,0 +1,75 @@
...
+	"go.uber.org/cadence/encoded"
+)
+
+// defaultDataConverter uses thrift encoder/decoder when possible, for everything else use json.
+type defaultDataConverter struct{}
+
+var defaultJsonDataConverter = &defaultDataConverter{}
+
+// DefaultDataConverter is default data converter used by Cadence worker
+var DefaultDataConverter = getDefaultDataConverter()
```
mfateev: Why do you we need two of these variables?

vancexu: `defaultJsonDataConverter` is to avoid creating struct for each get call. 
`DefaultDataConverter` is just for exposing, since getDefaultDataConverter method is not exposed.

mfateev: I'm confused. Why you just don't change visibility of getDefaultDataConverter?

Issue #791, opened 2019-07-18
Add memo test support


vancexu:

vancexu:
https://github.com/uber-go/cadence-client/pull/792

PR #795, opened 2019-07-19
 Add get memo inside workflow and test support

vancexu:
Added memo in workflow.GetInfo. 
Support test memo with env.SetMemoOnStart.

https://github.com/uber-go/cadence-client/issues/791
https://github.com/uber-go/cadence-client/issues/745


internal/internal_task_handlers.go
```diff
@@ -544,6 +544,7 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *s.PollForDecisio
 		ContinuedExecutionRunID:             attributes.ContinuedExecutionRunId,
 		ParentWorkflowDomain:                attributes.ParentWorkflowDomain,
 		ParentWorkflowExecution:             parentWorkflowExecution,
+		Memo:                                attributes.Memo,
```
mfateev: Not sure if Info is the right place if memo becomes updatable.

PR #799, opened 2019-07-23
Implement wrapper for native go WaitGroup for use in cadence workflow Context

johncburns1:
Simple implementation of native go sync.WaitGroup using workflow Selector and workflow Future.  This is a first pass.  Let me know any changes that I should make.


johncburns1:
Need to produce some unit tests for coverage test.

internal/internal_workflow.go
```diff
@@ -130,6 +130,15 @@ type (
 		defaultFunc *func()       // default case
 	}
 
+	// Implements WaitGroup interface
+	waitGroupImpl struct {
+		n        int           // the number of coroutines to wait on
+		index    int           // the index of the future in the Selector to set in WaitGroup.Done()
+		ready    bool          // indicates that all coroutines have completed and WaitGroup.Wait() is ready to unblock
+		mu       sync.Mutex    // mutex lock used to safely increment index and n
```
mfateev: No need to lock in the workflow code as it is already fully synchronized.

johncburns1:
Good simplification. I'll make the changes when I'm back down at my workstation... And add unit test.

internal/internal_workflow.go
```diff
@@ -1356,60 +1355,52 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
+// param delta int32 -> the value to increment the WaitGroup counter by
+func (wg *waitGroupImpl) Add(delta int32) {
+	state := wg.waiting
 	wg.n = wg.n + delta
-
-	return wg
+	if wg.n < 0 {
+		panic("negative WaitGroup counter")
+	}
+	if (wg.waiting && delta > 0) && (wg.n == delta) {
```
mfateev: Move it before incrementing the counter

internal/internal_workflow.go
```diff
@@ -1356,60 +1355,52 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
+		panic("WaitGroup misuse: Add called concurrently with Wait")
+	}
+	if (wg.n > 0) || (!wg.waiting) {
+		return
+	}
+	if wg.waiting != state {
```
mfateev: This check is not necessary as preemption in Cadence workflow is voluntary. So it is not possible that this method is preempted while running.

internal/internal_workflow.go
```diff
@@ -1356,60 +1355,52 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
+	if wg.waiting {
+		panic("WaitGroup is reused before previous Wait has returned")
+	}
+
+	wg.waiting = true
+	if err := wg.future.Get(ctx, &wg.waiting); err != nil {
+		panic(err)
+	}
+	if !wg.waiting {
+		wg.future = &futureImpl{channel: NewChannel(ctx).(*channelImpl)}
```
mfateev: Why not NewFuture?

internal/internal_workflow.go
```diff
@@ -1356,60 +1355,52 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
-		}
+	if wg.waiting {
+		panic("WaitGroup is reused before previous Wait has returned")
+	}
+
+	wg.waiting = true
+	if err := wg.future.Get(ctx, &wg.waiting); err != nil {
+		panic(err)
+	}
+	if !wg.waiting {
```
mfateev: I don't understand this check. Is any place when the future can be set to ready with true value?

internal/workflow.go
```diff
@@ -308,10 +307,7 @@ func NewNamedSelector(ctx Context, name string) Selector {
 
 // NewWaitGroup creates a new WaitGroup instance.
 func NewWaitGroup(ctx Context) WaitGroup {
-	return &waitGroupImpl{
-		mu:       sync.Mutex{},
-		selector: NewSelector(ctx).(*selectorImpl),
-	}
+	return &waitGroupImpl{future: &futureImpl{channel: NewChannel(ctx).(*channelImpl)}}
```
mfateev: Why not NewFuture?

internal/internal_workflow.go
johncburns1: No there is not, just unnecessarily defensive .  I can take it out though since it is unreachable code and will decrease test coverage.

internal/workflow.go
johncburns1: I originally implemented using NewFuture(), but I did not want to deal with the returned Settable as well.  I can switch it back and ignore the Settable if it would be more readable.

internal/internal_workflow.go
johncburns1: I originally implemented using NewFuture(), but I did not want to deal with the returned Settable as well.  I can switch it back and ignore the Settable if it would be more readable.

internal/internal_workflow.go
johncburns1: Great I'll remove the last check.

internal/internal_workflow.go
mfateev: Future interface doesn't have Set method. So you'll have to store both Future and Settable and use the appropriate interface in Done and Wait methods.

johncburns1: Ok.  I will use the interfaces.  What is the reason for using the interfaces vs. the futureImpl struct?

mfateev: I think it is just cleaner to use public APIs.

johncburns1:
@mfateev I am experimenting with some irregular native go sync.WaitGroup behavior.  One thing that I noticed is that if the native WaitGroup is blocking and waiting on goroutines that will never finish or waiting on a number of goroutines but Done() is not called the specified number of times to increment the counter back to 0, WaitGroup.Wait() will throw a fatal error:  "fatal error: all goroutines are asleep - deadlock!".  Is this behavior that should be replicated or should the future.Get() just continue to block and wait for a timeout?

mfateev:
It looks like the deadlock detection is done by go runtime, not by WaitGroup implementation. It can be triggered by deadlocks not involving WaitGroup. So I think we don't need to match this functionality in the mapper. The good news is that Cadence support getting stack trace of any workflow which makes troubleshooting deadlocks simpler.

mfateev:
Filed https://github.com/uber-go/cadence-client/issues/803

internal/internal_workflow.go
```diff
@@ -1347,3 +1356,59 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
+// positive delta that start when the counter is greater than zero, may happen
+// at any time.
+// Typically this means the calls to Add should execute before the statement
+// creating the goroutine or other event to be waited for.
+// If a WaitGroup is reused to wait for several independent sets of events,
+// new Add calls must happen after all previous Wait calls have returned.
+//
+// param delta int32 -> the value to increment the WaitGroup counter by
+func (wg *waitGroupImpl) Add(delta int32) {
+	if (wg.waiting && delta > 0) && (wg.n == 0) {
```
mfateev: I don't think it is possible. Wait call is not going to block if n==0 .

johncburns1: It's not. Ran into this while writing unit tests. It's not a possible case in workflow context. I'll remove.

internal/internal_workflow.go
```diff
@@ -1347,3 +1356,59 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
+	}
+	if wg.waiting {
+		panic("WaitGroup is reused before previous Wait has returned")
+	}
+
+	wg.waiting = true
+	if err := wg.future.Get(ctx, &wg.waiting); err != nil {
+		panic(err)
+	}
+	wg.future, wg.settable = NewFuture(ctx)
```
mfateev: I would lazily initialize it in the Add to avoid unnecessary creation in the cases when the WaitGroup is not reused.

johncburns1: I like the change.  This would make it such that I would have to either store the context or pass a context to the Done() and Add() methods.  Would this make it less straightforward to use?

mfateev: Agree. Adding context to Add is excessive.

johncburns1: So I guess the alternative would be to store the context used to initialize the Wait Group, but this would keep one from adding custom context properties such as timeout when waiting on the future to unblock.

internal/internal_workflow.go
```diff
@@ -1347,3 +1356,55 @@ func (h *queryHandler) execute(input []byte) (result []byte, err error) {
...
+// must happen before a Wait. Calls with a negative delta, or calls with a
+// positive delta that start when the counter is greater than zero, may happen
+// at any time.
+// Typically this means the calls to Add should execute before the statement
+// creating the goroutine or other event to be waited for.
+// If a WaitGroup is reused to wait for several independent sets of events,
+// new Add calls must happen after all previous Wait calls have returned.
+//
+// param delta int32 -> the value to increment the WaitGroup counter by
+func (wg *waitGroupImpl) Add(delta int32) {
```
mfateev: Any specific reason the delta is int32? The sync.WaitGroup.Add is int.

johncburns1: If I remember correctly, it looked like in sync.WaitGroup() were only using 32 bits to store the counter. I am storing the counter as an int32 and I wanted that to be consistent with Add()

mfateev: This is the WaitGroup signature:

`func (wg *WaitGroup) Add(delta int) {
`
I think we want to have exactly the same API

johncburns1: OK I'll make the change.

PR #802, opened 2019-07-24
 Support memo and search attributes in child workflow

vancexu:


internal/workflow.go
```diff
@@ -198,6 +198,14 @@ type (
 		// │ │ │ │ │
 		// * * * * *
 		CronSchedule string
+
+		// Memo - Optional non-indexed info that will be showed in list workflow.
```
shreyassrivatsan: showed -> shown

PR #815, opened 2019-08-05
Fix docker file

yycptt:


vancexu:
@longquanzheng Let's discuss the tag we used in cadence-server docker. 
Talked with yichao, this PR will just a temp fix to unblock his bug fix and make master build pass

PR #825, opened 2019-08-20
Expose binaryChecksum as workflow.GetInfo

mfateev:
Can be used instead of workflow.GetVersion(). The advantage is that it works even in the cases when a backwards incompatible change was introduced and broke existing workflows. Adding the condition based on the binaryVersion is not going to break workflows that already made progress using the new code.


mfateev:
Yes, DecisionTaskCompleted events contain them.

PR #828, opened 2019-08-28
Fix non-deterministic error caused by upsert search attributes

vancexu:
Currently when replay with upsertSearchAttributes, decision will not match and cause non-deterministic error.

This PR fix the bug.


vancexu:
Build failed because of unrelated issue. Local tests past without concerns. 
Need to merge right now because of blocking customer. 

The build issue is caused by cadence docker for some reason not updated schema, and sometimes not even create keyspace or have connection to cassandra. 
Need more efforts to address later.

PR #834, opened 2019-09-03
Fix doc typos

sevein:
The docs on [Child Workflow](https://godoc.org/go.uber.org/cadence/workflow#hdr-Child_Workflow) use `workflow.WithChildWorkflowOptions` instead of `workflow.WildChildOptions`.

Also, replaced `time out` where it seemed that the the intention was to use the noun, `timeout`.


sevein:
I can't see the errors in the two builds failing. What did I do wrong?

PR #835, opened 2019-09-05
Fix documentation for `activity.GetInfo`

bjacobso:


sevein:
@bjacobso interesting that your build pass but mine (https://github.com/uber-go/cadence-client/issues/834) didn't!

PR #838, opened 2019-09-13
Add ListArchivedWorkflow API

yycptt:


vancexu:
LGTM but I think we need @mfateev to double confirm team is on same page about having archival visibility as separate APIs.

PR #840, opened 2019-09-14
Use go modules

ernado:
Change was produced by these steps:
1) Perform go mod init
2) Update Makefile
3) Remove excessive replace directives in go.mod
4) Run tests
5) Perform go mod tidy
6) Remove dep and glide files

Feel free to request any changes if needed.

If there are test-only, or mock-only dependencies, and we don't want them in the root `go.mod` file, we can workaround it with separate module and `replace packagename ./...` directive, so their dependencies will be separated.

Also, seems like `CONTRIBUTING.md` file should be also updated.


mfateev:
It looks like the change broke the integration tests :(

ernado:
> It looks like the change broke the integration tests :(

I'll try to investigate it. It looks like they are flaky now, and I'm not sure about the cause.
I have no access to CI logs, it was something like connection timeout?

mfateev:
Yes: 

```
FAIL: TestIntegrationSuite (30.44s) | 0s
-- | --
  | Error Trace:	integration_test.go:305
  | integration_test.go:89
  | suite.go:88
  | Error:      	Expected nil, but got: &yarpcerrors.Status{code:2, name:"", message:"received unknown error calling service: \"cadence-frontend\", procedure: \"WorkflowService::RegisterDomain\", err: dial tcp 172.18.0.4:7933: connect: connection refused"}

```

ernado:
It looks like the side effect of slowing down container spin-up because of go modules download is causing tests to fail.

Locally I was getting the same error when trying to `make tests` without starting the docker compose.
Now I'm getting consistently successful `make tests`.

Will investigate further.

ernado:
Got it!

I've added explicit check for tcp connectivity during integration suite setup, now they should be less flaky. 

Feel free to request any changes. 
Note that I've adjusted `docker/buildkite/docker-compose{-local}.yml` compose files to include the `GO111MODULE=on` in environment. We can avoid this if you are ok with updating `docker/buildkite/Dockerfile` to `go1.13`.

cc: @mfateev @samarabbas

mfateev:
Awesome! Let us do a few additional checks and I believe we would be able to accept it tomorrow.

PR #843, opened 2019-09-17
Parent close policy implementation

longquanzheng:


client/client.go
```diff
@@ -369,6 +372,24 @@ const (
...
+	p := internal.ParentClosePolicyRequestCancel
+	return &p
+}
+
+// ParentClosePolicyTerminate means terminating the child workflow
+func ParentClosePolicyTerminate() *internal.ParentClosePolicy {
+	p := internal.ParentClosePolicyTerminate
+	return &p
+}
+
```
mfateev: Why is this so different from IDReusePolicy? Why the function instead of a constant?

internal/internal_workflow.go
```diff
@@ -182,6 +182,7 @@ type (
 		contextPropagators                  []ContextPropagator
 		memo                                map[string]interface{}
 		searchAttributes                    map[string]interface{}
+		parentClosePolicy                   *ParentClosePolicy
```
mfateev: Why is it a pointer? Can we always have a value?

internal/workflow.go
```diff
@@ -210,6 +210,11 @@ type (
 		// supported when Cadence server is using ElasticSearch). The key and value type must be registered on Cadence server side.
 		// Use GetSearchAttributes API to get valid key and corresponding value type.
 		SearchAttributes map[string]interface{}
+
+		// ParentClosePolicy - Optional policy to decide what to do for the child, if its parent is closed(any closing status).
+		// The default behavior is defined by Cadence server. For now it is Abandon for backward compatibility.
+		// In future the default behavior may become Terminate for simplicity. So it's better not rely on the default behavior.
+		ParentClosePolicy *ParentClosePolicy
```
mfateev: I don't like this approach. Let's always specify a policy with default to terminate and have server override it for backwards compatibility.

internal/internal_workflow.go
longquanzheng: done

client/client.go
```diff
@@ -369,6 +372,17 @@ const (
 	WorkflowIDReusePolicyRejectDuplicate WorkflowIDReusePolicy = internal.WorkflowIDReusePolicyRejectDuplicate
 )
 
+const (
+	// Default subject to change in the server config
+	ParentClosePolicyDefault ParentClosePolicy = internal.ParentClosePolicyDefault
```
mfateev: Remove ParentClosePolicyDefault and default the value to ParentClosePolicyTerminate

longquanzheng: For backward compatibility, do you want to let server always override terminate to be abandon when seeing a policy is terminate no matter what? There could be cases that customers specify terminiate as policy and we won't be able to distinguish them if we make the default to terminate.

PR #844, opened 2019-09-18
QueryWorkflowV2 API to support filtering query based on close status

andrewjdawson2016:
The name QueryWorkflowV2 is a horrible API name but I am not sure what to call it.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -59,6 +59,12 @@ type (
 	// WorkflowIDReusePolicy defines workflow ID reuse behavior.
 	WorkflowIDReusePolicy = internal.WorkflowIDReusePolicy
 
+	// QueryWorkflowV2Request defines the request to QueryWorkflowV2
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -308,6 +314,15 @@ type (
...
 		QueryWorkflow(ctx context.Context, workflowID string, runID string, queryType string, args ...interface{}) (encoded.Value, error)
 
+		// QueryWorkflowV2 queries a given workflow execution and returns the query result synchronously.
+		// See QueryWorkflowV2Request and QueryWorkflowV2Response for more information.
+		// The errors it can return:
+		//  - BadRequestError
+		//  - InternalServiceError
+		//  - EntityNotExistError
+		//  - QueryFailError
+		QueryWorkflowV2(ctx context.Context, request QueryWorkflowV2Request) (QueryWorkflowV2Response, error)
```
andrewjdawson2016: This is a really bad API name but I am not sure what to call it. Let's discuss offline.

PR #847, opened 2019-09-23
Support customizing binary checksum

longquanzheng:


internal/internal_worker.go
```diff
@@ -73,6 +73,8 @@ const (
 	defaultMaxConcurrentSessionExecutionSize = 1000 // Large concurrent session execution size (1k)
 
 	testTagsContextKey = "cadence-testTags"
+
+	binaryChecksumEnvName = "CADENCE-CLIENT-WORKER-BINARY-CHECKSUM"
```
mfateev: Do we already expose any other configuration through the environment variable?

longquanzheng: Not as far as I know.How about let's remove this env variable here? 
I think the SetBinaryChecksum should be enough, the caller of worker.Start() could take care of it themselves.

mfateev: Yes, let's remove for consistency.

PR #848, opened 2019-09-24
Add searchable change version

vancexu:
https://github.com/uber/cadence/issues/2409

When using GetVersion API to update workflow implementation, users want to list all workflows with specific changeID plus version, so that they have better idea on workflows on different change, and can operate on them.

This PR makes use of search attributes's support on array data type. 
When GetVersion generate marker, it will also generate upsertSearchAttributes decision. 

To search, using query: `CadenceChangeVersion = '<changeID>-<version>'` such as  `CadenceChangeVersion = 'change1--1' or CadenceChangeVersion = 'change2-1'`  (version -1 is DefaultVersion)


internal/internal_event_handlers.go
```diff
@@ -560,6 +567,9 @@ func validateVersion(changeID string, version, minSupported, maxSupported Versio
 func (wc *workflowEnvironmentImpl) GetVersion(changeID string, minSupported, maxSupported Version) Version {
 	if version, ok := wc.changeVersions[changeID]; ok {
 		validateVersion(changeID, version, minSupported, maxSupported)
+		if wc.isReplay {
```
mfateev: if ok then this is not the first call to GetVersion with this changeID. Why do we need to Upsert in this case?

internal/internal_event_handlers.go
```diff
@@ -568,16 +578,36 @@ func (wc *workflowEnvironmentImpl) GetVersion(changeID string, minSupported, max
 		// GetVersion for changeID is called first time in replay mode, use DefaultVersion
 		version = DefaultVersion
 	} else {
-		// GetVersion for changeID is called first time (non-replay mode), we need to generate a marker decision for it.
+		// GetVersion for changeID is called first time (non-replay mode), generate a marker decision for it.
+		// Also upsert search attributes to enable ability to search by changeVersion.
 		version = maxSupported
 		wc.decisionsHelper.recordVersionMarker(changeID, version, wc.GetDataConverter())
+		wc.UpsertSearchAttributes(createSearchAttributesForChangeVersion(changeID, version, wc.changeVersions))
```
mfateev: Why are we upserting all the versions instead of the new one?

internal/internal_event_handlers.go
vancexu: Because when replay, marker will be processed first then code will enter this block, it needs to do same upsert as first GetVersion call. Otherwise replay will error.

internal/internal_event_handlers.go
vancexu: Because value of search attribute can only be override, we cannot append data to array type.  
(array type is treated same as other basic type)

vancexu:
Please do not land this since build failed. I am looking into the root cause and will provide a fix.

PR #852, opened 2019-10-03
Handle timeout error construct by the old client version

yycptt:
Old client version generate timeout error with error reason "CadenceInternal:Timeout" while reason generated by the current version is "CadenceInternal:Timeout TimeoutType". 

This pr enhanced backward compatibility so that the current client can recognize the timeout error reason generated by an old client.

### DIFF
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -229,7 +229,10 @@ func constructError(reason string, details []byte, dataConverter DataConverter) 
 		details := newEncodedValues(details, dataConverter)
 		timeoutType, err := getTimeoutTypeFromErrReason(reason)
 		if err != nil {
-			details.Get(&timeoutType)
+			// prior client version uses details to indicate timeoutType
+			if err := details.Get(&timeoutType); err != nil {
+				panic(err)
+			}
 			return NewTimeoutError(timeoutType)
 		}
 		return NewTimeoutError(timeoutType, details)
@@ -294,7 +297,7 @@ func getTimeoutTypeFromErrReason(reason string) (s.TimeoutType, error) {
 	timeoutTypeStr := reason[strings.Index(reason, " ")+1:]
 	var timeoutType s.TimeoutType
 	if err := timeoutType.UnmarshalText([]byte(timeoutTypeStr)); err != nil {
-		// this happens when the timeout error is constructed by an old client
+		// this happens when the timeout error reason is constructed by an prior constructed by prior client version
 		return 0, err
 	}
 	return timeoutType, nil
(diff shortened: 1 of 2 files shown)


internal/internal_utils.go
```diff
@@ -286,12 +290,12 @@ func getMetricsScopeForLocalActivity(ts *metrics.TaggedScope, workflowType, loca
...
 }
 
-func getTimeoutTypeFromErrReason(reason string) s.TimeoutType {
+func getTimeoutTypeFromErrReason(reason string) (s.TimeoutType, error) {
 	timeoutTypeStr := reason[strings.Index(reason, " ")+1:]
 	var timeoutType s.TimeoutType
 	if err := timeoutType.UnmarshalText([]byte(timeoutTypeStr)); err != nil {
-		// this should never happen
-		panic(err)
+		// this happens when the timeout error is constructed by an old client
```
mfateev: I'm confused. I thought timeout error is constructed by a server, not by the client.

yycptt: The timeout event is generated by the server. If a workflow gets an timeout error from an activity and return that error directly, it results in a workflow failed event. The error reason in this failed event is generated by the client side.

internal/internal_utils.go
```diff
@@ -286,12 +290,12 @@ func getMetricsScopeForLocalActivity(ts *metrics.TaggedScope, workflowType, loca
...
+		return 0, err
```
vancexu: Return 0 is a little bit weird, usually function should return pointer and in error case return nil pointer.

mfateev: I've seen a lot of code returning 0 and empty strings. IMHO if error is not nil the value doesn't matter at all.

internal/internal_utils.go
mfateev: I see. Would you rewrite comment to something like:
constructed by client prior version <version>

internal/internal_utils.go
```diff
@@ -226,8 +226,12 @@ func getErrorDetails(err error, dataConverter DataConverter) (string, []byte) {
 // constructError construct error from reason and details sending down from server.
 func constructError(reason string, details []byte, dataConverter DataConverter) error {
 	if strings.HasPrefix(reason, errReasonTimeout) {
-		timeoutType := getTimeoutTypeFromErrReason(reason)
 		details := newEncodedValues(details, dataConverter)
+		timeoutType, err := getTimeoutTypeFromErrReason(reason)
+		if err != nil {
+			details.Get(&timeoutType)
```
mfateev: Doesn't this call return error ?

yycptt: Good point. I've update the code to check the error. If the error is not nil, I think the only thing I can do is panic. 

I also checked old versions of the code, it seems like this error is never handled. I am not sure if that's deliberate or not. Maybe we want to avoid panic and use the default value as timeout type?

PR #859, opened 2019-10-10
Use the tagged logger instead of the original one

Groxx:
`logger.With` returns a new logger instance which will include the passed fields, it doesn't mutate the logger it's called on.
So previously this was effectively a no-op.


Groxx:
Supporting evidence, for completeness: https://github.com/uber-go/zap/blob/master/logger.go#L159

PR #864, opened 2019-10-17
Handle connection error to Cadence FE gracefully

adilhafeez:
With this change if cadence serve is not listening we would see error immediately rather than waiting forever.

Following is error you'd see if cadence is not running and you try to connect to it,

```
➜  cadence-samples git:(fail_fast) ✗ ./bin/helloworld
2019-10-17T00:27:16.939-0700	INFO	common/sample_helper.go:65	Logger created.
2019-10-17T00:27:16.940-0700	DEBUG	common/factory.go:131	Creating RPC dispatcher outbound	{"ServiceName": "cadence-frontend", "HostPort": "127.0.0.1:7933"}
isServiceTransientError
unknown error
2019-10-17T00:27:16.941-0700	FATAL	common/sample_helper.go:83	Error occurred while calling Describe API	{"Domain": "samples-domain", "error": "code:unknown message:received unknown error calling service: \"cadence-frontend\", procedure: \"WorkflowService::DescribeDomain\", err: dial tcp 127.0.0.1:7933: connect: connection refused"}
github.com/uber-common/cadence-samples/cmd/samples/common.(*SampleHelper).SetupServiceConfig
	/Users/ahafeez/src/go/src/github.com/uber-common/cadence-samples/cmd/samples/common/sample_helper.go:83
main.main
	/Users/ahafeez/src/go/src/github.com/uber-common/cadence-samples/cmd/samples/recipes/helloworld/main.go:40
runtime.main
	/usr/local/Cellar/go/1.12.9/libexec/src/runtime/proc.go:200
```

Without this change it would wait forever,

```
➜  cadence-samples git:(fail_fast) ✗ ./bin/helloworld
2019-10-17T00:27:07.908-0700	INFO	common/sample_helper.go:65	Logger created.
2019-10-17T00:27:07.914-0700	DEBUG	common/factory.go:131	Creating RPC dispatcher outbound	{"ServiceName": "cadence-frontend", "HostPort": "127.0.0.1:7933"}
^C
```

### DIFF
--- a/internal/internal_retry.go
+++ b/internal/internal_retry.go
@@ -24,8 +24,6 @@ package internal
 
 import (
 	"context"
-	"fmt"
-	"reflect"
 	"time"
 
 	s "go.uber.org/cadence/.gen/go/shared"
--- a/internal/internal_retry.go
+++ b/internal/internal_retry.go
@@ -24,10 +24,13 @@ package internal
 
 import (
 	"context"
+	"fmt"
+	"reflect"
 	"time"
 
 	s "go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/internal/common/backoff"
+	"go.uber.org/yarpc/yarpcerrors"
 )
 
 const (
@@ -72,6 +75,12 @@ func isServiceTransientError(err error) bool {
 		*s.DomainNotActiveError,
 		*s.CancellationAlreadyRequestedError:
 		return false
+	}
+
+	if errStatus, ok := err.(*yarpcerrors.Status); ok {
+		if yarpcerrors.IsUnknown(errStatus) {
+			return false
+		}
 	}
 
 	if err == errShutdown {


vancexu:
Thanks for the contribution!

There is an issue with this change: 
when server is temporarily unavailable, we do want to retry. 
So we should treat this yarpc unknown error as transient error. 

One alternative to solve the issue is to add some throttled log if necessary.

adilhafeez:
hey @vancexu thanks for the feedback. 

> when server is temporarily unavailable, we do want to retry.

We want to capture the case when upstream is not acceptation connection (not listening to port where OS is sending RST or not responding at all to initial connection request). In plain network application its easy to detect that by looking for ConnectionRefusedException. For yarpc I'm not sure if its possible to bubble that type of exception to client code. So closest thing I could find was the StautsUnknown exception from yarpc. Let me know if you have any additional insights here from yarpc.

> So we should treat this yarpc unknown error as transient error.

Before this code change this exception is treated as transient error. And client code would keep on retrying connection even thought there is no one listening on that port. We are explicitly checking for non-transient errors but for transient errors it seems like its a catch-all flow.

https://github.com/uber-go/cadence-client/blob/master/internal/internal_retry.go#L67-L73

> One alternative to solve the issue is to add some throttled log if necessary.

Yes, that would be nice I think.

samarabbas:
This feels a little risky to categorize all yarpc CodeUnknown errors as non transient errors.  This is a broad category and it is unfortunate yarpc decided to use 'CodeUnknown' for connection issues.  Even in that case we would like to retry even not being able to connect to Cadence server.
I think actual problem is upstream caller where it is calling this API without and timeout.

adilhafeez:
> This feels a little risky to categorize all yarpc CodeUnknown errors as non transient errors

I do agree with that. I do feel that yarpc can send more concrete exception types (such that connection timeout vs connection refused etc) so that we can decide whether it is a transient error nor not. For example connection timeout could be a transient error.

vancexu:
Hi @adilhafeez, it's been a while since last discussion. I want to close this for now, and feel free to reopen or open other PR if new ideas pop up. Thank you.

Issue #797, opened 2019-07-22
Support Saga


meiliang86:

davidroman0O:
Do we have any news on that?

PR #870, opened 2019-10-26
Implement Saga

mingfang:
https://github.com/uber-go/cadence-client/issues/797


workflow/saga.go
```diff
@@ -0,0 +1,62 @@
...
+        activity interface{}
+        args     []interface{}
+    }
+)
+
+func NewSaga(options SagaOptions) Saga {
+    return &sagaImpl{options: &options}
+}
+
+func (s *sagaImpl) AddCompensation(ctx Context, activity interface{}, args ...interface{}) Saga {
```
mfateev: I don't think we want to store context here. For example if workflow is cancelled the contexts created from the root context will be cancelled and the compensation code is not going to execute correctly. I propose to add ctx to Compensate call and use context created from it using Workflow.newDisconnectedContext.

workflow/saga.go
```diff
@@ -0,0 +1,62 @@
+package workflow
```
mfateev: I would move it into its own package

workflow/saga.go
```diff
@@ -0,0 +1,62 @@
...
+
+type (
+    SagaOptions struct {
+        ParallelCompensation bool
+        ContinueWithError    bool
+    }
+
+    Saga interface {
```
mfateev: I propose an alternative interface that uses ctx to pass saga around:

sagaCtx := saga.newSaga(ctx, options)
...
saga.addCompensation(ctx, ....)
...
saga.Compensate(ctx)

This approach works with calls to downstream functions without need to pass Saga object explicitly as Context is passed everywhere already.

vancexu:
Hi @mingfang , since it's been a while I wonder do you have any plans to address the feedbacks and move forward on this PR?

mingfang:
> Hi @mingfang , since it's been a while I wonder do you have any plans to address the feedbacks and move forward on this PR?

Sorry I don't have time to make the necessary changes. Please feel free to take over this PR or close it.

PR #871, opened 2019-10-27
Support consistency query

andrewjdawson2016:
- Add ability to specify query consistency level in request
- Handle piggybacked queries as part of RespondDecisionTaskCompleted


internal/internal_task_handlers.go
```diff
@@ -1403,13 +1407,35 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
 		forceNewDecision = false
 	}
 
+	// handle piggybacked queries
+	var queryResults map[string]*s.WorkflowQueryResult
```
meiliang86: Do we limit how many of these can there be in one decision task?

internal/internal_task_handlers.go
```diff
@@ -664,6 +664,10 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 		return nil, errors.New("nil or empty history")
 	}
 
+	if task.Query != nil && len(task.Queries) != 0 {
```
meiliang86: Will we be able to clean this up (removing Query) when we have the new one implemented?

meiliang86:
Are we using these anywhere? Do we have unit tests for these scenarios?

PR #882, opened 2019-11-16
Fix testsuite handling of context deadline error

mfateev:
Fixing unit test panicing when an activity context deadline is exceeded.


assafsch:
This is great :) Any idea when can we expect this in a future release?
Much appreciated

Issue #887, opened 2019-11-22
Activity return NewCancel error keeps getting retried


vancexu:
If activity (or local activity) code return NewCanceledError(detail) with details, workflow will keep retry this activity instead of keep moving. 

Possible bug on workflowexecution retryLocalActivity

vancexu:
This will be include in release v0.10.1

PR #890, opened 2019-11-26
Fix localactivity should not retry on CancelError

vancexu:
https://github.com/uber-go/cadence-client/issues/887


internal/internal_workflow_testsuite_test.go
```diff
@@ -2250,9 +2259,85 @@ func (s *WorkflowTestSuiteUnitTest) Test_LocalActivityRetry() {
...
-	s.Equal("retry-done", result)
+	s.Equal(int32(2), result)
+}
+
+func (s *WorkflowTestSuiteUnitTest) Test_LocalActivityRetryOnCancel() {
+	attempts := 0
+	localActivityFn := func(ctx context.Context) (int32, error) {
+		attempts++
+		info := GetActivityInfo(ctx)
+		if info.Attempt < 2 {
```
yycptt: Use `attempts` instead of `info.Attempt`?

vancexu: I think this `info.Attempt` can only be used inside activity. But I need a test var for assert equal.

PR #891, opened 2019-11-26
Adding binary checksum to RespondDecisionTaskFailedRequest

anish531213:


Groxx:
👍I just encountered this gap recently, seems useful.  Helps to identify bad builds, not just good ones.

PR #894, opened 2019-11-29
Log an error message when an activity results in an non-nil error

### LINKED ISSUE #888: Log activity failures
Currently activity failures are not logged through zap.Logger. Only panics are logged. It makes troubleshooting activity retries harder as they are not recorded to the history.

bbassingthwaite:
Fixes #888

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1720,7 +1720,14 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 	if <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
 		return nil, ctx.Err()
 	}
-
+	if err != nil {
+		ath.logger.Error("Activity error.",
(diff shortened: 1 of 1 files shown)


internal/internal_task_handlers.go
```diff
@@ -1720,7 +1720,14 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 	if <-ctx.Done(); ctx.Err() == context.DeadlineExceeded {
 		return nil, ctx.Err()
 	}
-
+	if err != nil {
+		ath.logger.Error("Activity error.",
+			zap.String(tagWorkflowID, t.WorkflowExecution.GetWorkflowId()),
+			zap.String(tagRunID, t.WorkflowExecution.GetRunId()),
+			zap.String(tagActivityType, activityType),
+			zap.String("Error", err.Error()),
```
mfateev: Replace with zap.Error(err)

bbassingthwaite: Fixed

PR #899, opened 2019-12-03
Parallelizing some tests

Groxx:
A chunk of work done to parallelize tests, with the ultimate goal
being to parallelize as much as possible to detect (and prevent)
race conditions.

This is not currently complete (far from it), but it's already useful,
 and I figured I could get it out right away.

I'm happy to drop commits or roll things back or split them apart,
just let me know!
The metrics tests in particular I'm thinking might be worth rewriting
entirely, in favor of tally.NewTestScope, but it would be a bit more
substantial so I haven't done that yet.


internal/common/cache/lru_test.go
```diff
@@ -82,23 +85,39 @@ func TestLRUCacheConcurrentAccess(t *testing.T) {
 		"D": "dank",
 		"E": "ezpz",
 	}
-
-	for k, v := range values {
-		cache.Put(k, v)
+	keys := make([]string, 0, len(values))
```
venkat1109: I believe this is no longer used. If yes, please remove this.

internal/common/metrics/scope_test.go
```diff
@@ -190,6 +162,25 @@ func newTaggedMetricsScope() (*TaggedScope, io.Closer, *capturingStatsReporter)
 	return &TaggedScope{Scope: scope}, closer, reporter
 }
 
+// withScope runs your callback twice, once for "during replay" and once for "after replay" / "executing".
+// stats are captured, and the results are returned for your validation.
+func withScope(t *testing.T, cb func(scope tally.Scope)) (replayed *CapturingStatsReporter, executed *CapturingStatsReporter) {
```
venkat1109: thanks for refactoring this. Much more readable now.

internal/common/cache/lru_test.go
Groxx: It is in use: https://github.com/uber-go/cadence-client/blob/master/internal/internal_task_handlers.go#L373

Pinning isn't, but the cache itself is.
(that said, I'm happy to separate this commit from the branch and we can optionally tackle it later - I'll set that up.)

Groxx: removed and pushed a new series of commits btw

Groxx:
rebased and pushed

PR #907, opened 2019-12-05
history_event_iterator: added sanity check to detect missing events

venkat1109:
This patch adds validation in history_iterator to detect cases when history is missing some events. Currently, the iterator is oblivious to missing events / holes during replay. While this is never supposed to happen, if it does happen, it can result in strange decisions / non-deterministic errors, which are typically hard to debug.


venkat1109:
> Is it possible to add a unit test for the negative case?

Done.

PR #910, opened 2019-12-11
Sticky task list name

andrewjdawson2016:
This fixes an issue where sticky task list name is not actually unique if multiple workers exist within the same process. This is most commonly an issue in tests.


internal/internal_task_pollers.go
```diff
@@ -207,6 +209,7 @@ func newWorkflowTaskPoller(
 		taskHandler:                  taskHandler,
 		metricsScope:                 params.MetricsScope,
 		logger:                       params.Logger,
+		stickyUUID:                   uuid.New(),
```
venkat1109: Based on my understanding, you can have multiple pollers per cadence-worker instance. So, creating an instance within the constructor of a poller would mean more than one sticky task list per worker/task-list. Can you double check this ? The best place is ideally when worker is initialized.

PR #913, opened 2019-12-13
Fix nil pointer exception when retrying local activity

yycptt:


internal/internal_task_handlers.go
```diff
@@ -460,6 +479,18 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
...
 }
 
+func (w *workflowExecutionContextImpl) getEventHandler() *workflowExecutionEventHandlerImpl {
+	eventHandler := w.eventHandler.Load()
+	if eventHandler == nil {
+		return nil
+	}
+	eventHandlerImpl, ok := eventHandler.(*workflowExecutionEventHandlerImpl)
+	if !ok {
+		return nil
```
wxing1292: this should panic

internal/internal_task_handlers.go
```diff
@@ -913,7 +944,7 @@ func (w *workflowExecutionContextImpl) ProcessLocalActivityResult(workflowTask *
 		return nil, nil // nothing to do here as we are retrying...
 	}
 
-	err := w.eventHandler.ProcessLocalActivityResult(lar)
+	err := w.getEventHandler().ProcessLocalActivityResult(lar)
```
wxing1292: should you do a nil check here?

internal/internal_task_handlers.go
```diff
@@ -776,7 +807,7 @@ func (w *workflowExecutionContextImpl) ProcessWorkflowTask(workflowTask *workflo
 	}
 	w.SetCurrentTask(task)
 
-	eventHandler := w.eventHandler
+	eventHandler := w.getEventHandler()
```
wxing1292: should you do a nil check here?

internal/internal_task_handlers.go
```diff
@@ -1007,42 +1048,45 @@ func (w *workflowExecutionContextImpl) CompleteDecisionTask(workflowTask *workfl
 	if w.currentDecisionTask == nil {
 		return nil
 	}
+	eventHandler := w.getEventHandler()
```
wxing1292: should you do a nil check here?

internal/internal_task_handlers.go
```diff
@@ -1007,42 +1048,45 @@ func (w *workflowExecutionContextImpl) CompleteDecisionTask(workflowTask *workfl
...
 
-	completeRequest := w.wth.completeWorkflow(w.eventHandler, w.currentDecisionTask, w, w.newDecisions, !waitLocalActivities)
+	completeRequest := w.wth.completeWorkflow(eventHandler, w.currentDecisionTask, w, w.newDecisions, !waitLocalActivities)
 	w.clearCurrentTask()
 
 	return completeRequest
 }
 
 func (w *workflowExecutionContextImpl) hasPendingLocalActivityWork() bool {
+	eventHandler := w.getEventHandler()
```
wxing1292: should you do a nil check here?

internal/internal_task_handlers.go
yycptt: I don't think so. This is called when processing of the decision task, during which the lock of execution context is held and no one else can set the eventHandler to nil. The callback for local activity retry is a special case and a bug we are trying to fix.

Also if eventHandler can be nil here then we are having a serious issue. 

Same thing for rest of the comments.

internal/internal_task_handlers.go
yycptt: Agree. Updated.

internal/internal_task_handlers.go
```diff
@@ -460,6 +479,18 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
 	w.mutex.Unlock()
 }
 
+func (w *workflowExecutionContextImpl) getEventHandler() *workflowExecutionEventHandlerImpl {
+	eventHandler := w.eventHandler.Load()
```
longquanzheng: Why mutex is not enough for race condition?

internal/internal_task_handlers.go
```diff
@@ -930,7 +961,17 @@ func (w *workflowExecutionContextImpl) retryLocalActivity(lar *localActivityResu
...
+			// currently this is solved by changing eventHandler to an atomic.Value. Ideally, this retry timer should be
+			// part of the event loop for processing the workflow task.
+			eventHandler := w.getEventHandler()
+
+			// if decision heartbeat failed, the workflow execution context will be cleared and eventHandler will be nil
+			if eventHandler == nil {
+				return
+			}
+
+			if _, ok := eventHandler.pendingLaTasks[lar.task.activityID]; !ok {
```
venkat1109: not related to your change - but looks like there could be a go-routine leak in laTunner.sendTask(lar.task) when localActivityTaskPoller shutsdown due to worker.Stop() - the fix would be to have sendTask select on {taskCh, stopC} - consider fixing this

internal/internal_task_handlers.go
yycptt: Synced offline on this. I agree that we are misusing the lock. But fixing that requires rewriting lots of task processing logic and is a long term plan. This pr just a temp bug fix for preventing our customer running into this situation again.

internal/internal_task_handlers.go
yycptt: Makes sense. I will create another pr to address this issue.

yycptt: https://github.com/uber-go/cadence-client/pull/918

PR #918, opened 2019-12-18
Fix potential go-routine leak when retrying local activity

yycptt:
When retrying local activity a delayed callback will be scheduled to send the local activity task to task handler through the local activity tunnel. However, it's possible that when the callback is executed, the task handler has already shut down. If the task channel is already full, then it will cause the callback to be blocked and results in a go-routine leak. This pr fixes that.

Also fixed a lint issue which makes `make bins` fail.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -976,7 +976,10 @@ func (w *workflowExecutionContextImpl) retryLocalActivity(lar *localActivityResu
 			}
 
 			lar.task.attempt++
-			w.laTunnel.sendTask(lar.task)
+
+			if !w.laTunnel.sendTask(lar.task) {
+				lar.task.attempt--
+			}
 		})
 		return true
 	}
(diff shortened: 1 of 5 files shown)


yux0:
lgtm.

internal/internal_task_handlers.go
```diff
@@ -976,7 +976,10 @@ func (w *workflowExecutionContextImpl) retryLocalActivity(lar *localActivityResu
 			}
 
 			lar.task.attempt++
-			w.laTunnel.sendTask(lar.task)
+
+			if !w.laTunnel.sendTask(lar.task) {
```
venkat1109: Consider changing this to positive assertion:
```
if w.laTunnel.sendTask(..) {
    lar.task.attempt++
}
```

yycptt: I am not sure about this. If the processing logic uses the attempt field, then changing its value after the task is sent will lead to a race condition I think.

PR #924, opened 2019-12-30
Add signal option to signal API

longquanzheng:


internal/internal_workflow_client.go
```diff
@@ -308,7 +313,7 @@ func (wc *workflowClient) GetWorkflow(ctx context.Context, workflowID string, ru
 }
 
 // SignalWorkflow signals a workflow in execution.
-func (wc *workflowClient) SignalWorkflow(ctx context.Context, workflowID string, runID string, signalName string, arg interface{}) error {
+func (wc *workflowClient) SignalWorkflow(ctx context.Context, workflowID string, runID string, signalName string, arg interface{}, opts ...SignalWorkflowOption) error {
```
venkat1109: if we expose options, we should try and stick to the standard options pattern in Go 
see example here - https://gist.github.com/travisjeffery/8265ca411735f638db80e2e34bdbd3ae

longquanzheng: @mfateev I think this is reasonable suggesion. We also want to hear if you have any more ideas.

mfateev: We started from this options standard pattern. The user feedback was very negative. Everyone asked for a structure.

internal/internal_workflow_client.go
```diff
@@ -130,6 +130,11 @@ type (
 		// func which use a next token to get next page of history events
 		paginate func(nexttoken []byte) (*s.GetWorkflowExecutionHistoryResponse, error)
 	}
+
+	SignalWorkflowOption struct {
+		// RequestedID is for dedupping signals
+		RequestedID string
```
mfateev: This is very dangerous feature. Currently we don't perform automatic deduping because we don't want to keep id of every signal ever received in mutable state. And this is not enough as deduping should work across continue-as-new.

longquanzheng: This is mainly for https://github.com/uber/cadence/issues/2934 as a safe and less costly way to fix the issue. Do you suggest to switch to regular activity(which would be non-deterministic changes and cost multiple round trips)? or waiting for the feature to use marker for signal self?

mfateev: How are you going to perform deduping without keeping id of every signal in the mutable state?
What about just permitting calling SignalExternalWorkflowExecution on itself?

longquanzheng: Yes today we have implemented it in mutable state. It could be a problem if you keep sending signals with different requestedID. https://github.com/uber/cadence/blob/master/service/history/historyEngine.go#L1782 it's only being used by SignalExternalWorkflowExecution.
I think the correct way is to do it like we did for auto-reset points, have a MaxSignalRequestedIDPerWorkflow dynamic config, default to a number like 20. If the size exceeds that, then do same rotation. And also passing along continueAsNew. 
Most of the time, requestedID is for temporary retry so it doesn't have to keep all of them. 

Using SignalExternalWorkflowExecution will be slightly better than regular activity. Only the name is confusing. Maybe we should make another SignalSelf() API, implement it in the write way -- make it as part of the decision in RespondDecisionTaskCompleted. 

I think having that dedup in requestedID is still useful even we have SignalSelf().

mfateev: The problem is that this solution works in most cases, but is going to fail in some relatively rare cases. It would be nice to design something which is 100% reliable.

longquanzheng: That's true. How about we always clear the requestedIDs list in mutable state every time a decision is completed? This would allow sending signals with the same requestedID for different decision. So the dedup is per decision instead of per workflow lifetime. 

I think this is Okay because today our usage of requestedID in SignalExternalWorkflow is even weaker: It always remove the requestedID when a transfer task completed. 
See https://github.com/uber/cadence/blob/master/service/history/transferQueueActiveProcessor.go#L716

It will still help the issue uber/cadence#2934 because we only clear the requestedIDs on decision completed no decision failed/timeout.

mfateev: Instead of this feature I would allow calling SignalExternalWorkflowExecution to self. Obviously we can provide less confusing client side api for this.

venkat1109: I second Maxim's suggestion to keep it simple i.e. support client side signalToSelf feature

longquanzheng: Convinced.  Let me drop this PR .

longquanzheng:
Close with issue https://github.com/uber-go/cadence-client/issues/927

internal/internal_workflow_client.go
longquanzheng: added issue for tracking this https://github.com/uber-go/cadence-client/issues/927

PR #929, opened 2020-01-14
encoded: Unalias interfaces

### LINKED ISSUE #928: encoded: Unalias interfaces
The [go.uber.org/encoded](https://godoc.org/go.uber.org/cadence/encoded) package declares three interfaces as
aliases of `internal` package versions of those interfaces.

```go
package encoded

type DataConverter = internal.DataConverter
type Value = internal.Value
type Values = internal.Values
```

This is both, unnecessary and undesirable.

It is unnecessary because interfaces in Go are implicit. Unless you're
doing something very specialized, where interface identity matters more
than interface compliance, you can just redeclare the interfaces in the
`encoded` package and things will continue to work.

```go
package encoded

type Value interface {
  HasValue() bool
  Get(valuePtr interface{}) error
}
```

Any value that satisfies `internal.Value` will also satisfy
`encoded.Value`.

You can even add a bidirectional compile-time check to verify that the
interface doesn't go out of sync with its internal variant.

```go
var (
  _ Value = (internal.Value)(nil)
  _ internal.Value = (Value)(nil)
)
```

It is undesirable because it makes it difficult for users to use the
interface. They cannot just look at the godocs; they have to dig into
the internal packages. It also negatively affects the behavior of
tooling like mockgen (https://github.com/golang/mock/issues/244).

---

Given this, is there a strong reason to keep the aliases? Is interface
identity necessary or will compliance suffice?

abhinav:
Per reasoning in #928, this redeclares all three interfaces in the
`encoded` package and adds bi-directional compile-time checks to verify
that the interfaces do not fall out-of-sync.

The only reason I could find for the interfaces to be aliased was
`client.Client`'s re-use of an `internal.Client`.

    package client

    type Client interface {
      ...
      QueryWorkflow(...) (encoded.Value, error)
    }

    func NewClient(...) Client {
      return internal.NewClient(...)
    }

Where,

    package internal

    type Client interface {
      ...
      QueryWorkflow(...) (internal.Value, error)
    }

This mismatch in `QueryWorkflow`'s return value makes `internal.Client`
not match `client.Client`.

To resolve this, we introduce a `clientAdapter` which adapts an
`internal.Client` into a `client.Client`. It does so by embedding
`internal.Client` and implementing the compliant version of
`QueryWorkflow`.

Resolves #928

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -399,7 +399,7 @@ const (
 
 // NewClient creates an instance of a workflow client
 func NewClient(service workflowserviceclient.Interface, domain string, options *Options) Client {
-	return internal.NewClient(service, domain, options)
+	return &clientAdapter{internal.NewClient(service, domain, options)}
 }
 
 // NewDomainClient creates an instance of a domain client, to manage lifecycle of domains.
@@ -407,9 +407,20 @@ func NewDomainClient(service workflowserviceclient.Interface, options *Options) 
 	return internal.NewDomainClient(service, options)
 }
 
+// clientAdapter adapts an internal.Client into a client.Client.
+//
+// We need this because the internal client's QueryWorkflow relies on the
+// interface internal.Value, while client.Client relies on encoded.Value.
+// Although the interfaces are equivalent, they're not the same type, so we
+// need to re-implement the method.
+type clientAdapter struct{ internal.Client }
+
+func (a *clientAdapter) QueryWorkflow(ctx context.Context, workflowID string, runID string, queryType string, args ...interface{}) (encoded.Value, error) {
+	return a.Client.QueryWorkflow(ctx, workflowID, runID, queryType, args...)
+}
+
 // make sure if new methods are added to internal.Client they are also added to public Client.
-var _ Client = internal.Client(nil)
-var _ internal.Client = Client(nil)
+var _ Client = (*clientAdapter)(nil)
 var _ DomainClient = internal.DomainClient(nil)
 var _ internal.DomainClient = DomainClient(nil)
 
--- a/encoded/encoded.go
+++ b/encoded/encoded.go
@@ -26,10 +26,20 @@ import "go.uber.org/cadence/internal"
 type (
 
 	// Value is used to encapsulate/extract encoded value from workflow/activity.
-	Value = internal.Value
+	Value interface {
+		// HasValue return whether there is value encoded.
+		HasValue() bool
+		// Get extract the encoded value into strong typed value pointer.
+		Get(valuePtr interface{}) error
+	}
 
 	// Values is used to encapsulate/extract encoded one or more values from workflow/activity.
-	Values = internal.Values
+	Values interface {
+		// HasValues return whether there are values encoded.
+		HasValues() bool
+		// Get extract the encoded values into strong typed value pointers.
+		Get(valuePtr ...interface{}) error
+	}
 
 	// DataConverter is used by the framework to serialize/deserialize input and output of activity/workflow
 	// that need to be sent over the wire.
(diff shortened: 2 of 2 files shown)


linzhp:
@vitarb This avoid the mocking aliased type issues we ran into during build

abhinav:
CI appears to be misconfigured. The other jobs haven't fired.

vancexu:
Is the major reason for this change for unblocking "mocking aliased type issues we ran into during build" ? If so do we have other work around to disable that gomock? I am not very sure what build and why it needs mock.

The major reason for using alias is those structs or interfaces are identical. Every non-internal package currently is using alias to internal, not only encoded. So this change will make it a little bit inconsistent, also requires a little bit more work when update. 

@vitarb what's your input on this?

vancexu:
> CI appears to be misconfigured. The other jobs haven't fired.

@meiliang86 Are you the only github client repo admin now? Can you update he build pipeline to remove those duplicates pending checks?

vitarb:
I would prefer to not introduce duplication if possible because it makes it makes things more confusing and harder to maintain in the long run. Are there any reasons why we don't what to do what Zhongpeng suggested [here](https://github.com/uber-go/cadence-client/issues/928#issuecomment-574295383) and simply move types out of the internal package?

mfateev:
I"m not sure how Zhongpeng suggestion removes duplication. I understand it as making two copies of the same interface. 

The alternative is to move it back as it was to an interface in public package and concrete struct that implements it in the internal one.

vitarb:
As I understand @linzhp suggested moving interface definition outside of the internal package all together, which would avoid duplication. Then internal package can use definitions from the public package just fine.

mfateev:
It cannot use them just fine as circular dependencies are prohibited in Go even with internal packages. This is the reason we have to resort to the ugly type aliases.

vitarb:
We could move them into a separate package that doesn't depend on internal. This would probably be a little less ugly than having to deal with this problem in a first place, but I see your point 😕
@mfateev, do you prefer doing it the way @abhinav suggested?

mfateev:
Then we end up with the double of packages as we want the same pattern everywhere:
activityinterfaces
activity
clientinterfaces
client
encodedinterfaces
encoded
workerinterfaces
worker
workflowinterfaces
workflow

and good luck finding out which one to use when writing code.

mfateev:
The default Go pattern is to use concrete structure in the implementation and the interface in the public one. Something like:

```
package internal

type FooImpl struct {
}
func NewFoo() *FooImpl {
  return   &FooImpl{}
}

package public

type Foo interface {
   ...
}

func NewFoo() Foo {
    return internal.NewFoo()
}

```

But it doesn't work if there is more than one internal implementation of Foo. I'm not sure if it is going to work with encoded.Value.

The only other two options are aliasing or duplicating the interface.

mfateev:
Just to clarify. It works with multiple implementations of Foo. The problem is that when there are multiple implementations the internal package usually has to have some way to reference them in a single API and an internal interface that matches Foo is created.

abhinav:
> Is the major reason for this change for unblocking "mocking aliased type
> issues we ran into during build" ?

@vancexu That's what inspired me to look into this but I wouldn't say that
this is a fix/workaround for that.

The primary reason for this change is that aliasing interfaces is unidiomatic
Go (explained in #928). That it fixes the mocking issue is a welcome
side-effect. I understand that other places in the code use the same pattern,
but I would recommend fixing those as you get the chance too.

Yes, it will add an inconsistency to the code base but this isn't really a
stylistic issue. It's a behavior issue.

> I would prefer to not introduce duplication if possible because it makes it
> makes things more confusing and harder to maintain as they can run out of
> sync.

@vitarb The added complexity here is the cost of using the internal package to
avoid cycles. I would argue that the aim should be ease and simplicity on the
user-end (the public API) even if it means added complexity internally.

They cannot go out of sync specifically because of the compile-time checks
added below the interfaces.

    var _ Foo = (internal.Foo)(nil)
    var _ internal.Foo = (Foo)(nil)

If the code compiles, it is guaranteed to be in-sync.

> Are there any reasons why we don't what to do what Zhongpeng suggested
> here and simply move definitions out of the internal package?

You cannot move them out of the internal package because it would introduce
dependency cycles.

The issue is that two packages A and B need to consume types from each other,
which Go does not allow. The workaround is to move the code for both A and B
into the same internal package, and then re-export the different interfaces
from A and B. We've dealt with similar code organization issues in YARPC in
the past.

This change is proposing that the workaround (move code into internal) doesn't
affect the user-facing API negatively.

vitarb:
@abhinav what would cause compilation errors if a new method is added to the internal interface (but not to the external copy)?

abhinav:
> @abhinav what would cause compilation errors if a new method is added to the internal interface (but not to the external copy)?

The statement `var _ Foo = (internal.Foo)(nil)` will fail to compile if a
method was added to `internal.Foo` if `Foo` didn't have it. The inverse of
that statement `var _ internal.Foo = (Foo)(nil)` will fail to compile if a
method was added to the external `Foo` that wasn't present in the internal
version.

In the change I made, I added both versions for all affected interfaces. So if
any of those interfaces go out of sync in either direction, it will fail
compilation.

vitarb:
That's right, I just realized that we have conversions in both directions.

vitarb:
I was also curious which mocks did you try to create that are not part of the mocks package already?
Another approach here would be providing sufficient mocks as part of the client itself, removing the need to use mockgen on the consumer's end.

vancexu:
@abhinav @linzhp any other thoughts on Vitaly's comments above?

abhinav:
Hey, just to clarify: I'm not concerned with the mock issue--that just opened
up this line of investigation. I'm not blocked by this in any way.

I've explained the reasoning for why the current pattern is not ideal above
and in #928, and how the duplication is okay especially because it makes you
think about the public API vs blurring the line between internal and external.

As project owners, I'll defer to you whether you want to fix it.

vancexu:
Thanks again @abhinav , we will close it.

PR #930, opened 2020-01-16
Make session heartbeat timeout customizable

yycptt:


internal/session.go
```diff
@@ -60,10 +60,14 @@ type (
 	// ExecutionTimeout: required, no default
 	//     Specifies the maximum amount of time the session can run
 	// CreationTimeout: required, no default
-	//     Specfifies how long session creation can take before returning an error
+	//     Specifies how long session creation can take before returning an error
+	// HeartbeatTimeout: optional, default 20s
+	//     Specifies the heartbeat timeoutIf. If heartbeat is not received by server
```
yux0: timeoutIf -> timeout

yycptt: Done.

workflow/session.go
```diff
@@ -36,7 +36,10 @@ type (
 	// ExecutionTimeout: required, no default
 	//     Specifies the maximum amount of time the session can run
 	// CreationTimeout: required, no default
-	//     Specfifies how long session creation can take before returning an error
+	//     Specifies how long session creation can take before returning an error
+	// HeartbeatTimeout: optional, default 20s
+	//     Specifies the heartbeat timeoutIf. If heartbeat is not received by server
```
yux0: same

Issue #936, opened 2020-02-19
activity.GetInfo(ctx) does not populate all the fields(Eg: WorkflowType is nil)


RudraPurohit:
![Screen Shot 2020-02-18 at 11 26 13 PM](https://user-images.githubusercontent.com/17635180/74811672-d46a3380-52a6-11ea-86aa-aa9134a908fb.png)

PR #937, opened 2020-02-21
Add missing info population in the activity info. (#936)

luowenking1:


internal/internal_event_handlers.go
```diff
@@ -494,11 +494,12 @@ func (wc *workflowEnvironmentImpl) ExecuteLocalActivity(params executeLocalActiv
...
-		params:      &params,
-		callback:    callback,
-		retryPolicy: params.RetryPolicy,
-		attempt:     params.Attempt,
+		activityID:   activityID,
+		params:       &params,
+		callback:     callback,
+		retryPolicy:  params.RetryPolicy,
+		attempt:      params.Attempt,
+		workflowTask: &workflowTask{},
```
venkat1109: why is this needed ? localActivityTask shouldn't and doesn't need a non-nil workflowTask

internal/internal_task_pollers.go
```diff
@@ -464,15 +464,44 @@ func (lath *localActivityTaskHandler) executeLocalActivityTask(task *localActivi
...
+		scheduled = time.Unix(0, task.workflowTask.task.GetScheduledTimestamp())
+	}
+
+	if task.workflowTask != nil && task.workflowTask.task != nil {
+		started = time.Unix(0, task.workflowTask.task.GetStartedTimestamp())
+	}
+
+	if task.workflowTask != nil && task.workflowTask.task != nil {
+		taskToken = task.workflowTask.task.TaskToken
+	}
```
venkat1109: this doesn't make sense to me. For local activity task, these things don't really exist or mean anything

internal/internal_task_pollers.go
```diff
@@ -465,6 +465,9 @@ func (lath *localActivityTaskHandler) executeLocalActivityTask(task *localActivi
 	}
 
 	ctx := context.WithValue(rootCtx, activityEnvContextKey, &activityEnvironment{
+		workflowType:      &task.params.WorkflowInfo.WorkflowType,
```
venkat1109: just to be defensive, consider making a copy in a local var and then pass the addr of that local var here

internal/internal_workflow_testsuite.go
```diff
@@ -534,6 +534,38 @@ func (env *testWorkflowEnvironmentImpl) executeActivity(
 	}
 }
 
+func (env *testWorkflowEnvironmentImpl) executeLocalActivityParams(
```
venkat1109: what is the difference between this and executeLocalActivity ? Why do you need to return concrete type here ?Given that this is unit test, can you type case inside of your test inside of creating this additional method which does the same thing ?

luowenking1:
> Can you please add a test here for local activity :
> https://github.com/uber-go/cadence-client/blob/master/test/integration_test.go
> 
> Inside of your local activity function, assert that the context has the value for the workflowType

I have added the checking there but I did not create a new method. Instead I added the checking in an existence of a local activity testing.

Issue #857, opened 2019-10-08
Integrate With New Get History Raw API


wxing1292:
Integrate with the new get raw history API, since currently on the server side the history events are deserialized when loaded from DB and serialized again when dispatching to client.

Server side corresponding issue: https://github.com/uber/cadence/issues/2644

mkolodezny:
https://github.com/uber/cadence/pull/3463

mkolodezny:
https://github.com/uber-go/cadence-client/pull/1019

PR #938, opened 2020-02-26
Integrate With New Get History Raw API (#857)

luowenking1:


internal/common/persistence/serializer.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// CadenceDeserializationError is an error type for cadence deserialization
+	CadenceDeserializationError struct {
+		msg string
+	}
+
+	// UnknownEncodingTypeError is an error type for unknown or unsupported encoding type
+	UnknownEncodingTypeError struct {
+		encodingType s.EncodingType
+	}
```
yycptt: I think I may lack some context here. But is there any special reason for creating separate struct for those errors, instead of using `errors.New()` or `fmt.Errorf()`?

internal/common/util/util.go
```diff
@@ -59,3 +62,30 @@ func AwaitWaitGroup(wg *sync.WaitGroup, timeout time.Duration) bool {
 		return false
 	}
 }
+
+// DeSerializeBlobDataToHistoryEvents deserialize the blob data to history event data
+func DeSerializeBlobDataToHistoryEvents(
```
yycptt: De**s**erializeBlobDataToHistoryEvents?

internal/internal_task_pollers.go
```diff
@@ -746,22 +747,25 @@ func newGetHistoryPageFunc(
...
-		size := len(h.Events)
-		if size > 0 && atDecisionTaskCompletedEventID > 0 &&
-			h.Events[size-1].GetEventId() > atDecisionTaskCompletedEventID {
-			first := h.Events[0].GetEventId() // eventIds start from 1
-			h.Events = h.Events[:atDecisionTaskCompletedEventID-first+1]
-			if h.Events[len(h.Events)-1].GetEventType() != s.EventTypeDecisionTaskCompleted {
-				return nil, nil, fmt.Errorf("newGetHistoryPageFunc: atDecisionTaskCompletedEventID(%v) "+
-					"points to event that is not DecisionTaskCompleted", atDecisionTaskCompletedEventID)
+		h, err1 := util.DeSerializeBlobDataToHistoryEvents(resp.RawHistory, s.HistoryEventFilterTypeAllEvent)
+		if err1 == nil {
```
yycptt: if err1 != nil, the code will jump to L768 and return nil? Doesn't look right to me. 

In general, we follow the pattern where we first check if err is nil or not and then perform operations on the result.

internal/internal_workflow_client.go
```diff
@@ -480,44 +481,54 @@ func (wc *workflowClient) TerminateWorkflow(ctx context.Context, workflowID stri
...
 }
 
-// GetWorkflowHistory return a channel which contains the history events of a given workflow
-func (wc *workflowClient) GetWorkflowHistory(ctx context.Context, workflowID string, runID string,
-	isLongPoll bool, filterType s.HistoryEventFilterType) HistoryEventIterator {
-
+// PollWorkflowHistory performs long polling of the history blob data from server and deserialize to history event construct data
+// workflowID is required, other parameters are optional.
+// If runID is omit, it will terminate currently running workflow (if there is one) based on the workflowID.
+func (wc *workflowClient) PollWorkflowHistory(ctx context.Context, workflowID string, runID string,
```
yycptt: nit: `PollForWorkflowHistory` to be consistent with the API name?

internal/internal_workflow_client.go
```diff
@@ -480,44 +481,54 @@ func (wc *workflowClient) TerminateWorkflow(ctx context.Context, workflowID stri
...
 					defer cancel()
-					response, err1 = wc.workflowService.GetWorkflowExecutionHistory(tchCtx, request, opt...)
-					return err1
+					rawResponse, err := wc.workflowService.PollForWorkflowExecutionRawHistory(tchCtx, request, opt...)
+
+					if err != nil {
+						return err
+					}
+					historyEvents, err := util.DeSerializeBlobDataToHistoryEvents(rawResponse.RawHistory, filterType)
+
```
yycptt: lets first check if err != nil, same for the function below.

client/client.go
```diff
@@ -167,16 +167,29 @@ type (
...
+		//		iter := GetWorkflowHistory(ctx, workflowID, runID)
+		//		events := []*shared.HistoryEvent{}
+		//		for iter.HasNext() {
+		//			event, err := iter.Next()
+		//			if err != nil {
+		//				return err
+		//			}
+		//			events = append(events, event)
+		//		}
+		GetWorkflowHistory(ctx context.Context, workflowID string, runID string) HistoryEventIterator
```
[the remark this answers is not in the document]
luowenking1: I have revert back this interface and hide the implementation of GetRawHistoryData and PollForHistoryData behind calling of this API.

client/client.go
```diff
@@ -185,7 +198,7 @@ type (
 		//			}
 		//			events = append(events, event)
 		//		}
-		GetWorkflowHistory(ctx context.Context, workflowID string, runID string, isLongPoll bool, filterType s.HistoryEventFilterType) HistoryEventIterator
+		PollWorkflowHistory(ctx context.Context, workflowID string, runID string, filterType s.HistoryEventFilterType) HistoryEventIterator
```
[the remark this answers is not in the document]
luowenking1: I have hided this API.

internal/internal_workflow_client.go
```diff
@@ -483,41 +484,62 @@ func (wc *workflowClient) TerminateWorkflow(ctx context.Context, workflowID stri
...
 	Loop:
 		for {
 			err = backoff.Retry(ctx,
 				func() error {
-					var err1 error
 					tchCtx, cancel, opt := newChannelContext(ctx, func(builder *contextBuilder) {
-						if isLongPoll {
-							builder.Timeout = defaultGetHistoryTimeoutInSecs * time.Second
-						}
+						builder.Timeout = defaultGetHistoryTimeoutInSecs * time.Second
```
venkat1109: can you change the defaultGetHistoryTimeoutInSecs to be a duration instead and rename it to simply defaultGetHistoryTimeout

internal/internal_workflow_client.go
```diff
@@ -483,41 +484,62 @@ func (wc *workflowClient) TerminateWorkflow(ctx context.Context, workflowID stri
...
 			HistoryEventFilterType: &filterType,
 			NextPageToken:          nexttoken,
 		}
-
-		var response *s.GetWorkflowExecutionHistoryResponse
-		var err error
 	Loop:
 		for {
 			err = backoff.Retry(ctx,
 				func() error {
```
venkat1109: this func() is simply too long for it to be an anonymous inline func. Can you create a separate helper method and pass that method to backoff.Retry() ?

internal/internal_workflow_client.go
```diff
@@ -483,41 +484,62 @@ func (wc *workflowClient) TerminateWorkflow(ctx context.Context, workflowID stri
...
-						}
+						builder.Timeout = defaultGetHistoryTimeoutInSecs * time.Second
 					})
 					defer cancel()
-					response, err1 = wc.workflowService.GetWorkflowExecutionHistory(tchCtx, request, opt...)
-					return err1
+					rawResponse, err := wc.workflowService.PollForWorkflowExecutionRawHistory(tchCtx, request, opt...)
+
+					if err != nil {
+						return err
```
venkat1109: do we log or emit separate metrics for these deserliazation errors in the caller ? because these are typically bugs or data corruption. If not, lets add it.

venkat1109:
Also, can you update the integration tests to run once with / without this feature turned on

internal/common/metrics/service_wrapper.go
```diff
@@ -46,44 +46,46 @@ type (
...
+	scopeNameListArchivedWorkflowExecutions     = CadenceMetricsPrefix + "ListArchviedExecutions"
+	scopeNameScanWorkflowExecutions             = CadenceMetricsPrefix + "ScanWorkflowExecutions"
+	scopeNameCountWorkflowExecutions            = CadenceMetricsPrefix + "CountWorkflowExecutions"
+	scopeNamePollForActivityTask                = CadenceMetricsPrefix + "PollForActivityTask"
+	scopeNamePollForDecisionTask                = CadenceMetricsPrefix + "PollForDecisionTask"
+	scopeNameRecordActivityTaskHeartbeat        = CadenceMetricsPrefix + "RecordActivityTaskHeartbeat"
+	scopeNameRecordActivityTaskHeartbeatByID    = CadenceMetricsPrefix + "RecordActivityTaskHeartbeatByID"
+	scopeNameRegisterDomain                     = CadenceMetricsPrefix + "RegisterDomain"
+	scopeNameRequestCancelWorkflowExecution     = CadenceMetricsPrefix + "RequestCancelWorkflowExecution"
+	scopeNameRespondActivityTaskCanceled        = CadenceMetricsPrefix + "RespondActivityTaskCanceled"
```
venkat1109: except for this newline, looks like nothing changed in this file ?can you revert it just to make the diff cleaner

internal/internal_task_pollers.go
```diff
@@ -749,7 +750,19 @@ func newGetHistoryPageFunc(
...
+
+		if resp.RawHistory != nil {
+			var err1 error
+			h, err1 = serializer.DeserializeBlobDataToHistoryEvents(resp.RawHistory, s.HistoryEventFilterTypeAllEvent)
+			if err1 != nil {
+				return nil, nil, nil
+			}
+		} else {
+			h = resp.History
+		}
```
venkat1109: this same logic may be needed in more than one place.. can you create a helper method that takes response and returns deserialized historyEvents and error ?

PR #939, opened 2020-02-26
Saga implementation

arunpoudel:
Picking up from #870 

I understand that there are going to be more boilerplate to write when working with this package, but my argument is that allowing interface{} to be passed as action and compensation is non-idiomatic go, and we should diverse from it as much as possible.


arunpoudel:
No problem, thanks for the response

Issue #134, opened 2017-05-30
Automatic heartbeating for activities


samarabbas:

ibarrajo:
Automatic heart beating was implemented in v0.17.0 for the golang cadence-client.

https://github.com/uber-go/cadence-client/releases/tag/v0.17.0

PR #941, opened 2020-02-26
Clarify auto heartbeating support

sevein:
Clarify that `AutoHeartBeat` is a no-op since auto heartbeating has not been implemented yet.

Relates to https://github.com/uber-go/cadence-client/issues/134.

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -114,11 +114,6 @@ type (
 		// Default value is 2
 		MaxConcurrentDecisionTaskPollers int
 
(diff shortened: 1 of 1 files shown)


sevein:
Ok done, sorry for the late response. It introducing a backward-incompatible change but I guess you're assuming that's a way to communicate users that `AutoHeartBeat` isn't really a thing?

mfateev:
Yes, I think it is better to break the build than assume that it is working.

PR #951, opened 2020-03-28
use the latest thrift

bhcleek:
Update github.com/apache/thrift to a version that's compatible with
other packages that are commonly vendored into digitalocean packages.


Sunkwan-Kwon:
@bhcleek Hello. I got the same issue with you, so I also have forked cadence-client for modifying that.
I think this PR should be merged into the master branch of this original repository. Could you reopen this PR? Or did you have any reason that closed this PR?

PR #952, opened 2020-04-02
Retry getHistory request from passive cluster in case of replication lag

meiliang86:


vancexu:
It's been sometime, is this ready for review?

meiliang86:
@vancexu ready for review

PR #955, opened 2020-04-08
Update cadence-idl to use task list for visibility data

mkolodezny:
follow up for https://github.com/uber/cadence-idl/pull/24


go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/stretchr/testify v1.3.0
 	github.com/uber-go/mapdecode v1.0.0 // indirect
 	github.com/uber-go/tally v3.3.1+incompatible
+	github.com/uber/cadence-idl v0.0.0-20200408153039-b07c6291fa3a // indirect
```
yycptt, line 37: Please update idl submodule. There should be no changes in the go.mod and go.sum file as `cadence-idl` is not imported anywhere. I believe if you run `go mod tidy`, then the changes here will be reverted.

PR #957, opened 2020-04-10
Add comment on GetWorkflow

vancexu:
and fix typos


mfateev:
Why would it timeout if workflow is not found? It should fail fast in this case.

client/client.go
```diff
@@ -118,6 +118,7 @@ type (
 		//  - Get(ctx context.Context, valuePtr interface{}) error: which will fill the workflow
 		//    execution result to valuePtr, if workflow execution is a success, or return corresponding
 		//    error. This is a blocking API.
+		// If workflow not found, the Get() will timeout.
```
yux0, line 121: More generic, the service side API doesn't know if the workflow is found. Basically, it will wait for the workflow to close and return the closed event. So this API will timeout if the workflow does not closed in time.

vancexu:
> Why would it timeout if workflow is not found? It should fail fast in this case.

I agree it should fail fast in this case. 
Dig into the then found and fixed an issue on server side that caused this timeout. 
https://github.com/uber/cadence/pull/3188

Now it immediately return EntityNotFound.

PR #963, opened 2020-04-27
doc: explain how to abandon child workflows

seriousben:
When using the `ParentClosePolicyAbandon` policy for a child workflow, it is important to wait for the Child Workflow to have started to ensure its execution.

We've been hit by this timing issue especially for child workflows started as the last operation of the parent workflow.

Hopefully this documentation change can help others.

### DIFF
--- a/workflow/doc.go
+++ b/workflow/doc.go
@@ -224,10 +224,10 @@ for an activity it invoked.
 		WorkflowID:                   "BID-SIMPLE-CHILD-WORKFLOW",
 		ExecutionStartToCloseTimeout: time.Minute * 30,
 	}
-	ctx = workflow.WithChildOptions(ctx, cwo)
+	childCtx = workflow.WithChildOptions(ctx, cwo)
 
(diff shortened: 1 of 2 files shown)


workflow/doc.go
```diff
@@ -262,6 +262,27 @@ The workflow.ExecuteChildWorkflow() function is very similar to the workflow.Exe
...
+
+        // Do not terminate when parent closes.
+		ParentClosePolicy: client.ParentClosePolicyAbandon,
+	}
+	childCtx = workflow.WithChildOptions(ctx, cwo)
+
+	future := workflow.ExecuteChildWorkflow(childCtx, SimpleChildWorkflow, value)
+
+	// Wait for the child workflow to start
+	if err := f.GetChildWorkflowExecution().Get(ctx, nil); err != nil {
```
vancexu, line 281: future instead of f

seriousben: Thanks, just fixed it.

PR #965, opened 2020-04-27
Improve session framework heartbeat mechanism

yycptt:
Currently go client performs heartbeat batching to reduce load on the server side. However, it's very likely to cause the entire session to fail if the batched heartbeat failed due to some transient error. 

Consider the example where heartbeat timeout is 12s when the session activity heartbeat every 4s. At 4th second, the first heartbeat will be sent and a batch window will be created. The second(at 8th second), and the third (at the 12th second) heartbeat will be batched as the batch window is 0.8*12=9.6s. At 13.6s, this batched heartbeat will be sent. If this one failed due to transient errors like timeout, then the session will heartbeat timeout as the next heartbeat will be sent on the 16th second, while the last successful heartbeat is at 4th second. In addition, we can't perform retry on the batched heartbeat as the error is dropped.

So this pr does two things:
1. Add a `skipBatching` parameter to the `Heartbeat` method of the `ServiceInvoker` interface, so that session can skip heartbeat batching and control the heartbeat rate. Since this API is not available to customer, we don't need to worry about customer code skip heartbeat batching. Also the session framework has complete control over the heartbeat rate and can make sure the heartbeat rate is very low (< 1rps).
2. Retry heartbeat on transient errors like timeout. The retry policy is not aggressive, the initial backoff interval is 1 sec and max interval is 2 sec.

Test locally with modified (added 30s latency to all activities) fileprocessing sample and the workflow can successfully complete even with 50% injected error rate on heartbeating.


internal/internal_task_handlers.go
```diff
@@ -1670,7 +1670,7 @@ func (i *cadenceInvoker) Heartbeat(details []byte) error {
 			i.Unlock()
 
 			if detailsToReport != nil {
-				i.Heartbeat(*detailsToReport)
+				i.Heartbeat(*detailsToReport, false)
```
yux0, line 1673: As you mentioned offline, put a TODO here to revisit the race condition issue

yycptt: Done.

PR #969, opened 2020-05-07
Support returning WorkflowExecutionAlreadyStartedError in tests

seriousben:
Without this change, it is not possible to return a `WorkflowExecutionAlreadyStartedError` in tests.

This makes the following possible:
```
s.env.OnWorkflow(Workflow, mock.Anything, mock.Anything).
		Once().
		Return(&shared.WorkflowExecutionAlreadyStartedError{
			RunId: &runID,
		})
```


seriousben:
Just added a test.

PR #972, opened 2020-05-13
Fix inconsistent Jaeger baggage injection/extraction

quessliu:
**Context**

We've observed an issue where

1) Workflow gets started with Jaeger Baggage that we base64 encode. Ex: 
`eyJzZXJ2aWNlcyI6W3sicG9vbCI6ImJpdHMtdWNpIiwic2VydmljZV9uYW1lIjoiZnVsZmlsbG1lbnQiLCJob3N0IjoiYWdlbnQzNC1waHgyLnByb2QudWJlci5pbnRlcm5hbCIsInBvcnRzIjp7Imh0dHAiOjMxMDI1LCJodHRwMiI6MzEwMTgsInRjaGFubmVsIjozMTAzMX19XX0=`

2) Subsequent activity gets scheduled with Jaeger Baggage that is url-escaped (`opentracing.HTTPHeaders` does this https://github.com/jaegertracing/jaeger-client-go/blob/ff64c7c174e5f532e57276fcfafef2d7114d26ab/propagation.go#L82)
Ex: 
`eyJzZXJ2aWNlcyI6W3sicG9vbCI6ImJpdHMtdWNpIiwic2VydmljZV9uYW1lIjoiZnVsZmlsbG1lbnQiLCJob3N0IjoiYWdlbnQzNC1waHgyLnByb2QudWJlci5pbnRlcm5hbCIsInBvcnRzIjp7Imh0dHAiOjMxMDI1LCJodHRwMiI6MzEwMTgsInRjaGFubmVsIjozMTAzMX19XX0%3D`
The inconsistent injection and extraction leads to issues where the baggage gets corrupted due to the mismatched encoding.

**Testing Plan**

Added a unit test that fails without the change and succeeds with change to verify this bug and fix.

cc @yurishkuro

### DIFF
--- a/internal/tracer.go
+++ b/internal/tracer.go
@@ -100,7 +100,7 @@ func (t *tracingContextPropagator) InjectFromWorkflow(
 	if spanContext == nil {
 		return nil
 	}
-	return t.tracer.Inject(spanContext, opentracing.HTTPHeaders, tracingWriter{hw})
+	return t.tracer.Inject(spanContext, opentracing.TextMap, tracingWriter{hw})
 }
 
 func (t *tracingContextPropagator) ExtractToWorkflow(
--- a/internal/tracer_test.go
+++ b/internal/tracer_test.go
@@ -116,3 +116,34 @@ func TestTracingContextPropagatorWorkflowContextNoSpan(t *testing.T) {
 	returnCtx, err = ctxProp.ExtractToWorkflow(returnCtx, NewHeaderReader(header))
 	assert.NoError(t, err)
 }
+
+func TestConsistentInjectionExtraction(t *testing.T) {
+	t.Parallel()
+	tracer, closer, err := jaeger_config.Configuration{ServiceName: "test-service"}.NewTracer()
+	require.NoError(t, err)
+	defer closer.Close()
+	ctxProp := NewTracingContextPropagator(zap.NewNop(), tracer)
+
+	span := tracer.StartSpan("test-operation")
+	// base64 encoded string '{}'
+	var baggageVal = "e30="
+	span.SetBaggageItem("request-tenancy", baggageVal)
+	assert.NotNil(t, span.Context())
+	ctx := contextWithSpan(Background(), span.Context())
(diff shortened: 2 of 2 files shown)


internal/tracer.go
```diff
@@ -100,7 +100,7 @@ func (t *tracingContextPropagator) InjectFromWorkflow(
 	if spanContext == nil {
 		return nil
 	}
-	return t.tracer.Inject(spanContext, opentracing.HTTPHeaders, tracingWriter{hw})
+	return t.tracer.Inject(spanContext, opentracing.TextMap, tracingWriter{hw})
```
yurishkuro, line 103: Actually, both may be correct. It's not clear from the context of this function which medium the context is being encoded into. If it's HTTP request, then the old code was correct. If it's some other medium, like storing the serialized context in the database, then both formats are fine, as long as the same format is used when reading and decoding the stored context.

quessliu: Yeah, good point. I think Cadence is using Tchannel/Thrift for its RPCs and this also gets persisted in DB. Like you mentioned, the problem here is inconsistent encoding protocols for reading and writing the context.

meiliang86: So does it mean that HTTPHeaders will work for both cases? Then why do we need this change?

quessliu: HTTPHeaders could work for Cadence (though we would then need to change line 110 to make sure we decode using `opentracing.HTTPHeaders` so that we properly URL-decode.

From the opentracing docs:

```
// HTTPHeaders represents SpanContexts as HTTP header string pairs.
	//
	// Unlike TextMap, the HTTPHeaders format requires that the keys and values
	// be valid as HTTP headers as-is (i.e., character casing may be unstable
	// and special characters are disallowed in keys, values should be
	// URL-escaped, etc)
```

I don't think it matters in this case, but using the HTTPHeaders encoding seems to be slightly stricter and may lose data like case-sensitivity while encoding.

meiliang86: Got it. Thanks!

PR #974, opened 2020-05-17
Expose WithRetryPolicy in workflow package

sagikazarmark:
Signed-off-by: Mark Sagi-Kazar <mark.sagikazar@gmail.com>

### DIFF
--- a/workflow/activity_options.go
+++ b/workflow/activity_options.go
@@ -32,6 +32,9 @@ type ActivityOptions = internal.ActivityOptions
 // LocalActivityOptions doc
 type LocalActivityOptions = internal.LocalActivityOptions
 
+// RetryPolicy specify how to retry activity if error happens.
+type RetryPolicy = internal.RetryPolicy
+
 // WithActivityOptions adds all options to the copy of the context.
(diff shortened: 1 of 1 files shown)


workflow/activity_options.go
```diff
@@ -71,3 +74,8 @@ func WithHeartbeatTimeout(ctx Context, d time.Duration) Context {
 func WithWaitForCancellation(ctx Context, wait bool) Context {
 	return internal.WithWaitForCancellation(ctx, wait)
 }
+
+// WithRetryPolicy adds wait for the cacellation to the copy of the context.
```
sevein, line 78:
```suggestion
// WithRetryPolicy adds wait for the cancellation to the copy of the context.
```

sagikazarmark: You fixed a typo in a copy paste error. 😆

mfateev:
RetryPolicy is already exposed: https://github.com/uber-go/cadence-client/blob/master/cadence.go

sagikazarmark:
The type itself is, but that's a different package and it would lead to an import cycle.

This PR is about the `WithRetryPolicy` context mutator.

sagikazarmark:
Thanks!

PR #977, opened 2020-05-29
Fix aggressive poller for non-retriable error

vancexu:
Currently poller will retry immediately with error response such as BadRequestError. 
This doesn't make sense, because all the retry will get same error response until correct poller is deployed.
As a result, server frontend be DDOS by such poller.  

An example to repro is: update badbinary for a domain.

This PR fix the issue. 
If err is non-retriable, just hang. 
With any other kinds of error, do exponential retry


internal/internal_worker_base.go
```diff
@@ -260,7 +260,11 @@ func (bw *baseWorker) pollTask() {
 		if err != nil && enableVerboseLogging {
 			bw.logger.Debug("Failed to poll for task.", zap.Error(err))
 		}
-		if err != nil && isServiceTransientError(err) {
+		if err != nil {
+			if isNonRetriableError(err) {
+				bw.logger.Error("Worker stop polling task.", zap.Error(err))
+				return
```
yycptt, line 266: Is the intention here to stop all poller goroutines in the decision/activity worker? If so, is it possible to trigger a shutdown directly here (by calling `Stop()` maybe), instead of stopping the pollers one by one? Personally, I think that will be a little bit easier to understand. It will also stop all the poller goroutines instead of making them block on the `select` which will only be unblocked during worker shutdown.

internal/internal_retry.go
```diff
@@ -70,7 +70,8 @@ func isServiceTransientError(err error) bool {
 		*s.DomainAlreadyExistsError,
 		*s.QueryFailedError,
 		*s.DomainNotActiveError,
-		*s.CancellationAlreadyRequestedError:
+		*s.CancellationAlreadyRequestedError,
+		*s.ClientVersionNotSupportedError:
```
meiliang86, line 74: Why is this transient?

internal/internal_worker_base.go
```diff
@@ -277,6 +281,18 @@ func (bw *baseWorker) pollTask() {
 	}
 }
 
+func isNonRetriableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err.(type) {
+	case *shared.BadRequestError,
+		*shared.ClientVersionNotSupportedError:
```
meiliang86, line 290: ClientVersionNotSupportedError will not happen in prod. What is the BadRequestError that you want to protect?

internal/internal_retry.go
vancexu: it is not transient, so return false

internal/internal_worker_base.go
vancexu:
```
&gen.BadRequestError{
				Message: fmt.Sprintf("binary %v already marked as bad deployment", binaryChecksum),
			}
```
This is the one help me find the bug. When set badbinary for a domain, all workers with that badbinary should stop working, but without this fix it end up in dead loop polling same error.

I checked and believe all other BadRequest that poller can get from `PollForDecisionTask` call should also be protected. Because currently once BadRequest happens, it means something like tasklist or domain not correct/exist.

yux0: Is this a bug on the auto reset? The auto reset happens when a workflow(marked with a bad binary checksum) try to make progress. But the BadRequestError returns from the frontend and workflow will never make progress. 

And if we do this, how can it be auto reset?

internal/internal_worker_base.go
vancexu: Make sense as discussed in team. Updated

internal/internal_worker_base.go
vancexu: Not a bug on auto reset. As discussed, when different binary (either older or newer) worker out, workflow will continue

internal/internal_utils.go
```diff
@@ -275,7 +275,7 @@ func awaitWaitGroup(wg *sync.WaitGroup, timeout time.Duration) bool {
 
 func getKillSignal() <-chan os.Signal {
 	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
```
yycptt, line 278: I may lack some context here. But I think it's better to use `os` package if we can as it's system independent?  Also, from this page (https://golang.org/pkg/syscall/) seems like `syscall` package has been locked down.

vancexu: They are same thing so just make it unified, in os.
```
Interrupt Signal = syscall.SIGINT
```

internal/internal_worker_base.go
meiliang86: See if you can add a unit test case to test this logic (i.e. mock return bad request and verify that worker got shutdown).

Issue #872, opened 2019-11-03
Context Propagators with LocalActivity


talgendler:
If I execute an **Activity** all my context propagators are called to extract/inject headers, however **LocalActivity** doesn't support context propagators. Are there any plans to support this ?

assafsch:
I believe this issue can be closed now.

PR #986, opened 2020-06-08
Add Context Propagation support for Local Activities

assafsch:
**What changed?**
Context Propagation is now supported for Local Activities

**Why?**
Context Propagation wasn't working for Local Activities

**How did you test it?**
Added an integration test. Couldn't find any good place to add a unit-test to.

**Potential risks**
Can't say

NOTE: I used the `StringMapPropagator` from the `internal` package. Sadly, I couldn't use it directly and had to clone into the `test` package.

### DIFF
--- a/test/activity_test.go
+++ b/test/activity_test.go
@@ -90,7 +90,7 @@ func (a *Activities) fail(_ context.Context) error {
 func (a *Activities) DuplicateStringInContext(ctx context.Context) (string, error) {
 	originalString := ctx.Value(contextKey(testContextKey))
 	if originalString == nil {
-		return "", fmt.Errorf("context did not propagate to local activity")
+		return "", fmt.Errorf("context did not propagate to activity")
 	}
 	return strings.Repeat(originalString.(string), 2), nil
 }
--- a/test/activity_test.go
+++ b/test/activity_test.go
@@ -87,19 +87,12 @@ func (a *Activities) fail(_ context.Context) error {
 	return errFailOnPurpose
 }
 
-func (a *Activities) InspectActivityInfo(ctx context.Context, domain, taskList, wfType string) error {
-	a.append("inspectActivityInfo")
-	info := activity.GetInfo(ctx)
-	if info.WorkflowDomain != domain {
-		return fmt.Errorf("expected domainName %v but got %v", domain, info.WorkflowDomain)
+func (a *Activities) DuplicateStringInContext(ctx context.Context) (string, error) {
+	originalString := ctx.Value(contextKey(testContextKey))
+	if originalString == nil {
+		return "", fmt.Errorf("context did not propagate to local activity")
 	}
-	if info.WorkflowType == nil || info.WorkflowType.Name != wfType {
-		return fmt.Errorf("expected workflowType %v but got %v", wfType, info.WorkflowType)
-	}
-	if info.TaskList != taskList {
-		return fmt.Errorf("expected taskList %v but got %v", taskList, info.TaskList)
-	}
-	return nil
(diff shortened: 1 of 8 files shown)


assafsch:
The following issue should be linked to this PR:
https://github.com/uber-go/cadence-client/issues/872

test/test_utils.go
```diff
@@ -107,3 +114,70 @@ func newRPCClient(
 	client := workflowserviceclient.New(dispatcher.ClientConfig(serviceName))
 	return &rpcClient{Interface: client, dispatcher: dispatcher}, nil
 }
+
+// stringMapPropagator propagates the list of keys across a workflow,
+// interpreting the payloads as strings.
+// BORROWED FROM 'internal' PACKAGE TESTS.
```
meiliang86, line 120: Help remove the class from internal package and change all the reference to this one?

test/activity_test.go
```diff
@@ -87,19 +87,12 @@ func (a *Activities) fail(_ context.Context) error {
...
-	}
-	if info.WorkflowType == nil || info.WorkflowType.Name != wfType {
-		return fmt.Errorf("expected workflowType %v but got %v", wfType, info.WorkflowType)
-	}
-	if info.TaskList != taskList {
-		return fmt.Errorf("expected taskList %v but got %v", taskList, info.TaskList)
+func (a *Activities) DuplicateStringInContext(ctx context.Context) (string, error) {
+	originalString := ctx.Value(contextKey(testContextKey))
+	if originalString == nil {
+		return "", fmt.Errorf("context did not propagate to local activity")
```
meiliang86, line 93: nit: remove "local" in error message, as the activity implementation itself is agnostic of whether it is local or not.

test/test_utils.go
```diff
@@ -22,22 +22,29 @@
...
-type Config struct {
-	ServiceAddr string
-	ServiceName string
-	IsStickyOff bool
-	Debug       bool
-}
+	"go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
+	"go.uber.org/cadence/workflow"
+)
+type (
```
meiliang86, line 36: nit: add new line above type def.

assafsch:
hi @meiliang86 - do you want me to fix according to you comments before merging?
I'm not familiar with "nit"

test/activity_test.go
assafsch: got it

test/test_utils.go
assafsch: got it

test/test_utils.go
assafsch: sure

assafsch: thinking about it some more....is it a good approach to have code in an implementation package (internal) reference and rely on a struct defined in the test package?

also, I'm thinking this can introduce an import cycle.
test  > workflow > internal > test. How about making stringMapPropagator public? or having a "common" package used by both the internal and test packages?

meiliang86: My understanding is that the reference in internal package is also for testing purpose, right?
It does not seem that there's a common package outside of internal. So I am ok leaving this as is for now.

meiliang86:
> hi @meiliang86 - do you want me to fix according to you comments before merging?
> I'm not familiar with "nit"

Yeah just fix the formatting issue and I'll merge the PR when tests pass.

test/test_utils.go
assafsch: Ok, so I've fixed your other 2 comments (newline and error message).
Merge at will :)

Thanks, very happy to contribute!

PR #992, opened 2020-06-22
Fixes for struct activity naming

vytautas-karpavicius:
**What changed?**
These changes allow executing activities within structs by directly passing function pointer instead of manually constructing function name.

For example, given the following registration:
```
func register(w worker.Worker) {
  w.RegisterActivityWithOptions(&myActivity{}, activity.RegisterOptions{Name: "some-prefix"})
}

type myActivity struct {}
func (m *myActivity) Run() error {}
```

One can execute it with:
```
var my *myActivity
future := workflow.ExecuteActivity(ctx, my.Run)
```

Instead of manually constructing string name like:
```
future := workflow.ExecuteActivity(ctx, "some-prefix_Run")
```

**Why?**
Manual activity name string construction is not straightforward, involves internal logic how this string is formatted by cadence client and is a subject to get outdated if activity method get renamed.

**How did you test it?**
Unit tests, integration tests.

**Potential risks**

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -57,6 +57,9 @@ type (
 
 	// RegisterActivityOptions consists of options for registering an activity
 	RegisterActivityOptions struct {
+		// When an activity is a function the name is an actual activity type name.
+		// When an activity is part of a structure then each member of the structure becomes an activity with
+		// this Name as a prefix + activity function name.
 		Name                          string
 		DisableAlreadyRegisteredCheck bool
 	}
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -1087,10 +1087,10 @@ func isError(inType reflect.Type) bool {
 }
 
 func getFunctionName(i interface{}) string {
-	fullName, ok := i.(string)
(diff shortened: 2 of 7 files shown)


vytautas-karpavicius:
> LGTM. Any idea how conflict is handled if two activity types happen to contain the same method name?
> Might worth documenting this behavior.

The behavior would be the same as any other (function based) activity would get registered with the same name. Unless `DisableAlreadyRegisteredCheck` is set, it will panic with error indicating duplicate registration. Otherwise behavior is unpredictable, as previously registered activity will get overwritten.

meiliang86:
> > LGTM. Any idea how conflict is handled if two activity types happen to contain the same method name?
> > Might worth documenting this behavior.
> 
> The behavior would be the same as any other (function based) activity would get registered with the same name. Unless `DisableAlreadyRegisteredCheck` is set, it will panic with error indicating duplicate registration. Otherwise behavior is unpredictable, as previously registered activity will get overwritten.

Got it. Thanks!

PR #994, opened 2020-06-29
Backwards compatible activity type names

vytautas-karpavicius:
**What changed?**
This change reverts shortening when no activity alias (prefix) is provided.
The use of prefix shortens the name, assuming the users responsibility to
choose proper prefix. Additionally added an option to enable shorted names.

**Why?**
Previously merged shortened activity type names of form {Method} is not
compatible with current users of this client, since multiple activities
in different packages with the same name results in naming conflicts.

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/registry_test.go
+++ b/internal/registry_test.go
@@ -1,0 +1,154 @@
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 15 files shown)


internal/registry_test.go
```diff
@@ -0,0 +1,154 @@
...
+// THE SOFTWARE.
+
+package internal
+
+import (
+	"testing"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWorkflowRegistration(t *testing.T) {
```
meiliang86, line 28: Is EnableShortName only available for activity registration? why is that?

vytautas-karpavicius: For now - yes. But this is a good point, I will add it for workflows as well.

PR #998, opened 2020-07-11
Fixed Windows compilation failure

h7kanna:
**What changed?**
Replaced syscall usage

**Why?**
Compilation failure of the cadence worker on Windows
vendor\go.uber.org\cadence\internal\internal_worker_base.go:268:5: undefined syscall.Kill

**How did you test it?**
Successfully built on Windows machine

**Potential risks**

None

### DIFF
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -27,6 +27,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


davidroman0O:
Maybe I'm commenting at the wrong package or issue but i still got that:

```bash
 go get go.uber.org/cadence
# go.uber.org/cadence/internal
C:\Users\me\go\pkg\mod\go.uber.org\cadence@v0.13.3\internal\internal_worker_base.go:268:5: undefined: syscall.Kill
```

I've added replace from another issue:

```
replace github.com/apache/thrift => github.com/apache/thrift v0.0.0-20190309152529-a9b748bb0e02
```

So I don't have the other error, now I'm stuck with what this MR is supposed to fix, I may did something wrong, any idea?

allaryin:
The problem here seems to be twofold:
1. This MR is only in master, it hasn't made its way into a tagged release yet. Cf. https://github.com/uber-go/cadence-client/blob/v0.13.4/internal/internal_worker_base.go#L268. (I'm unsure how they cherrypicked the tags to avoid this commit, something weird in their promotion process, I guess)
2. They're sending `SIGINT`, which the docs explicitly state is unsupported on windows - https://golang.org/pkg/os/#Process.Signal. The correct signal would be `SIGTERM` - but that isn't necessary to compile, just to function as expected ;)

Issue #943, opened 2020-02-28
Error "unable to decode argument" if return nil in local activity


meiliang86:
Not possible to make LocalActivity that returns nil in result. Workflow fails with

```
  "reason": "cadenceInternal:Generic",
  "details": "unable to decode argument: 0, **struct {}, with json error: EOF",
```

I've made a sample to demonstrate this problem:
https://github.com/olegmedvedev/cadence-local-activity-sample

zhaytee:
Hi, we just encountered this issue. We use local activities for fetching db objects, and those activities return nil if the object is not found (i.e. we don't consider that to be an error). So as you can imagine, this is a fairly severe issue for our Cadence workflows.

Using a custom DataConverter reveals that the ToData() func is never called in this case, only FromData(), which results in FromData() trying to decode zero data. Hence the EOF error.

Any timeline on a fix? Currently using Go client 0.12.2

Thank you!

PR #1011, opened 2020-07-22
Fix decode issue for local activities returning nil result (#943)

vytautas-karpavicius:
**What changed?**
Fix decode issue for local activities returning nil result (#943)

**Why?**
When local activity returns nil result, it is encoded into nil byte
slice ([]byte). Later this is converted into string field ResultJSON
within localActivityMarkerData structure. Such casting results in
empty string, as this is its zero value in golang.

Reverse conversion is not simetrical. Casting empty string back to
byte slice results in empty byte slice. This later causes EOF error
when trying to decode it back to the original result.

The fix is to check whether it is non-empty string and only then
assign the result back. This will be simetrical and will not cause
decoding error later.

**How did you test it?**
This change also adds additional replay tests, that covers this
scenario among many other local activity related cases.

**Potential risks**


vancexu:
Another related task when you have time, could you also check if there are similar edge cases in other marker events (SideEffect/MutableSideEffect/GetVersion)?

meiliang86:
Fixes #943

Issue #1015, opened 2020-08-05
Add rpc request metrics with retry in consideration


meiliang86:
**Is your feature request related to a problem? Please describe.**
Right now we have rpc metrics for single requests that does not factor in retry options. Since client does retries automatically, this does not tell the accurate story about the actual failure rate.

**Proposed Solution**
Add metrics after all the retry attempts are exhausted. We can start with sync APIs (start/signal/signalWithStart), and maybe implement this for all if that's not messing up the code too much.

anish531213:
@meiliang86 I didn't understand the purposed solution. As we have backoff retry, these retries should still be counted in the failure rate. Do you mean that we only count this as one failure if there is context timeout?

PR #1018, opened 2020-08-19
add rpc request metrics with retry in consideration

anish531213:
**What changed?**
(#1015) Introduced new metrics for sync APIs that factors in retries as well.

**Why?**
Right now we have rpc metrics for single requests that does not factor in retry options. Since client does retries automatically, this does not tell the accurate story about the actual failure rate.

**How did you test it?**
Not verified yet.

**Potential risks**
No risk.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -521,6 +521,10 @@ func NewClient(service workflowserviceclient.Interface, domain string, options *
 		domain:             domain,
 		registry:           newRegistry(),
 		metricsScope:       metrics.NewTaggedScope(metricScope),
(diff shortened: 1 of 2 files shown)


vancexu:
I am still confused on what we try to achieve here. 
Currently, client calls to APIs such as StartWF/SignalWithStart will not count RPC layer retry at all. 
Say I call client.StartWorkflow once and it is success, no matter how many server call it actually did, client will only emit 1 request for `cadence-workflow-start`; it it fail, it won't emit `cadence-workflow-start` metric.

What goal are we trying to achieve here? @meiliang86

PR #1025, opened 2020-09-29
Fix CONTRIBUTING.md for dependency

longquanzheng:
**What changed?**
README

**Why?**
It's required by the script https://github.com/uber-go/cadence-client/blob/master/versioned_go_build.sh
Without glide the build will not succeed. 
I am not sure why we need it. 

**How did you test it?**

**Potential risks**


yux0 (approved):

longquanzheng:
not need it anymore

PR #1029, opened 2020-10-19
Add option to dispatch activity tasks locally

mkolodezny:
The activity dispatch optimization feature to allow worker to dispatch activity tasks through local tunnel after decisions are made. If the activity execution fails the retry(if any) goes the regular path(poll from server).
This is a performance optimization to skip activity scheduling efforts.

The corresponding server pr https://github.com/uber/cadence/pull/3609

**How did you test it?**
unit & integration tests
e2e testing

**Potential risks**
No risk. It gets enabled per domain by server and off by default


internal/internal_task_pollers.go
```diff
@@ -854,10 +960,37 @@ func (atp *activityTaskPoller) poll(ctx context.Context) (interface{}, error) {
 	return &activityTask{task: response, pollStartTime: startTime}, nil
 }
 
+func (atp *activityTaskPoller) pollLocallyDispatchedTask(ctx context.Context) (interface{}, error) {
+	task := atp.ldaTunnel.getTask()
```
meiliang86, line 964: should combine with the previous method.

internal/internal_worker.go
```diff
@@ -78,14 +78,15 @@ type (
...
-		localActivityWorker *baseWorker
-		identity            string
-		stopC               chan struct{}
+		executionParameters             workerExecutionParameters
+		workflowService                 workflowserviceclient.Interface
+		domain                          string
+		poller                          taskPoller // taskPoller to poll and process the tasks.
+		worker                          *baseWorker
+		localActivityWorker             *baseWorker
+		locallyDispatchedActivityWorker *baseWorker
```
meiliang86, line 87: We don't need a new worker for this. Can reuse the existing activity worker.

mkolodezny: Discussed offline: lets compare the implementations complexity.

mkolodezny: Decided to create the activity worker outside of workflow worker. Later we will refactor the code such that the taskPoller interface returns a channel for tasks, and move the for loop into the task poller implementation.

internal/worker.go
```diff
@@ -146,6 +146,11 @@ type (
 		// default: false
 		DisableActivityWorker bool
 
+		// EnableActivityLocalDispatch allows workflow worker to dispatch activity tasks through local tunnel after decisions are made.
+		// This is an performance optimization to skip activity scheduling efforts.
+		// If it is enabled then DisableActivityWorker option must be false.
+		EnableActivityLocalDispatch bool
```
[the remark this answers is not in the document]
yycptt: If I understand the code correctly, the same parameter will be used for both normal and local dispatch activity workers, then enabling this feature will double the # of activity workers (# of goroutines) on user worker hosts. 

Those added goroutines will also compete with the normal activity worker. If user worker is already under high load, enabling this feature may result in even longer activity execution time as there will be more concurrent activities on the host. I am not sure if this will be a concern or not.

internal/internal_task_pollers.go
```diff
@@ -164,6 +176,42 @@ func (lat *localActivityTunnel) sendTask(task *localActivityTask) bool {
 	}
 }
 
+func newLocallyDispatchedActivityTunnel(stopCh <-chan struct{}) *locallyDispatchedActivityTunnel {
+	return &locallyDispatchedActivityTunnel{
+		taskCh: make(chan *locallyDispatchedActivityTask, 25),
```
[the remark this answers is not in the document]
meiliang86: For now they don't share the same worker/processor, so there should be no contention. But this could happen if we try to combine the pollers (see comment above).
The severity of this issue is something we need to verify. There are various ways to mitigate this if it happens (e.g. add heuristics to decide whether to poll remotely or not, add cancellation logic after poll, etc).

internal/worker.go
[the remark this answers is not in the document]
meiliang86: Yes it will double the # of goroutines.
Ideally we should just combine the poller and leave the processor (and its go-routines intact). This was discussed below and it requires some more refactoring not specific to this task. So we decided to postpone to a later PR.
# of goroutines might not be a big issue for worker. Regardless there might be other resource constraints and we need to tune the parameters to find a good default value.

internal/internal_task_pollers.go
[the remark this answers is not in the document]
meiliang86: Yes making this configurable is a good suggestion.

internal/internal_task_pollers.go
```diff
@@ -164,6 +176,42 @@ func (lat *localActivityTunnel) sendTask(task *localActivityTask) bool {
...
+func (ldat *locallyDispatchedActivityTunnel) getTask() *locallyDispatchedActivityTask {
+	var task *locallyDispatchedActivityTask
+	select {
+	case task = <-ldat.taskCh:
+	case <-ldat.stopCh:
+		return nil
+	}
+
+	select {
+	case ready := <-task.readyCh:
```
meiliang86, line 195: Will this be head of line blocking? Not sure how big an issue it is in production. Should we consider separating to two channels internally?

mkolodezny: Discussed, that's related to the Yichao's comment above. If all pollers are busy with long running activities the task can stuck for a while, then we rely on the retry from the server. Will put a comment to make it clear.

yycptt (approved):
I think we should add some client side integration tests for this feature as the next step?

PR #1031, opened 2020-10-27
do not start workers if no activities or workflows are registered

mantas-sidlauskas:
**What changed?**
Worker is not started if no workflows or activities are registered. We will print info message instead.

**Why?**
To reduce resource wasting 

**How did you test it?**
Unit tests

**Potential risks**

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -368,7 +368,13 @@ func (ww *workflowWorker) Run() error {
 
 // Shutdown the worker.
 func (ww *workflowWorker) Stop() {
-	close(ww.stopC)
+	select {
+	case <-ww.stopC:
+		// channel is already closed
+	default:
+		close(ww.stopC)
+	}
+
 	// TODO: remove the stop methods in favor of the workerStopChannel
 	ww.localActivityWorker.Stop()
 	ww.worker.Stop()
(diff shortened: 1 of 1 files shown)


internal/internal_worker.go
```diff
@@ -821,7 +839,10 @@ func (aw *aggregatedWorker) Start() error {
 		}
 	}
 
-	aw.logger.Info("Started Worker")
+	if workerStarted {
```
meiliang86, line 842: Remove this log. Instead log after workflowWorker.Start() and activityWorker.Start() individually.

internal/internal_worker.go
```diff
@@ -783,28 +795,34 @@ func (aw *aggregatedWorker) Start() error {
...
 	}
 
+	workerStarted := false
+
 	if !isInterfaceNil(aw.workflowWorker) {
 		if len(aw.registry.getRegisteredWorkflowTypes()) == 0 {
-			aw.logger.Warn(
-				"Starting worker without any workflows. Workflows must be registered before start.",
+			aw.logger.Info(
+				"Worker has no workflows registered. Workflows must be registered before start. Skipping",
```
meiliang86, line 803: "Worker has no workflows registered. Workflows must be registered before start. Skipping" -> "Worker has no workflows registered, so workflow worker will not be started."
Same for activity worker.

meiliang86 (approved):
Looks good. Minor suggestions to the log messages.

internal/internal_worker.go
mantas-sidlauskas: done

internal/internal_worker.go
mantas-sidlauskas: thanks!

PR #1036, opened 2020-11-11
Make logger of test workflow environment changeable

shiwano:
**What changed?**
It makes the logger of test workflow environment changeable.

**Why?**
In workflow/activity test, I want to replace the logger with the NOP logger in order to suppress logging to stdout.
Then I tried to change the logger by `env.SetWorkerOptions` method like the below:

```go
env := s.NewTestActivityEnvironment()
env.SetWorkerOptions(worker.Options{
	Logger: zap.NewNop(),
})
```

But it is not worked because internal `testWorkflowEnvironmentImpl.setWorkerOptions` method ignore the change.
I think it would be good to be able to set the logger here.

**How did you test it?**
I added the following code to `go.mod` and made sure it works as intended:
```
replace go.uber.org/cadence => github.com/shiwano/cadence-client v0.0.0-20201111132232-8c04107a1d1d
```

**Potential risks**
Nothing comes to mind.


meiliang86 (approved):

meiliang86:
Thanks for your contribution @shiwano !

PR #1038, opened 2020-11-25
Enable query timeout of 20 seconds

andrewjdawson2016:


internal/internal_utils.go
```diff
@@ -99,8 +101,17 @@ func chanTimeout(timeout time.Duration) func(builder *contextBuilder) {
 	}
 }
 
+// newChannelContext - Get a rpc channel context for query
+func newChannelContextForQuery(ctx context.Context, options ...func(builder *contextBuilder)) (context.Context, context.CancelFunc, []yarpc.CallOption) {
```
yycptt, line 105: Is this new function used anywhere?

andrewjdawson2016: Oh gosh 🤦 ty

internal/internal_utils.go
```diff
@@ -110,8 +121,10 @@ func newChannelContext(ctx context.Context, options ...func(builder *contextBuil
 			// Make sure to not set rpc timeout lower than minRPCTimeout
 			if rpcTimeout < minRPCTimeout {
 				rpcTimeout = minRPCTimeout
-			} else if rpcTimeout > maxRPCTimeout {
+			} else if rpcTimeout > maxRPCTimeout && !isQuery {
 				rpcTimeout = maxRPCTimeout
+			} else if rpcTimeout > maxQueryRPCTimeout && isQuery {
+				rpcTimeout = maxQueryRPCTimeout
```
just-at-uber, line 127: If its easy to add, can we have a unit test which checks:
if not query, rpcTimeout = 5 seconds
if query, rpcTimeout = 20 seconds

yycptt (approved):

PR #1049, opened 2021-01-28
add missing workflowtype and activitytype metric tags

mkolodezny:
workflowtype and activitytype  tags should be added to as many metrics as possible, for easier debugging of one particular workflow / activity, and easier overall health view split by workflow.


meiliang86 (approved):
Change looks good to me. How did you test it?

Issue #1051, opened 2021-02-03
Activity auto heartbeat 


longquanzheng:
**Is your feature request related to a problem? Please describe.**
Most people want to heartbeat their activities when the workers are still alive. So that when hosts going down the activities can be retried. Currently it requires users to implement heartbeat which is tedious to do. 

This will be more useful than graceful shutdown feature: graceful shutdown doesn't guarantee to work as expected because when a host going down, it probably cannot wait too long. Sometimes it's just crash within a second.

https://github.com/uber-go/cadence-client/issues/818

**Proposed Solution**
Provide an option to automatically sending heartbeat for activities. 

**Additional context**
NA

PR #1053, opened 2021-02-04
Option to automatically record activity heartbeats

willgorman:
Fixes https://github.com/uber-go/cadence-client/issues/1051

**What changed?**
This change adds EnableAutoHeartbeat to RegisterActivityOptions. If enabled, cadence-client will automatically send a heartbeat for the registered activity if it has a HeartbeatTimeout > 0. The heartbeat frequency will be half of the HeartbeatTimeout. With this change it becomes possible to ensure retry of interrupted activities with only a config change for existing activities.

**Why?**
When an activity is interrupted by a worker process termination by default Cadence does not realize that the activity has been lost and apply any RetryPolicy it may have.  Activities with a HeartbeatTimeout that use `activity.RecordHeartbeat` may be retried if the heartbeat is missed but adding heartbeat code to every activity that needs to retry can be burdensome so this adds the option to do it automatically.

**How did you test it?**
Tested locally using a modified workflow from https://github.com/uber-common/cadence-samples with a HeartbeatTimeout and RetryPolicy for an activity and a sleep in the same activity so that the worker could easily be terminated while the activity was running.  After restarting the worker this confirmed that the activity completed on a second attempt and the workflow was completed as well.

**Potential risks**
Hopefully minimal since the option is disabled by default.  It also still applies batching to the activity heartbeats so it should not add many extra heartbeats even if used with activities that already have heartbeats.  I am not sure if it could interfere with activities that already use details to record progress with heartbeats.


internal/worker.go
```diff
@@ -205,6 +205,10 @@ type (
 		// Optional: Sets opentracing Tracer that is to be used to emit tracing information
 		// default: no tracer - opentracing.NoopTracer
 		Tracer opentracing.Tracer
+
+		// Optional: Automatically send heartbeats for any activity with a HeartbeatTimeout
+		// default: false
+		ActivityAutoHeartbeat bool
```
longquanzheng, line 211: Having this option is fine for me

However it will be more ideal if there is another option also letting activity registration to disable autoHeartbeat for some activities.
Because people may want to have some with auto heartbeat,but some not. I believe 95% of the time people want autoHeartbeat, but some 5% they also want to do heartbeat in their code to control the heartbeat details.

internal/worker.go
```diff
@@ -205,6 +205,10 @@ type (
 		// Optional: Sets opentracing Tracer that is to be used to emit tracing information
 		// default: no tracer - opentracing.NoopTracer
 		Tracer opentracing.Tracer
+
+		// Optional: Automatically send heartbeats for any activity with a HeartbeatTimeout
```
longquanzheng, line 209: Also we should state it clear here that this cannot be used with calling heartbeat in the code with details. It will lead to wrong behavior(losing the details progerss)

longquanzheng:
@willgorman thank you for the contribution!

internal/internal_task_handlers.go
```diff
@@ -1819,6 +1821,26 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
...
 	ctx, span := createOpenTracingActivitySpan(ctx, ath.tracer, time.Now(), activityType, t.WorkflowExecution.GetWorkflowId(), t.WorkflowExecution.GetRunId())
 	defer span.Finish()
+
+	if ath.autoHeartbeat && t.HeartbeatTimeoutSeconds != nil && *t.HeartbeatTimeoutSeconds > 0 {
+		go func() {
+			autoHbInterval := time.Duration(*t.HeartbeatTimeoutSeconds) * time.Second / 2
+			ticker := time.NewTicker(autoHbInterval)
+			defer ticker.Stop()
+			for {
+				select {
```
longquanzheng, line 1831: I am not super familar with this client yet. Should this goroutine also listen to :
```
	closeCh               chan struct{}
	workerStopChannel     <-chan struct{}
```

internal/worker.go
willgorman: That makes sense, I had wondered if it would be better to have it set per activity instead of for the entire worker.  If this approach interferes with existing heartbeats that record details from the activity then it might be hard to use if it's enabled for all activities in the worker.  I can try to add that also.  Just to make sure I understand what that would look like:

* Leave this option here to make it possible to enable for all activities
* Add a new `DisableAutoHeartbeat` field to [RegisterActivityOptions](https://github.com/uber-go/cadence-client/blob/0256258b905b677f2f38fcacfbda43398d236309/internal/activity.go#L59).  This would probably need to be stored in `activityExecutor` after registration and exposed to `activityTaskHandlerImpl` by adding a new method to the internal `activity` interface.

longquanzheng: Yes, that's what I meant. 

If we are going this route, there are some more details I want to add:
* Probably rename `DisableAutoHeartbeat` to `DisableAutoHeartbeatIfWorkerHasEnabled`. Because `DisableAutoHeartbeat` sounds like it's already enabled by default. (or other better names if you can think of)
* This must be used only when AutoHeartbeat if enabled for worker, otherwise should return error
* Add clear comments on how to use it, and why this.

However, I feel that's quite confusing and error-prone. Because there are two options to control one behavior. Another route is that we only hanve `enableAutoHeartbeat` in RegisterActivityOptions. That will be much more clear for users. That will require every activity to use the option if they need autoHeartbeat, instead of a global option for the whole worker. But I believe the amount of code shouldn't be too much different -- as the option can be a global one and shared by all registration. 

I am leaning more tho the second approach, because it should be more friendly to users like you What do you think?

willgorman: I agree that it seems confusing to have an option in the worker and an overriding option when registering activities.  After giving it some more thought, it does feel like activity registration is the best place to have the option and not the worker.  It's the implementation of the activity that determines whether an auto-heartbeat is appropriate or not so it shouldn't depend on how the worker is configured.

internal/internal_task_handlers.go
```diff
@@ -1671,18 +1707,18 @@ func (i *cadenceInvoker) Heartbeat(details []byte, skipBatching bool) error {
...
-			i.Unlock()
 
 			if detailsToReport != nil {
 				// TODO: there is a potential race condition here as the lock is released here and
-				// locked again in the Hearbeat() method. This possible that a heartbeat call from
+				// locked again in the Heartbeat() method. This possible that a heartbeat call from
 				// user activity grabs the lock first and calls internalHeartBeat before this
 				// batching goroutine, which means some activity progress will be lost.
-				i.Heartbeat(*detailsToReport, false)
+				i.heartbeatAndScheduleNextRun(*detailsToReport)
```
yycptt, line 1719: Do we still have the race condition given that `Unlock()` has been moved after this call?

internal/internal_task_handlers.go
```diff
@@ -1623,27 +1623,63 @@ type cadenceInvoker struct {
 	heartBeatTimeoutInSec int32       // The heart beat interval configured for this activity.
 	hbBatchEndTimer       *time.Timer // Whether we started a batch of operations that need to be reported in the cycle. This gets started on a user call.
 	lastDetailsToReport   *[]byte
+	detailsToReport       *[]byte
 	closeCh               chan struct{}
 	workerStopChannel     <-chan struct{}
 }
 
-func (i *cadenceInvoker) Heartbeat(details []byte, skipBatching bool) error {
+func (i *cadenceInvoker) Heartbeat(details []byte) error {
```
yycptt, line 1631: nit: probably we need to add some comment here saying `Heartbeat` should not be used together with the other two method as it may cause the progress to be lost?

internal/internal_task_handlers.go
```diff
@@ -1819,6 +1855,26 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
...
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					hbErr := invoker.BackgroundHeartbeat()
+					if IsCanceledError(hbErr) {
+						return
+					}
```
yycptt, line 1872: My understanding is activity can ignore the cancellation and continue execution, in which case we should continue the background heartbeat?

willgorman: Ah, ok.  I didn't really understand that about cancellation but I think that makes sense if that's the case. If the cancelled activity keeps running but the heartbeat stops then a retry of the activity could be started while the original activity is still running?  Sounds like this should just ignore any errors and always keep running until the context is done?

internal/internal_task_handlers.go
meiliang86: No I don't think so. I'll remove this comment.

internal/internal_task_handlers.go
meiliang86: Yes will add comment.

internal/internal_task_handlers.go
meiliang86: Checked. Added workerStopChannel. I don't think we need to check closeCh an it's the same as ctx.Done.

internal/internal_task_handlers.go
meiliang86: Change the behavior to continue heartbeat in case of error, until ctx is done.

meiliang86:
@yycptt @longquanzheng This should be ready to review.

internal/internal_task_handlers.go
```diff
@@ -1819,6 +1852,33 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
...
+			for {
+				select {
+				case <-ath.workerStopCh:
+					return
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					hbErr := invoker.BackgroundHeartbeat()
+					if hbErr != nil && IsCanceledError(hbErr) {
+						ath.logger.Error("Activity auto heartbeat error.",
```
yycptt, line 1870: nit: shall we also record errors other than canceled?

yycptt (approved):

internal/activity.go
```diff
@@ -279,8 +283,16 @@ func RecordActivityHeartbeat(ctx context.Context, details ...interface{}) {
...
 type ServiceInvoker interface {
-	// Returns ActivityTaskCanceledError if activity is cancelled
-	Heartbeat(details []byte, skipBatching bool) error
+	// All the heartbeat methods will return ActivityTaskCanceledError if activity is cancelled.
+	// Heartbeat sends a record heartbeat request to Cadence server directly without buffering.
+	// It should only be used by the sessions framework.
+	Heartbeat(details []byte) error
+	// BatchHeartbeat sends heartbeat on the first attempt, and batches additional requests
+	// to send it later according to heartbeat timeout.
+	BatchHeartbeat(details []byte) error
```
longquanzheng, line 292: That's so much clearer! Thanks and good job!

internal/internal_task_handlers.go
```diff
@@ -1622,28 +1622,64 @@ type cadenceInvoker struct {
...
+	return i.heartbeatAndScheduleNextRun(details)
+}
+
+func (i *cadenceInvoker) heartbeatAndScheduleNextRun(details []byte) error {
 	isActivityCancelled, err := i.internalHeartBeat(details)
 
 	// If the activity is cancelled, the activity can ignore the cancellation and do its work
 	// and complete. Our cancellation is co-operative, so we will try to heartbeat.
-	if (err == nil || isActivityCancelled) && !skipBatching {
+	if err == nil || isActivityCancelled {
```
longquanzheng, line 1679: I don't understand this...why we still need to do the work if activity is already canceled?

longquanzheng (approved):

internal/internal_task_handlers.go
meiliang86: Yes. I actually only want to log error if it's not cancelledError. Updated.

willgorman:
Thanks for your help with this @meiliang86 and @longquanzheng!

PR #1055, opened 2021-02-16
Visibility query constructor for workflow shadower

yycptt:
**What changed?**
Visibility query constructor for workflow shadower

**Why?**
Visibility query constructor for workflow shadower

**How did you test it?**
Tested locally

**Potential risks**
N/A

### DIFF
--- a/internal/query_constructor.go
+++ b/internal/query_constructor.go
@@ -1,0 +1,138 @@
+package internal
+
+import (
+	"fmt"
+	"math"
+	"strings"
+	"time"
+
+	"go.uber.org/cadence/.gen/go/shared"
+)
+
+type (
+	// WorkflowStatus represents the status of a workflow
+	WorkflowStatus string
+)
+
+const (
+	// WorkflowStatusOpen is the WorkflowStatus for open workflows
+	WorkflowStatusOpen WorkflowStatus = "OPEN"
+	// WorkflowStatusClosed is the WorkflowStatus for closed workflows
+	WorkflowStatusClosed WorkflowStatus = "CLOSED"
+)
+
+var (
+	// WorkflowStatusCompleted is the WorkflowStatus for completed workflow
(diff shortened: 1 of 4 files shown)


internal/query_constructor.go
```diff
@@ -0,0 +1,138 @@
...
+	q.appendPartialQuery(strings.Join(workflowStatusQueries, " or "))
+	return q
+}
+
+func (q *queryConstructorImpl) StartTime(minStartTime, maxStartTime time.Time) QueryConstructor {
+	startTimeQueries := make([]string, 0, 2)
+	if !minStartTime.IsZero() {
+		startTimeQueries = append(startTimeQueries, fmt.Sprintf(keyStartTime+` >= %v`, minStartTime.UnixNano()))
+	}
+	if !maxStartTime.Equal(maxTimestamp) {
```
yux0, line 101: What if the maxStartTime is zero?

internal/query_constructor.go
```diff
@@ -0,0 +1,138 @@
...
+		startTimeQueries = append(startTimeQueries, fmt.Sprintf(keyStartTime+` <= %v`, maxStartTime.UnixNano()))
+	}
+
+	q.appendPartialQuery(strings.Join(startTimeQueries, " and "))
+	return q
+}
+
+func (q *queryConstructorImpl) Query() string {
```
yux0, line 109: maybe named this query builder and build() function?

internal/query_constructor.go
yycptt: This component assumes whatever value got passed in should be used in the final query. Shadower is responsible for converting and filling unspecified values such as sampling rate, max timestamp etc. but that logic is not in this diff.

internal/query_constructor.go
yycptt: Makes sense. Will update.

yycptt: Renamed to `QueryBuilder` and `Build()`

yux0 (approved):

PR #1058, opened 2021-02-17
Workflow shadower

yycptt:
**What changed?**
Workflow shadower

**Why?**
For automating workflow replay tests

**How did you test it?**
Tested locally

**Potential risks**
N/A


internal/workflow_shadower.go
```diff
@@ -0,0 +1,308 @@
+// Copyright (c) 2017-2020 Uber Technologies Inc.
```
yux0, line 1: maybe update the LICENSE file to 2017-2021

internal/workflow_shadower.go
```diff
@@ -0,0 +1,308 @@
...
+	"go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
+	"go.uber.org/cadence/.gen/go/shared"
+	"go.uber.org/cadence/internal/common"
+	"go.uber.org/cadence/internal/common/backoff"
+	"go.uber.org/cadence/internal/common/util"
+	"go.uber.org/zap"
+)
+
+var (
+	errInvalidTimeFilter = errors.New("should specify either TimeRange or Min, Max timestamp")
```
yux0, line 43: just min, max timeStamp

internal/workflow_shadower.go
```diff
@@ -0,0 +1,308 @@
...
+func (s *WorkflowShadower) Run() error {
+	if !atomic.CompareAndSwapInt32(&s.status, statusInitialized, statusStarted) {
+		return errors.New("Workflow shadower already started")
+	}
+
+	return s.shadowWorker()
+}
+
+// Stop stops WorkflowShadower and wait up to one miniute for all goroutines to finish before returning
+func (s *WorkflowShadower) Stop() {
```
yux0, line 134: Do you plan to define this in the interface?

internal/workflow_shadower.go
yycptt: Good catch. I also added license header to the query_builder files I committed in the previous diff. 

I will send out a separate diff to update the LICENSE file.

internal/workflow_shadower.go
yycptt: Looks like the error is not needed after removing the `TimeRange` field. Removed.

internal/workflow_shadower.go
yycptt: Probably not at this time. I will keep the implementation here for now and if user needs that capability after we rollout the feature, I can easily add it to the interface.

internal/workflow_shadower.go
yycptt: https://github.com/uber-go/cadence-client/pull/1059

internal/workflow_shadower.go
```diff
@@ -0,0 +1,305 @@
...
+		if err != nil {
+			return err
+		}
+
+		for _, execution := range scanWorkflowResp.Executions {
+			if s.clock.Now().After(expirationTime) {
+				return nil
+			}
+
+			if rand.Float64() >= s.options.SamplingRate {
```
yux0, line 178: if the sampling is in the scanWorkflowExecutionsWithRetry, will it better if we want to convert it to an activity?

yycptt: Yes, I have already converted into in an activity in a later diff. But the activity diff is build on top of this one, so would like to land this one first.

yux0 (approved):

PR #1060, opened 2021-02-18
Pin versions with go.mod, similar to server

Groxx:
The versioning strategy unfortunately is leaking some dependencies, and codegen isn't reproducible any more.
This update mimics how github.com/uber/cadence does dependencies, and is more "normal" for go modules anyway.

Implications of these changes include:
- Some build and runtime dependencies have changed versions in relatively minor ways.  This mostly brings them more in line with the versions used to generate code.
    - I have not tried to upgrade any back to the previous versions, but it would be easy to try.
    - Generated code appears identical.
- Users may end up downloading more dependencies (e.g. staticcheck), maybe not, depending on how Go decides to handle non-buildable transitive dependencies.  They've always _intended_ to prune them implicitly, but I do not know if that is happening yet.
    - They do not appear to be _constrained_ by these dependencies, however, as they have no actual import, and do not appear in their go.mod files.  Upgrading or downgrading tools in a test project worked and did not require `replace` directives.
- For good or bad, transitive dependencies of tools and this library are now bound together.
    - This is pretty common and not usually an issue, but if it becomes one: we can switch conflicting tools to a separate tools.mod and deal with the internal-only quirks that causes.
    - I can do that ^ right now if we want to be paranoid, and it'd eliminate the chance of tool versions impacting users, but it opens the door to version skew.  Neither option is ideal ¯\\\_(ツ)_/¯

And since this does away with the "X commits / versions behind current" notification: I've added `make deps` (and `make deps-all`), which works the same way as in uber/cadence/pull/4000.  Hopefully it's understandable enough to modify easily.


go.mod
```diff
@@ -37,20 +36,19 @@ require (
 	github.com/uber/jaeger-client-go v2.22.1+incompatible
 	github.com/uber/jaeger-lib v2.2.0+incompatible // indirect
 	github.com/uber/tchannel-go v1.14.0
-	go.uber.org/atomic v1.5.1
+	go.uber.org/atomic v1.4.0
```
Groxx, line 39: This (and x/lint) downgrade is a bit odd.  Possibly it's not imported locally + transitive dependencies changed and this is the new minimum?
It's very likely safe / good to upgrade to latest.  Latest does make a breaking change though, disallowing comparison between atomic wrappers (which is a good thing, as that's racy behavior).

Groxx:
Current deps output for comparison:
```
❯ make deps
220 days: github.com/uber/jaeger-client-go  	v2.22.1+incompatible -> v2.25.0+incompatible
371 days: golang.org/x/net  	v0.0.0-20200114155413-6afb5195e5aa -> v0.0.0-20210119194325-5f4716e94777
466 days: github.com/opentracing/opentracing-go  	v1.1.0 -> v1.2.0
501 days: go.uber.org/atomic  	v1.4.0 -> v1.7.0
503 days: honnef.co/go/tools  	v0.0.1-2019.2.3 -> v0.1.1
573 days: github.com/uber/tchannel-go  	v1.14.0 -> v1.21.0
733 days: github.com/kisielk/errcheck  	v1.2.0 -> v1.5.0
740 days: github.com/stretchr/testify  	v1.3.0 -> v1.7.0
871 days: go.uber.org/zap  	v1.8.0 -> v1.16.0
908 days: github.com/uber-go/tally  	v3.3.1+incompatible -> v3.3.17+incompatible
954 days: golang.org/x/lint  	v0.0.0-20180428170328-470b6b0bb300 -> v0.0.0-20201208152925-83fdc39ff7b5
958 days: go.uber.org/goleak  	v0.10.0 -> v1.1.10
1047 days: go.uber.org/yarpc  	v1.29.1 -> v1.52.0
1059 days: go.uber.org/thriftrw  	v1.11.0 -> v1.26.0
1167 days: golang.org/x/time  	v0.0.0-20170927054726-6dc17368e09b -> v0.0.0-20201208040808-7e3f01d25324
1352 days: github.com/pborman/uuid  	v0.0.0-20160209185913-a97ce2ca70fa -> v1.2.1
1435 days: github.com/sirupsen/logrus  	v0.11.5 -> v1.8.0
1506 days: github.com/apache/thrift  	v0.0.0-20161221203622-b2a4d4ae21c7 -> v0.14.0
```

just-at-uber (approved):

meiliang86 (approved):

PR #1064, opened 2021-03-12
Implement shadower worker

yycptt:
**What changed?**
Implement shadower worker

**Why?**
workflow shadowing

**How did you test it?**
Tested locally for shadowing hello workflow sample workflows

**Potential risks**
N/A, shadow worker is disabled by default.


internal/worker.go
```diff
@@ -205,6 +205,17 @@ type (
 		// Optional: Sets opentracing Tracer that is to be used to emit tracing information
 		// default: no tracer - opentracing.NoopTracer
 		Tracer opentracing.Tracer
+
+		// Optional: Enable worker for running shadowing workflows to replay existing workflows
+		// If set to true, worker will run in shadow mode and all other workers (decision, activity, session)
+		// will be disabled to prevent them from updating existing workflow states.
+		// default: false
+		EnableShadowWorker bool
```
yux0, line 213: Could we use *ShadowOptions to indicate if the shadow worker is enabled?

internal/internal_worker.go
```diff
@@ -865,6 +866,23 @@ func (aw *aggregatedWorker) Start() error {
 		}
 	}
 
+	if !isInterfaceNil(aw.shadowWorker) {
+		if err := aw.shadowWorker.Start(); err != nil {
```
yux0, line 870: Should it stop other workers regardless the error?

yycptt: Sorry, I don't quite follow. Currently, all other workers will be stopped if there's any error on start up.

internal/worker.go
yycptt: Yes, technically it will work. But I kind of think having a separate option for switching shadower on and off is more useful. So user can always have the same code for specifying the options, and using another option to turn it on or off. What's your thought?

yux0: No strong opinion on it

internal/internal_worker.go
yux0: Oh, I see. Should it stop other workers if the shadow worker starts?

yycptt: Yes, it's already doing it. Check [L1274](https://github.com/uber-go/cadence-client/pull/1064/files#diff-81f3fd107107ea44dc42925f83354a318280b38ddb7223548848cfc436c1da58R1274) below.

yux0 (approved):

internal/worker.go
yux0: Second thought. Customers can comment out the shadow options if they temporary disable the shadow worker?

PR #1066, opened 2021-03-18
Fix the problem of ignoring the detail parameter in TerminateWorkflow function 

lantian789:
**What changed?**
Add the **Details** field in TerminateWorkflowExecutionRequest in the terminateWorkflowExecutionRequest function

**Why?**
TerminateWorkflowExecutionRequest has the **Details** fields, but it was not included in the request and the details parameter passed to this function is unused. It should be included in the request so that detail parameter passed into this function is used.

**How did you test it?**
After making this change to my local cadence client project as I mentioned above, I replaced the go.uber.org/cadence project with my updated local cadence client project in my own workflow project. In order to validate my changes, I triggered a workflow and terminated it with detail information, and I checked the result through the cadence web UI and the detail information was clearly showed instead of showing null all the time in the past in this scenarios. 

**Potential risks**
I did not aware any potential risks by making this changes to this project. 

**Before making the change:**
Screen Shot 2021-03-18 at 4 09 20 PM

**After making the change:**
Screen Shot 2021-03-18 at 3 36 46 PM

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -473,6 +473,7 @@ func (wc *workflowClient) TerminateWorkflow(ctx context.Context, workflowID stri
 			RunId:      getRunID(runID),
 		},
 		Reason:   common.StringPtr(reason),
+		Details:  details,
 		Identity: common.StringPtr(wc.identity),
 	}
 


yycptt (approved):
Thanks for the fix!

lantian789:
Could you merge my pull request as soon as possible ?

PR #1069, opened 2021-03-19
Minor test enhancement for local activity retries

Groxx:
While looking into a user's issue with local activities retrying and not seeming to honor NonRetriableErrorReasons, I noticed that we had no tests for non-retriable local activity errors.
    
So this test now mimics `Test_ActivityRetry`, but with local activities and maybe better names.  No problems found.

### DIFF
--- a/internal/internal_workflow_testsuite_test.go
+++ b/internal/internal_workflow_testsuite_test.go
@@ -2438,8 +2438,6 @@ func (s *WorkflowTestSuiteUnitTest) Test_LocalActivityRetry() {
 		if err != nil {
 			return "", err
 		}
-
-		Sleep(ctx, time.Hour)
 		return result, nil
 	}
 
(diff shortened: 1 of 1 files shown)


internal/internal_workflow_testsuite_test.go
```diff
@@ -2421,11 +2428,18 @@ func (s *WorkflowTestSuiteUnitTest) Test_LocalActivityRetry() {
...
+		s.Equal("bad-bug", badBug.Reason())
+
+		var result string
+		err = ExecuteLocalActivity(ctx, retriableFn).Get(ctx, &result)
 		if err != nil {
-			return int32(-1), err
+			return "", err
 		}
+
+		Sleep(ctx, time.Hour)
```
yycptt, line 2442: what's the purpose of this Sleep?

Groxx: ah, sorry.  leftovers from testing it in a workflow.Go (to make sure that wasn't changing behavior).  I'll remove it.

yycptt (approved):

PR #1070, opened 2021-03-23
Set default shadow workflow status to OPEN

yycptt:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/workflow_shadower.go
+++ b/internal/workflow_shadower.go
@@ -63,8 +63,8 @@ type (
 
 		// Optional: A list of workflow status.
 		// The list will be used to construct WorkflowQuery. Only workflows with status listed will be replayed.
-		// accepted values (case-insensitive): OPEN, CLOSED, COMPLETED, FAILED, CANCELED, TERMINATED, CONTINUED_AS_NEW, TIMED_OUT
-		// default: empty list, which matches all workflow status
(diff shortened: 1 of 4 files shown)


internal/workflow_shadower.go
```diff
@@ -63,8 +63,8 @@ type (
 
 		// Optional: A list of workflow status.
 		// The list will be used to construct WorkflowQuery. Only workflows with status listed will be replayed.
-		// accepted values (case-insensitive): OPEN, CLOSED, COMPLETED, FAILED, CANCELED, TERMINATED, CONTINUED_AS_NEW, TIMED_OUT
-		// default: empty list, which matches all workflow status
+		// accepted values (case-insensitive): OPEN, CLOSED, ALL, COMPLETED, FAILED, CANCELED, TERMINATED, CONTINUED_AS_NEW, TIMED_OUT
+		// default: OPEN, which matches only open workflows
 		WorkflowStatus []string
```
yycptt, line 68: @yux0 What's your opinion on changing the type for `[]string` to `[]WorkflowStatus` so it's typed? Then I can expose all the constants defined in `query_build.go`. I am not sure if that will confuse users or not, I mean having both the WorkflowCloseStatus type defined by thrift and WorkflowStatus type defined by the client.

yux0: Yeah. I think it should be fine.

yux0 (approved):

PR #1071, opened 2021-03-24
Minor cleanup of a test file

Groxx:
Just noticed some junk while reading.
This test probably shouldn't be passing given the logs... but this rewrite does not change behavior.

### DIFF
--- a/test/replaytests/reply_test.go
+++ b/test/replaytests/reply_test.go
@@ -21,47 +21,23 @@
 package replaytests
 
 import (
+	"strings"
 	"testing"
 
-	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


mkolodezny (approved):

emrahs (approved):
Let's also rename the file next time :)

Groxx:
> Let's also rename the file next time :)

Yeah, I noticed that too :)  tbh I think I want to modify this further, and I'll tackle it then - there are some other issues in this test:
- the replays are failing to replay, but there's no error (a long-standing issue with this api)
- I can't `-count 2` (or anything beyond 1) or it starts logging after tests finish, which crashes the test process (by design).
- following ^, I assume there's something running after the replay func, but there's no "shut down the replay worker" method.
- filename
- moar tests

PR #1094, opened 2021-05-01
Ignore workflowPanicError in query task and add warn logs

yycptt:
**What changed?**
Ignore workflowPanicError in query task and add warn logs

**Why?**
Bug fix in https://github.com/uber-go/cadence-client/pull/1082 checks workflowPanicError instead PanicError for query task, this may break existing customers that are using query feature.

**How did you test it?**
Unit test

**Potential risks**

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1474,6 +1474,8 @@ func (wth *workflowTaskHandlerImpl) completeWorkflow(
 			wth.logger.Warn("Encountered PanicError in workflow query task",
 				zap.String(tagWorkflowID, task.WorkflowExecution.GetWorkflowId()),
 				zap.String(tagRunID, task.WorkflowExecution.GetRunId()),
+				zap.String(tagPanicError, panicErr.Error()),
+				zap.String(tagPanicStack, panicErr.StackTrace()),
 			)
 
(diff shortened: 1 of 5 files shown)


Shaddoll (approved):

internal/internal_task_handlers.go
```diff
@@ -917,12 +924,19 @@ ProcessEvents:
 		}
 	}
 	if nonDeterministicErr == nil && w.err != nil {
-		if isReplayTest {
-			// NOTE: we should check the following error regardless if it's in replay test or not
-			// but since we are not checking it previously, it may break existing customers workflow
-			if panicErr, ok := w.err.(*workflowPanicError); ok && panicErr.value != nil {
-				if _, isStateMachinePanic := panicErr.value.(stateMachineIllegalStatePanic); isStateMachinePanic {
+		if panicErr, ok := w.err.(*workflowPanicError); ok && panicErr.value != nil {
+			if _, isStateMachinePanic := panicErr.value.(stateMachineIllegalStatePanic); isStateMachinePanic {
```
meiliang86, line 928: What are the other possible cases? Why not log for all?

yycptt: Only `stateMachineIllegalStatePanic` error is counted as non-deterministic error. 

Other cases may be panic in the workflow code that is caught by the library. For those cases, there's another check (with logging) when completing decision task and responding decision task failed.

PR #1096, opened 2021-05-11
Truncate history for query based on nextEventID

yycptt:
**What changed?**
Truncate history for query based on nextEventID

**Why?**
Query may replay history event passed nextEventID and encounter state machine error, which is treated as non-deterministic error.

**How did you test it?**
Unit test.

**Potential risks**
In the worst case, only partial history will be applied for query.


internal/internal_task_pollers.go
```diff
@@ -798,13 +799,25 @@ func (wtp *workflowTaskPoller) poll(ctx context.Context) (interface{}, error) {
...
+	startEventID := response.GetStartedEventId()
+	nextEventID := response.GetNextEventId()
+	if nextEventID != 0 &&
+		startEventID != 0 &&
+		nextEventID-1 != startEventID {
+		wtp.logger.Warn("Invalid PollForDecisionTaskResponse, nextEventID doesn't match startedEventID",
+			zap.Int64("StartedEventID", startEventID),
+			zap.Int64("NextEventID", nextEventID),
+		)
+		wtp.metricsScope.Counter(metrics.DecisionPollInvalidCounter).Inc(1)
```
yux0, line 811: Should it return error or stop processing the response if it is invalid?

internal/internal_task_pollers.go
```diff
@@ -896,6 +916,18 @@ func newGetHistoryPageFunc(
 			}
 			return h, nil, nil
 		}
+
+		// atDecisionTaskCompletedEventID (startedEventID) is 0 for query task
+		// maxEventID is 0 for old version of the server
+		// TODO: Apply the check to decision tasks (remove the last condition)
+		// after validating maxEventID always equal to atDecisionTaskCompletedEventID (startedEventID).
+		// For now only apply to query task to be safe.
+		if size > 0 && maxEventID != 0 && h.Events[size-1].GetEventId() > maxEventID && atDecisionTaskCompletedEventID == 0 {
```
yux0, line 925: nit: maybe have some functions and save some comments:
if isQueryTaskWithNextEventID(maxEventID, atDecisionTaskCompletedEventID) && shouldTruncateHistory(size, maxEventID) {
...
}

internal/internal_task_pollers.go
yycptt: Yes, but for this release I would like to be extra careful and first validating if the assumption here is true. If we don't see any logs and metrics for invalid decision we can return error here.

yux0 (approved):

PR #1099, opened 2021-06-01
[grpc] Added sdk errors and their conversion

vytautas-karpavicius:
**What changed?**
Introduces error types distinct from rpc types. Ultimately we don't want to leak any rpc types via public client API surface. Therefore adding internal error types and mapping from yarpc error codes/details to them. These are based on errors returned by the [server](https://github.com/uber/cadence/blob/master/common/types/mapper/proto/errors.go#L34).

Note: this is merging to go-sdk branch used for grpc development (not master).

**Why?**
GRPC migration

**How did you test it?**
Unit test.

**Potential risks**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/apache/thrift v0.13.0
 	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a
+	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/golang/mock v1.4.4
 	github.com/kisielk/errcheck v1.2.0
 	github.com/opentracing/opentracing-go v1.1.0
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -16,6 +17,7 @@ require (
 	github.com/uber/jaeger-client-go v2.22.1+incompatible
 	github.com/uber/tchannel-go v1.16.0
 	go.uber.org/atomic v1.7.0
+	go.uber.org/fx v1.10.0 // indirect
```
yux0, line 20: Just curious, why the fx gets introduced here?

yux0 (approved):

go.mod
vytautas-karpavicius: This is the first time use of generated proto types. That package also includes YARPC interfaces which in turn references fx.

PR #1100, opened 2021-06-02
Decouple Client from worker part of SDK

vytautas-karpavicius:
**What changed?**
Decouple `Client` from "worker" part of the SDK. There was only a single place where NewClient is instantiated from within worker. Several other places (related to heartbeats) instead call a common function shared between Client and workers. Do the same here as well.

**Why?**
This coupling makes it harder to make grpc changes, as switching `workflowserviceclient.Interface` in one place quickly propagates everywhere. With this change, it becomes possible to independently change `Client` and workers - thus enabling smaller changes.

**How did you test it?**
Existing tests.

**Potential risks**

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -326,24 +326,7 @@ func (wc *workflowClient) SignalWorkflow(ctx context.Context, workflowID string,
 	if err != nil {
 		return err
 	}
-
-	request := &s.SignalWorkflowExecutionRequest{
-		Domain: common.StringPtr(wc.domain),
-		WorkflowExecution: &s.WorkflowExecution{
-			WorkflowId: common.StringPtr(workflowID),
-			RunId:      getRunID(runID),
-		},
-		SignalName: common.StringPtr(signalName),
-		Input:      input,
-		Identity:   common.StringPtr(wc.identity),
-	}
-
-	return backoff.Retry(ctx,
-		func() error {
-			tchCtx, cancel, opt := newChannelContext(ctx)
-			defer cancel()
-			return wc.workflowService.SignalWorkflowExecution(tchCtx, request, opt...)
-		}, createDynamicServiceRetryPolicy(ctx), isServiceTransientError)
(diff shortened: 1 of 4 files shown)


internal/internal_workflow_client.go
```diff
@@ -326,24 +326,7 @@ func (wc *workflowClient) SignalWorkflow(ctx context.Context, workflowID string,
 	if err != nil {
```
mkolodezny, line 326: nit: the encoding could be moved to common signal function

internal/session.go
```diff
@@ -540,9 +540,20 @@ func (env *sessionEnvironmentImpl) AddSessionToken() {
 
 func (env *sessionEnvironmentImpl) SignalCreationResponse(ctx context.Context, sessionID string) error {
 	activityEnv := getActivityEnv(ctx)
-	client := activityEnv.serviceInvoker.GetClient(activityEnv.workflowDomain, &ClientOptions{})
-	return client.SignalWorkflow(ctx, activityEnv.workflowExecution.ID, activityEnv.workflowExecution.RunID,
-		sessionID, env.getCreationResponse())
+
+	signalInput, err := encodeArg(activityEnv.dataConverter, env.getCreationResponse())
```
mkolodezny, line 544: idem

mkolodezny (approved):
LGTM, just a minor comment

internal/session.go
yycptt, line 544: Just to clarify, the use of `activityEnv.dataConverter` instead of the default data converter is a bug fix?

vytautas-karpavicius: No bug. This is actually a good observation. Signal input here is controller by client itself, so I should probably leave default converter.

internal/internal_workflow_client.go
vytautas-karpavicius: I would prefer to keep it outside, at least with this PR.
It is consistent with other functions used there that way. I could refactor all of them with another PR if you like.

internal/session.go
vytautas-karpavicius: Updated.

PR #1103, opened 2021-06-07
Support Client Feature flags

demirkayaender:
**What changed?**
Shipping WorkflowExecutionAlreadyCompletedError broke some customers' workflow logic since it was changing previous contract between their workflow and cadence server. With this change, they can enable the new feature only when they are ready. 

Changes need to go in with the following orders
- IDL changes: https://github.com/uber/cadence-idl/pull/73
- Client-side changes: https://github.com/uber-go/cadence-client/pull/1103
- Server-side changes: https://github.com/uber/cadence/pull/4257
- Samples changes: https://github.com/uber-common/cadence-samples/pull/45

**Why?**
Without this change, shipping new features will break customer workflows.

**How did you test it?**
`go test -run TestIntegrationSuite github.com/uber/cadence/host -testify.m TestSignalWorkflow$ -v`

CLI + helloworkflow and cron workflow

**Potential risks**

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -47,6 +47,9 @@ type (
 	// Options are optional parameters for Client creation.
 	Options = internal.ClientOptions
 
+	// FeatureFlags define which breaking changes can be enabled for client
+	FeatureFlags = internal.FeatureFlags
+
 	// StartWorkflowOptions configuration parameters for starting a workflow execution.
 	StartWorkflowOptions = internal.StartWorkflowOptions
 
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -24,6 +24,7 @@ package internal
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/signal"
@@ -54,6 +55,8 @@ const (
 	clientImplHeaderName  = "cadence-client-name"
 	clientImplHeaderValue = "uber-go"
 
+	clientFeatureFlagsHeaderName = "cadence-client-feature-flags"
+
 	// defaultRPCTimeout is the default tchannel rpc call timeout
 	defaultRPCTimeout = 10 * time.Second
 	//minRPCTimeout is minimum rpc call timeout allowed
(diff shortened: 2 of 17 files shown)


client/client.go
```diff
@@ -47,6 +47,9 @@ type (
 	// Options are optional parameters for Client creation.
 	Options = internal.ClientOptions
 
+	// FeatureFlags define which breaking changes can be enabled for client
+	FeatureFlags = internal.FeatureFlags
+
```
Groxx, line 52: this diff is fairly large already, but tbh I wonder if it's worth saying "nope, we should stop doing this immediately", and doing the new type + cast-or-whatever.  IDEs and code-generators really don't play well with these internal-aliases...

internal/internal_utils.go
```diff
@@ -64,15 +67,53 @@ const (
...
-	yarpcCallOptions = []yarpc.CallOption{
+	_yarpcCallOptions = []yarpc.CallOption{
 		yarpc.WithHeader(libraryVersionHeaderName, LibraryVersion),
 		yarpc.WithHeader(featureVersionHeaderName, FeatureVersion),
 		yarpc.WithHeader(clientImplHeaderName, clientImplHeaderValue),
 	}
 )
 
+func fromInternalFeatureFlags(featureFlags FeatureFlags) s.FeatureFlags {
+	return s.FeatureFlags{
```
demirkayaender, line 86: Remove client only header here

internal/internal_workflow_client.go
```diff
@@ -1026,7 +1028,9 @@ func (dc *domainClient) Describe(ctx context.Context, name string) (*s.DescribeD
 	var response *s.DescribeDomainResponse
 	err := backoff.Retry(ctx,
 		func() error {
-			tchCtx, cancel, opt := newChannelContext(ctx)
+
+			// TODO change here, opt should have the feature flags
```
demirkayaender, line 1032: remove here

demirkayaender (commented):
Talked to Emrah and added offline comments as PR comment, will address them

emrahs (approved):

PR #1111, opened 2021-07-09
Clear workflow state when not cached and not complete

Groxx:
This resolves a bug a user encountered, where a full sticky cache + a query for a non-cached workflow would result in a goroutine leak due to the associated event handler never being shut down. Since this leak retains all in-workflow data, it can eventually lead to an out-of-memory crash, though it does not cause any logic errors (these abandoned goroutines are forever idle once abandoned).  There are probably also other scenarios where this is possible.

Since the code paths leading to unlock are rather complex, I've included a DPanic log (panics in debug mode, logs at panic level elsewhere) and metric to check an assumption that I believe to be correct, but have not been able to be entirely confident about.  Once this has run in production for a while on a few domains, and verified to not happen, we can likely remove that log and metric.

---

Separately: this function seems to be far too complex, and is almost certainly duplicating checks made elsewhere, which should not be duplicated like this.  There have been multiple issues with state-clearing that have lead to adding conditions to this func, which is a clear sign of a code smell.

State / cache decisions like this should be made in exactly one place ever, and built up as obviously as possible, to ensure gaps like this never occur.  In this case we'll likely need to invert the dependency flow somehow, so callers control when cache is cleared based on whether or not it is cached, rather than double-checking internally like this.

We should also probably add go.uber.org/goleak to basically all of our tests, to help ensure we do not have goroutine leaks.  This may not have been caught by that, as the steps leading to it are a bit odd and rely on singleton config (sticky cache size), but it may find or prevent others.  (we should really rip out these globals too)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,11 +134,13 @@ bins: thriftc $(ALL_SRC) $(BUILD)/copyright lint $(BUILD)/dummy
 unit_test: $(BUILD)/dummy
 	@mkdir -p $(COVER_ROOT)
 	@echo "mode: atomic" > $(UT_COVER_FILE)
-	@for dir in $(UT_DIRS); do \
+	@failed=0; \
+	for dir in $(UT_DIRS); do \
 		mkdir -p $(COVER_ROOT)/"$$dir"; \
-		go test "$$dir" $(TEST_ARG) -coverprofile=$(COVER_ROOT)/"$$dir"/cover.out || exit 1; \
+		go test "$$dir" $(TEST_ARG) -coverprofile=$(COVER_ROOT)/"$$dir"/cover.out || failed=1; \
 		cat $(COVER_ROOT)/"$$dir"/cover.out | grep -v "mode: atomic" >> $(UT_COVER_FILE); \
-	done;
+	done; \
+	exit $$failed
 
 integ_test_sticky_off: $(BUILD)/dummy
 	@mkdir -p $(COVER_ROOT)
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -475,7 +475,8 @@ func (w *workflowExecutionContextImpl) Lock() {
 	w.mutex.Lock()
 }
 
-func (w *workflowExecutionContextImpl) Unlock(err error) {
+func (w *workflowExecutionContextImpl) Unlock(isInCache bool, err error) {
+	cleared := false
 	if err != nil || w.err != nil || w.isWorkflowCompleted || (w.wth.disableStickyExecution && !w.hasPendingLocalActivityWork()) {
 		// TODO: in case of closed, it assumes the close decision always succeed. need server side change to return
 		// error to indicate the close failure case. This should be rear case. For now, always remove the cache, and
@@ -484,8 +485,26 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
 			removeWorkflowContext(w.workflowInfo.WorkflowExecution.RunID)
 		} else {
 			// sticky is disabled, manually clear the workflow state.
+			if isInCache {
+				// sanity check, afaict this never happens.  but log/metric to collect evidence if it does.
+				// if it occurs it's almost certainly a logic bug (leads to double-clearing), though possibly not a damaging one.
+				// we should able to remove this after running for a while in canary / prod, our collection of users trigger quite a few edge cases.
+				w.wth.logger.DPanic("Workflow state should not have been cleared, it should be handled by cache eviction",
+					zap.String(tagWorkflowType, w.workflowInfo.WorkflowType.Name),
+					zap.String(tagWorkflowID, w.workflowInfo.WorkflowExecution.ID),
+					zap.String(tagRunID, w.workflowInfo.WorkflowExecution.RunID),
+					zap.Stack(tagPanicStack))
+				// temporary, remove after running in production for a while
+				w.wth.metricsScope.Counter("bug-clear-state")
+			}
 			w.clearState()
+			cleared = true
 		}
+	}
+	// !isInCache is the main scenario being caught here, but double-clearing is undesirable.
+	// it *might* be harmless to double-clear, but it's certainly irrational, and risks future problems.
+	if !isInCache && !cleared {
+		w.clearState()
 	}
 
 	w.mutex.Unlock()
@@ -648,7 +667,7 @@ func (wth *workflowTaskHandlerImpl) createWorkflowContext(task *s.PollForDecisio
 func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 	task *s.PollForDecisionTaskResponse,
 	historyIterator HistoryIterator,
-) (workflowContext *workflowExecutionContextImpl, err error) {
+) (workflowContext *workflowExecutionContextImpl, isInCache bool, err error) {
 	metricsScope := wth.metricsScope.GetTaggedScope(tagWorkflowType, task.WorkflowType.GetName())
 	defer func() {
 		if err == nil && workflowContext != nil && workflowContext.laTunnel == nil {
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -134,11 +134,13 @@ bins: thriftc $(ALL_SRC) $(BUILD)/copyright lint $(BUILD)/dummy
...
 		mkdir -p $(COVER_ROOT)/"$$dir"; \
-		go test "$$dir" $(TEST_ARG) -coverprofile=$(COVER_ROOT)/"$$dir"/cover.out || exit 1; \
+		go test "$$dir" $(TEST_ARG) -coverprofile=$(COVER_ROOT)/"$$dir"/cover.out || failed=1; \
 		cat $(COVER_ROOT)/"$$dir"/cover.out | grep -v "mode: atomic" >> $(UT_COVER_FILE); \
-	done;
+	done; \
+	exit $$failed
 
 integ_test_sticky_off: $(BUILD)/dummy
 	@mkdir -p $(COVER_ROOT)
```
Groxx, line 146: this runs all tests even if one fails, rather than stopping as soon as one package fails.  seems better?

internal/internal_task_handlers_test.go
```diff
@@ -416,7 +417,7 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_ActivityTaskScheduled() {
 	t.NotNil(response.Decisions[0].CompleteWorkflowExecutionDecisionAttributes)
 }
 
-func (t *TaskHandlersTestSuite) TestWorkflowTask_QueryWorkflow_Sticky() {
+func (t *TaskHandlersTestSuite) TestWorkflowTask_QueryWorkflow() {
 	// Schedule an activity and see if we complete workflow.
 	taskList := "sticky-tl"
```
Groxx, line 422: "sticky-tl" does nothing special whatsoever, which is the same as now nonsticky uses "tl" which does nothing special whatsoever.

I'm not sure what these tests were doing when they were written, but it feels like bad copypasta.

Groxx:
As a sample for when this fails, this new test fails when I comment out the new `clearState()` call:
```
=== RUN   TestTaskHandlersTestSuite/TestRegression_QueriesDoNotLeakGoroutines
2021-07-09T20:26:35.644-0700	INFO	internal/internal_worker.go:226	No metrics scope configured for cadence worker. Use NoopScope as default.
2021-07-09T20:26:35.644-0700	INFO	internal/internal_worker.go:230	No DataConverter configured for cadence worker. Use default one.
    internal_task_handlers_test.go:1553: 
        	Error Trace:	internal_task_handlers_test.go:1553
        	Error:      	Not equal: 
        	            	expected: 17
        	            	actual  : 18
        	Test:       	TestTaskHandlersTestSuite/TestRegression_QueriesDoNotLeakGoroutines
        	Messages:   	expected the query to leak no new goroutines.  before query:
        	            	found unexpected goroutines:
        	            	[Goroutine 179 in state chan receive, with go.uber.org/cadence/internal.(*coroutineState).initialYield on top of the stack:
        	            	goroutine 179 [chan receive]:
        	            	go.uber.org/cadence/internal.(*coroutineState).initialYield(0xc000450ac0, 0x3, 0xc000624000, 0x22)
        	            		/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:795 +0xe5
        	            	go.uber.org/cadence/internal.(*coroutineState).yield(0xc000450ac0, 0xc000624000, 0x22)
        	            		/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:805 +0xb0
        	            	go.uber.org/cadence/internal.(*channelImpl).Receive(0xc00044e240, 0x23088a8, 0xc000130480, 0x0, 0x0, 0xc000077150)
        	            		/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:620 +0x4fb
        	            	go.uber.org/cadence/internal.TestPanic.func1(0x23088a8, 0xc000130480)
        	            		/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_coroutines_test.go:681 +0x39a
        	            	go.uber.org/cadence/internal.(*dispatcherImpl).newNamedCoroutine.func1(0x29243c8, 0x1, 0xc000077130, 0x23088a8, 0xc000130480, 0xc000450ac0)
        	            		/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:883 +0x14d
        	            	created by go.uber.org/cadence/internal.(*dispatcherImpl).newNamedCoroutine
        	            		/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:874 +0x196
        	            	
        	            	 Goroutine 249 in state select, with go.uber.org/cadence/internal.(*cadenceInvoker).heartbeatAndScheduleNextRun.func1 on top of the stack:
        	            	 ......
        	            	 you get the idea
        	            	 ......
    internal_task_handlers_test.go:1487: failed to clean up goroutines.
        Original state:
        found unexpected goroutines:
        [Goroutine 179 in state chan receive, with go.uber.org/cadence/internal.(*coroutineState).initialYield on top of the stack:
        goroutine 179 [chan receive]:
        go.uber.org/cadence/internal.(*coroutineState).initialYield(0xc000450ac0, 0x3, 0xc000624000, 0x22)
        	/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:795 +0xe5
        go.uber.org/cadence/internal.(*coroutineState).yield(0xc000450ac0, 0xc000624000, 0x22)
        	/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:805 +0xb0
        go.uber.org/cadence/internal.(*channelImpl).Receive(0xc00044e240, 0x23088a8, 0xc000130480, 0x0, 0x0, 0xc000077150)
        	/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:620 +0x4fb
        go.uber.org/cadence/internal.TestPanic.func1(0x23088a8, 0xc000130480)
        	/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_coroutines_test.go:681 +0x39a
        go.uber.org/cadence/internal.(*dispatcherImpl).newNamedCoroutine.func1(0x29243c8, 0x1, 0xc000077130, 0x23088a8, 0xc000130480, 0xc000450ac0)
        	/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:883 +0x14d
        created by go.uber.org/cadence/internal.(*dispatcherImpl).newNamedCoroutine
        	/Users/stevenl/gocode/src/go.uber.org/cadence/internal/internal_workflow.go:874 +0x196
        
        ... like before, lots of other stacks
```
Unfortunately, because there's no structure (these are just error strings), there's no quick way to filter this list down to just the new stack trace.
I'm not too concerned though - if we were not leaking goroutines elsewhere, this would be a list of 1 turning into a list of 2, which is trivial to check by hand.

internal/internal_task_handlers.go
```diff
@@ -484,9 +485,27 @@ func (w *workflowExecutionContextImpl) Unlock(err error) {
...
+				// temporary, remove after running in production for a while
+				w.wth.metricsScope.Counter("bug-clear-state")
+			}
 			w.clearState()
+			cleared = true
 		}
 	}
+	// !isInCache is the main scenario being caught here, but double-clearing is undesirable.
+	// it *might* be harmless to double-clear, but it's certainly irrational, and risks future problems.
+	if !isInCache && !cleared {
```
vytautas-karpavicius, line 506: Is there any reason why we are not checking a cache here directly?

Seems like it would be more straightforward to check if workflow is cached here and if so clear the state. Passing `isInCache` around add additional complexity. Or am I missing something here?

Groxx: Probably fine - I'll poke at it a bit more.  some of this is I think just leftovers from before I had a better feel for how the cache was being used / how pinning behaved.

Groxx: updated 👍 .  this is better, yeah

vytautas-karpavicius (approved):
This looks good now.

PR #1116, opened 2021-08-02
Implementation of JWTAuthorizationProvider 

noiarek:
**What changed?**
Implementation of `JWTAuthorizationProvider` in both worker and client

**Why?**
Part of [this](https://docs.google.com/document/d/1a-Xo9FSCLubYlIQHB_eX5j4Yv2GDv_82J94At2dAL3o/edit#) proposal

**How did you test it?**
Running the server with oauth file

```
./cadence-server --env development_oauth start
```

and modifying [cadence-samples](https://github.com/uber-common/cadence-samples)

Adding Authorization value in NewClient [here](https://github.com/uber-common/cadence-samples/blob/master/cmd/samples/common/factory.go#L89)

Adding Authorization value in NewDomainClient [here](https://github.com/uber-common/cadence-samples/blob/master/cmd/samples/common/factory.go#L89)

Adding Authorization value in worker.Options [here](https://github.com/uber-common/cadence-samples/blob/master/cmd/samples/recipes/helloworld/main.go#L18)

**Potential risks**
None


internal/common/auth/service_wrapper.go
```diff
@@ -0,0 +1,206 @@
...
+        authProvider: authorizationProvider,
+    }
+}
+
+func (w *workflowServiceAuthWrapper) getYarpcJWTHeader() yarpc.CallOption {
+    token := w.authProvider.GetAuthToken()
+    return yarpc.WithHeader(jwtHeaderName, string(token))
+}
+
+func (w *workflowServiceAuthWrapper) DeprecateDomain(ctx context.Context, request *shared.DeprecateDomainRequest, opts ...yarpc.CallOption) error {
```
noiarek, line 53: Hey @longquanzheng , it seems in the metrics implementation of workflowserviceclient, a scope is provided (code [here](https://github.com/uber-go/cadence-client/blob/master/internal/common/metrics/service_wrapper.go#L93)) and the scope is used at every function to handle the possible error (see [here](https://github.com/uber-go/cadence-client/blob/master/internal/common/metrics/service_wrapper.go#L136)). 

The implementation you recommended didn't include a scope

```
service = metrics.NewWorkflowServiceWrapper(service, metricScope)
if authProvider != nil{
    service = auth.NewWorkflowServiceWrapper(service, authProvider)
}
```

So I'm wondering if:

* Either we don't do anything with the scope and we just return the error like this implementation
* We use the same scope as metric
* We create a new one and the new function signate would look like `auth.NewWorkflowServiceWrapper(service, scope, authProvider)`

Also let me know if line 54 looks correct to you

Thanks!

longquanzheng: It looks good. This wrapper will wrap the wrapper of service metrics so you don’t need to do it again.

longquanzheng: Actually, we should do this instead:
```
if authProvider != nil{
    service = auth.NewWorkflowServiceWrapper(service, authProvider)
}
service = metrics.NewWorkflowServiceWrapper(service, metricScope)
```
Just a note that this is not overriding the service wrapper with metrics. It takes into a previous one and make another wrapper so the metrics scope is still being used in the previous one. 
Reverting the order here so that the second service warper can capture the erorr of the previous one and emit error metrics. 

Does it make sense?

noiarek: Yeah it makes sense, I need to check the parameters because it's the second time I ask something like this :) Thanks @longquanzheng !

longquanzheng: Thank you for pointing it out too :D

internal/common/auth/service_wrapper.go
```diff
@@ -0,0 +1,206 @@
...
+import (
+    "context"
+    "go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
+    "go.uber.org/cadence/.gen/go/shared"
+    "go.uber.org/cadence/worker"
+    "go.uber.org/yarpc"
+)
+
+const (
+    jwtHeaderName = "cadence-client-jwt"
```
longquanzheng, line 32: I think this should be exactly the same as the server: `"cadence-authorization"`

internal/internal_worker.go
```diff
@@ -193,6 +195,15 @@ type (
 		// flags to turn on/off some server side features
 		FeatureFlags FeatureFlags
 	}
+
+	JWTClaims struct {
```
noiarek, line 199: Is this the right place to place this struct @longquanzheng ? Or do you have a file where you share common structs between server and client?

worker/worker.go
```diff
@@ -173,6 +173,12 @@ type (
 		Run() error
 	}
 
+	AuthorizationProvider interface {
```
noiarek, line 176: This is giving me a cyclic error @longquanzheng.

If I would like to place this on `"go.uber.org/cadence/.gen/go/shared"` but I'm wondering if everything inside of `.gen` is generated and if that's the case how could I generate that struct there? Or where do you recommend me to place it?

longquanzheng: Yeah don’t put it under .gen. There should be other places to put it . Would it work if you put it under auth/ ? And then use type aliases to export to public

internal/internal_worker.go
longquanzheng: This is not really going to be shared. I think we just need to make sure the structure is the same as server by contract. I would like to put it under auth/ or new pkg under it if any cycle dependency happens. Because this is an option feature and having things centralized in this package would help us to maintaining the code base. This internal package is already sorta exploding.

internal/jwt_authorization.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+
+func NewJwtAuthorizationProvider(privateKey []byte) auth.AuthorizationProvider {
+    return &JWTAuthProvider{
+        PrivateKey: privateKey,
+    }
+}
+
+func (j *JWTAuthProvider) GetAuthToken() ([]byte, error) {
```
noiarek, line 41: @longquanzheng had to change the function signature from `[]byte` to `([]byte, error)` because as you can see, it can return an error a couple of times. is that ok?

longquanzheng: Yeah sounds good. That makes sense to me

internal/internal_worker.go
```diff
@@ -29,6 +29,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"go.uber.org/cadence/internal/common/auth"
```
longquanzheng, line 32: This should be moved to the below section/group

internal/worker.go
```diff
@@ -225,6 +226,10 @@ type (
 		// Optional: Flags to turn on/off some server side options
 		// default: all the features in the struct are turned off
 		FeatureFlags FeatureFlags
+
+		// Optional: Authorization interface to get the Auth Token
+		// default: empty token
```
longquanzheng, line 231: Default should be no provider

longquanzheng (commented):
Over all looks great 👍

longquanzheng (approved):
LGTM, it would be better to have some unit tests.

noiarek:
@longquanzheng test added, check `internal/common/auth/service_wrapper_test.go` and `internal/jwt_authorization_test.go`

Issue #1045, opened 2020-12-09
Non deterministic error for childWorkflow ID after reset


longquanzheng:
In go client when starting a child workflow, if workflow ID is not specified,  by default the childWF ID is determined by the parent workflow's runID.

See code:
https://github.com/uber-go/cadence-client/blob/7c70757e7c7fab5e3f7fc8bc4a9c5a3acef01a8b/internal/internal_event_handlers.go#L377

And reset will terminate the current run and start a new run.
So after the runID change after reset, the default childWorkflowID is also changed. Therefore it's non deterministic.

Ideally childWorkflow shouldn’t be dependent on parent runID. Java client uses a different way so it doesn’t have the issue. Go client probably should follow the same approach. 

### Prevention is always better 
To avoid running into this, please always provide a stable workflowID when starting childWorkflows. 

* Please use [SideEffect](https://cadenceworkflow.io/docs/go-client/side-effect/) to generate the UUID if you don't have one from business logic. Example:
```golang 
var childID string 
workflow.SideEffect(ctx, func(ctx workflow.Context) interface{} {
		return uuid.New().String()
	}).Get(&childID)
cwo := workflow.ChildWorkflowOptions{
	WorkflowID:                   childID,
	ExecutionStartToCloseTimeout: time.Minute,
}
ctx = workflow.WithChildOptions(ctx, cwo)
err := workflow.ExecuteChildWorkflow(ctx, XyzChildWorkflow,...)
```

### Mitigation
If you already started childWorkflows without specifying workflowIDs
The only way to use reset with childWorkflow is that you prepare a mapping between parent workflow IDs to childWorkflowIDs

 * First of all, update the workflow code to use stable workflowID for childWorkflow for new executing only, using [Versioning](https://cadenceworkflow.io/docs/go-client/workflow-versioning/). Example code:
```golang
var childID string 
v := workflow.GetVersion(ctx, "startChildWfXyzForReset", workflow.DefaultVersion, 1)
if v == workflow.DefaultVersion {
    // when workflow executed this code with DefaultVersion, it was using empty childWF ID, let's keep using empty childWorkflowID like before
   childID = ""
}else{
   // when workflow executed this code with Version 1, let's use a stable workflowID for child so that we won't run into this issue. 
   workflow.SideEffect(ctx, func(ctx workflow.Context) interface{} {
		return uuid.New().String()
	}).Get(&childID)
}

cwo := workflow.ChildWorkflowOptions{
[...]

PR #1118, opened 2021-08-14
Fix childWF decision panic after reset due to runID change

longquanzheng:
**What changed?**
Use originalRunID to fix childWF decision panic after reset due to runID change

**Why?**
For https://github.com/uber-go/cadence-client/issues/1045
Use originalRunID is the right fix for this, and is backward-compatible. 
https://github.com/uber/cadence-idl/blob/48a14966426238c15570a13deab5a7905c0a89c1/thrift/shared.thrift#L556

See more discussion in https://uber-cadence.slack.com/archives/CLS1SM941/p1607515148008700 
There are a couple of users running into this issues. 

**How did you test it?**
Asking a user to test it. 

**Potential risks**
No, it should be backward compatible. Without reset, orignalRunID should be the same as runID.


yycptt (approved):
👍 Would you mind creating a task for adding a canary test for this case?

PR #1121, opened 2021-08-17
Use the same timer duration validation with server

demirkayaender:
**What changed?**
Server performs timer duration validation against duration in seconds. However, client performs the same checks against duration type. Although, the client side checks look safe enough, we had a recent incident where server received invalid timer parameters. Although it's unlikely, this validation discrepancy probably happened due to the float to int conversion discrepancies. Therefore, we are applying the same type of validation in client to eliminate that possibility. 

**Why?**

**How did you test it?**

**Potential risks**


just-at-uber (approved):

yycptt (approved):

mkolodezny (approved):

mkolodezny (commented):
do we need this for java sdk as well?

demirkayaender:
@mkolodezny Good catch! I checked java client and it seems to have it already: https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/sync/SyncDecisionContext.java#L494-L515

PR #1123, opened 2021-08-26
Break long poll loop early when context remaining timeout is too low

vytautas-karpavicius:
**What changed?**
Stop long poll on the client side when too little time is left on the context.

**Why?**
To prevent unnecessary call to the server will would timeout anyway.

Relate Server PR: https://github.com/uber/cadence/pull/4425

**How did you test it?**
Locally.

**Potential risks**

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -49,6 +49,7 @@ var _ DomainClient = (*domainClient)(nil)
 const (
 	defaultDecisionTaskTimeoutInSecs = 10
 	defaultGetHistoryTimeoutInSecs   = 25
+	minLongPollInterval              = 100 * time.Millisecond
 )
 
 var (
@@ -512,6 +513,15 @@ func (wc *workflowClient) GetWorkflowHistory(
 		var err error
 	Loop:
 		for {
+			// Stop polling the server if the remaining context timeout is too low.
+			// This prevents unnessesary final call which will result in a timeout anyway.
+			if deadline, ok := ctx.Deadline(); ok {
+				remainingTime := deadline.Sub(time.Now())
+				if remainingTime < minLongPollInterval {
+					return nil, context.DeadlineExceeded
+				}
+			}
+
 			err = backoff.Retry(ctx,
 				func() error {
 					var err1 error


zawadzkidiana (commented):
will follow up shortly!

dianazawadzki:
> Thanks for this, @vytautas-karpavicius, and apologies for the years-long silence. The diagnosis is real — sending a long-poll GetWorkflowExecutionHistory call when the caller's context is about to expire anyway just burns a round trip that's guaranteed to time out.
> 
> Reviewing this against current `master`, though, I believe the underlying problem has since been solved a different way in the same loop (`GetWorkflowHistoryWithOptions`, the successor to the `GetWorkflowHistory` this diff patches):
> 
> ```go
> tchCtx, cancel, opt := newChannelContext(ctx, wc.featureFlags, func(builder *contextBuilder) {
>     if request.IsLongPoll {
>         builder.Timeout = defaultGetHistoryTimeoutInSecs * time.Second
>         deadline, ok := ctx.Deadline()
>         if ok && deadline.Before(time.Now().Add(builder.Timeout)) {
>             // insufficient time for another poll, so this needs to be the last attempt
>             isFinalLongPoll = true
>         }
>     }
> })
> ...
> if isFinalLongPoll {
>     // essentially a deadline exceeded, the last attempt did not get a result.
>     return nil, fmt.Errorf("timed out waiting for the workflow to finish: %w", context.DeadlineExceeded)
> }
> ```
> 
> That's arguably a bit more precise than this PR's fixed `minLongPollInterval = 100ms` guard: instead of skipping the final attempt outright once < 100ms remains, it still lets the last poll happen (with whatever time is actually left) but knows not to retry afterward, and it wraps `context.DeadlineExceeded` into the error it already returns rather than special-casing an early return before the retry loop.
> 
> Given that, I think this PR is superseded and can probably be closed — but flagging in case there's an edge case the `isFinalLongPoll` check doesn't cover that motivated the original 100ms threshold specifically (would be good to hear if you still saw the "call that's certain to time out" behavior after that logic landed).
> 
> A couple of smaller notes for the record, in case any of this gets revisited:
> 
> * Typo: "unnessesary" → "unnecessary" in the comment.
> * `minLongPollInterval` reads as a polling _interval_, but it's actually being used as a minimum-remaining-context-time buffer — a name like `minRemainingTimeForPoll` would be less ambiguous.
> * No test was added for the new early-return branch, which would've been worth having given it's a change to a hot retry/timeout path.
> 
> Thanks again for tracking this down and for the reference to the paired server-side PR — sorry the review loop here took this long.
> 

Thanks Claude, will review this myself now!
-Diana

PR #1131, opened 2021-09-24
Filling out Selector docs, and some adjacent ones I noticed

Groxx:
Recently I learned that Selector was basically undocumented, so I wanted to get around to improving that.
And then @longquanzheng beat me to it!  #685 is also an old PR that's related, though I didn't build off of that.
But here's a bunch more text for people to read.

No doubt there are more docs to improve, but hopefully this is a step in the right direction.

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -46,92 +46,229 @@ var (
 )
 
 type (
-	// Channel must be used instead of native go channel by workflow code.
-	// Use workflow.NewChannel(ctx) method to create Channel instance.
+	// Channel must be used in workflows instead of a native Go chan.
+	//
+	// Use workflow.NewChannel(ctx) to create an unbuffered Channel instance,
+	// workflow.NewBufferedChannel(ctx, size) to create a Channel which has a buffer,
+	// or workflow.GetSignalChannel(ctx, "name") to get a Channel that can contain encoded data sent from other systems.
+	//
+	// workflow.GetSignalChannel is named differently because you are not "creating" a new channel.  Signal channels
+	// are conceptually singletons that exist at all times, and they do not have to be "created" before a signal can be
+	// sent to a workflow.  The workflow will just have no way to know that the data exists until it inspects the
+	// appropriate signal channel.
+	//
+	// Both NewChannel and NewBufferedChannel have "Named" constructors as well.
+	// These names will be visible in stack-trace queries, so they can help with debugging, but they do not otherwise
+	// impact behavior at all, and are not recorded anywhere (so you can change them without versioning your code).
 	Channel interface {
 		// Receive blocks until it receives a value, and then assigns the received value to the provided pointer.
-		// Returns false when Channel is closed.
-		// Parameter valuePtr is a pointer to the expected data structure to be received. For example:
-		//  var v string
-		//  c.Receive(ctx, &v)
+		// It returns false when Channel is closed and all data has already been consumed from the channel, in the same
+		// way as Go channel reads work.
+		//
+		// This is equivalent to `v, more := <- aChannel`.
+		//
+		// valuePtr must be assignable, and will be used to assign (for in-memory data in regular channels) or decode
+		// (for signal channels) the data in the channel.
+		//
+		// If decoding or assigning fails:
+		// - an error will be logged
+		// - the value will be dropped from the channel
+		// - Receive will automatically try again
+		// - This will continue until a successful value is found, or the channel is emptied and it resumes blocking.
+		//   Closed channels with no values will always succeed, but they will not change valuePtr.
+		//
+		// Go would normally prevent incorrect-type failures like this at compile time, but the same cannot be done
+		// here.  If you need to "try" to assign to multiple things, similar to a Future you can use:
+		// - for signal channels, a []byte pointer.  This will give you the raw data that Cadence received, and no
+		//   decoding will be attempted, so you can try it yourself.
+		// - for other channels, an interface{} pointer.  All values are interfaces, so this will never fail, and you
+		//   can inspect the type with reflection or type assertions.
 		Receive(ctx Context, valuePtr interface{}) (more bool)
 
-		// ReceiveAsync try to receive from Channel without blocking. If there is data available from the Channel, it
-		// assign the data to valuePtr and returns true. Otherwise, it returns false immediately.
+		// ReceiveAsync tries to receive from Channel without blocking.
+		// If there is data available from the Channel, it assigns the data to valuePtr and returns true.
+		// Otherwise, it returns false immediately.
+		//
+		// This is equivalent to:
+		//  select {
+		//  case v := <- aChannel: ok = true
+		//  default: ok = false
+		//  }
+		//
+		// Decoding or assigning failures are handled like Receive.
 		ReceiveAsync(valuePtr interface{}) (ok bool)
 
-		// ReceiveAsyncWithMoreFlag is same as ReceiveAsync with extra return value more to indicate if there could be
-		// more value from the Channel. The more is false when Channel is closed.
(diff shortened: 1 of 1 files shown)


internal/workflow.go
```diff
@@ -46,92 +46,218 @@ var (
...
+		// This behaves the same as:
+		//  select {
+		//  case aChannel <- v: ok = true
+		//  default: ok = false
+		//  }
 		SendAsync(v interface{}) (ok bool)
 
-		// Close close the Channel, and prohibit subsequent sends.
+		// Close closes the Channel, and prohibits subsequent sends.
+		// As with a normal Go channel that has been closed, sending to a closed channel will panic.
```
Groxx, line 119: Closing a _signal_ channel in tests works... but sending to it causes a panic in the testing stack, which wouldn't happen in a real workflow.
So I haven't documented how closed signal channels behave yet, because I don't really know for sure.

internal/workflow.go
```diff
@@ -46,92 +46,218 @@ var (
...
+		// (for signal channels) the data in the channel.
+		//
+		// If decoding or assigning fails:
+		// - an error will be logged
+		// - the value will be dropped from the channel
+		// - Receive will automatically try again
+		// - This will continue until a successful value is found, or the channel is emptied and it resumes blocking.
+		//   Closed channels with no values will always succeed, but they will not change valuePtr.
+		//
+		// Go would normally prevent this at compile time, but the same cannot be done here.
```
Groxx, line 75: note to self: "prevent incorrect-type failures like this"

Shaddoll (approved):

mkolodezny (approved):
looks great, thanks!

internal/workflow.go
```diff
@@ -46,92 +46,229 @@ var (
 )
 
 type (
-	// Channel must be used instead of native go channel by workflow code.
-	// Use workflow.NewChannel(ctx) method to create Channel instance.
+	// Channel must be used in workflows instead of a native Go chan.
+	//
+	// Use workflow.NewChannel(ctx) to create an unbuffered Channel instance,
+	// workflow.NewBufferedChannel(ctx, size) to create a Channel which has a buffer,
+	// or workflow.GetSignalChannel(ctx, "name") to get a Channel that can contain encoded data sent from other systems.
```
longquanzheng, line 53: `encoded data sent from other systems.` is a little confusing to me. 

I think we just need to mention this is a special channel that maintained/created by Cadence just for workflow signal feature.

internal/workflow.go
```diff
@@ -46,92 +46,229 @@ var (
...
+	//  		}
+	//  	})
+	//  	s.AddFuture(workflow.NewTimer(ctx, time.Hour), func(f workflow.Future) {
+	//  		timedout = true
+	//  	})
+	//  	s.Select(ctx)
+	//  	if timedout {
+	//  		break
+	//  	}
+	//  }
```
longquanzheng, line 186: Q: 
I guess there is no difference, but do you have any idea/preference? 

If we only keep the `s.Select(ctx)` in the for loop:
```golang
timedout := false
s := workflow.NewSelector(ctx)
s.AddReceive(chA, func(c workflow.Channel, more bool) {
	if more {
		var i int
		c.Receive(ctx, &i)
		counter += i
	}
})
s.AddReceive(chB, func(c workflow.Channel, more bool) {
	if more {
		var i int
		c.Receive(ctx, &i)
		counter += i
	}
})
s.AddFuture(workflow.NewTimer(ctx, time.Hour), func(f workflow.Future) {
	timedout = true
})

for {
  s.Select(ctx)
  if timedout {
	  break
  }
}
```

internal/workflow.go
Groxx: yeah, I can reword this.
the main difference is that the internal channels _do not_ encode their data, so they can pass non-serializable data around.  it's a somewhat significant difference, IMO it needs to be covered somewhere.

but it doesn't have to be here / in this way :)  lemme think about it / I'm open to suggestions!

internal/workflow.go
```diff
@@ -46,92 +46,229 @@ var (
...
+	//
+	// Finally, note that Select will not return until a condition's needs are met, like a Go selector - canceling the
+	// Context used to construct the Selector, or the Context used to Select, will not (directly) unblock a Select call.
+	// Read Select for more details.
 	Selector interface {
-		// AddReceive adds a ReceiveChannel to the selector. f is invoked when the channel has data or closed.
-		// ok == false indicates the channel is closed
-		AddReceive(c Channel, f func(c Channel, ok bool)) Selector
-		// AddSend adds a SendChannel to the selector. f is invoke when the channel is available to send
+		// AddReceive waits to until a value can be received from a channel.
```
longquanzheng, line 198: maybe 
`AddReceive adds a callback f, to wait and invoke f until a value can be received from a channel`

internal/workflow.go
Groxx: for this one there is a difference - in the original with `case <-time.After`, the timer is restarted on each loop.  the same thing happens in the translated version.

I could pull the timer out separately (`after := time.After(time.Hour)` + `case <-after`).  semantically they're a bit different (one waits for N things within 1 hour, the other waits for N things _each_ hour), no real preference which we demonstrate.
but I do want a timeout of some kind in one of these docs chunks, since it's not the most obvious part of the API, and timeouts are kinda common.

internal/workflow.go
longquanzheng: actually, I don't like mine(too hard to read). So feel free to keep it.

Groxx: looks like I did screw up the text here in any case.  will fix 👍

internal/workflow.go
```diff
@@ -46,92 +46,229 @@ var (
...
+		//   call, and then wait for other Selector conditions in the second Select call (or block forever if there are
+		//   no other eligible conditions).
+		// - Adding the same Future twice, then calling Select twice, will invoke each callback once.
+		// - Adding the same Future to two different Selectors, then calling Select once on each Selector, will invoke
+		//   each Selector's callback once.
+		//
+		// Therefore, with a Future "f" that is or will become ready, this is an infinite loop that will consume as much
+		// CPU as possible:
+		//  for {
+		//  	workflow.NewSelector(ctx).AddFuture(f, func(f workflow.Future){}).Select(ctx)
```
longquanzheng, line 234: TIL! 

Looks like we should encourage putting only `Select(ctx)` into forloop?

internal/workflow.go
longquanzheng: yeah I think just `other systems` bugs me a little. Without context I have no clue what is `other` mean here. I think just say `contain encoded data sent by workflow signal` should be good for me.

Good point about the data encoding. May worth mentioning other channels are `workflow internal channels so that they can pass in/out variables that are not serializable`

internal/workflow.go
Groxx: I dunno tbh.  with simple demos like this, yeah - I think it's somewhat safer to construct it once and select multiple times.

But _mutating the select inside a callback_ is somewhat common (e.g. starting a new activity and waiting on it too), and from what I've seen people have fewer problems with that kind of setup (and with complex selects in general) when they rebuild it from scratch each time.  
it forces them to track state outside the loop earlier on, which is a good thing - they can't _remove_ those futures once they've been handled, so they are fairly often eventually forced to rebuild it in the loop, and that's not always a simple change.

internal/workflow.go
Groxx: on second (or third or whatever) thought: I should probably add a comment about intended behavior anyway.  no need to leave people to puzzle it out (timeout vs idle-timeout) while reading about more relevant things.

internal/workflow.go
Groxx: another way to look at it I guess:
- one way risks an infinite loop
- one way risks a deadlock

I don't really think there's a "best" option there - regardless of which you choose, you _must_ do it right.

PR #1134, opened 2021-09-28
New section for canceled-context in workflows, other minor cleanups

Groxx:
Canceled-context documentation is currently strewn about in individual function documentation, though it's a fairly important concept that needs to be understood.

In particular, users sometimes are under the impression that "all things which accept a context will error", which can lead to nasty surprises like `selector.Select(ctx)`'s blocking behavior (causing a deadlocked workflow).  Selector's docs have been improved in #1131, and this will hopefully make its behavior easier to learn before it is needed.

And, since I noticed some small other things while writing this chunk of docs, I've rolled those changes into here as well.


workflow/doc.go
```diff
@@ -139,8 +139,79 @@ Time related functions:
...
+As an incomplete summary, these actions will all fail immediately (or when the activity completes, if WaitForCancellation
+is true), and the associated error returns (possibly within a Future) will be a workflow.CanceledError:
+
+  - workflow.ExecuteActivity and workflow.ExecuteLocalActivity
+    (see Activity's documentation for more details on canceled activities)
+  - workflow.ExecuteChildWorkflow
+  - workflow.Await
+  - workflow.Sleep
+  - workflow.Timer
+
```
Groxx, line 201: signaling and RequestCancelExternalWorkflow both do not mention CanceledError... but they seem like things that should probably no-op on cancel.  anyone know what they do?

workflow/doc.go
```diff
@@ -286,14 +357,14 @@ pattern, extra care needs to be taken to ensure the child workflow is started be
...
-activity returns an error as error.NewCustomError("err-reason", details), that error will be converted to
-*error.CustomError. There are other types of errors like error.TimeoutError, error.CanceledError and error.PanicError.
+activity returns an error as errors.New() or fmt.Errorf(), those errors will be converted to workflow.GenericError. If the
+activity returns an error as workflow.NewCustomError("err-reason", details), that error will be converted to
+*workflow.CustomError. There are other types of errors like workflow.TimeoutError, workflow.CanceledError and workflow.PanicError.
 So the error handling code would look like:
 
 	err := workflow.ExecuteActivity(ctx, YourActivityFunc).Get(ctx, nil)
 	switch err := err.(type) {
-	case *error.CustomError:
```
Groxx, line 296: we have no `error` package ¯\\\_(ツ)\_/¯

activity/doc.go
```diff
@@ -112,10 +118,13 @@ payload containing progress information.
 Activity Cancellation
 
 When an activity is cancelled (or its workflow execution is completed or failed) the context passed into its function
-is cancelled which sets its Done channel’s closed state. So an activity can use that to perform any necessary cleanup
-and abort its execution. Currently cancellation is delivered only to activities that call RecordHeartbeat.
+is cancelled which closes its Done() channel. So an activity can use that to perform any necessary cleanup
+and abort its execution.
 
-Async/Manual Activity Completion
```
Groxx, line 118: `/` prevents godoc from interpreting it as a header

mkolodezny (approved):

workflow/doc.go
```diff
@@ -139,8 +139,79 @@ Time related functions:
...
+some behavior be shut down, while allowing you to run cleanup logic in activities or elsewhere.
+
+As a general guideline, doing anything with I/O with a canceled context (e.g. executing an activity, starting a
+child workflow, sleeping) will fail rather than cause external changes.  Detailed descriptions are available in
+documentation on functions that change their behavior with a canceled context; if it does not mention canceled-context
+behavior, its behavior does not change.
+For exact behavior, make sure to read the documentation on functions that you are calling.
+
+As an incomplete summary, these actions will all fail immediately (or when the activity completes, if WaitForCancellation
+is true), and the associated error returns (possibly within a Future) will be a workflow.CanceledError:
```
longquanzheng, line 193: `or when the activity completes, if WaitForCancellation is true`

Is this the default behavior? I tested this before seems like so. Maybe we should remove activity from this summary and make it as a special case.

workflow/doc.go
longquanzheng: even if we have, I think using `workflow.XyzError is more friendly than `error.XyzError` because there will be a lot of `error` pkg loaded into the IDE.

workflow/doc.go
Groxx: default behavior is false, so it does not wait.

are you thinking to just drop the exception note (or say "read executeactivity for more details"), or put it in its own list since it (and maybe executechildworkflow?  I'm not sure how it completes either) is kinda unique?

workflow/doc.go
Groxx: yeah, "error" and "errors" are kinda risky in that way.  lots of chances for collisions -> annoying manual aliases.

workflow/doc.go
longquanzheng: > default behavior is false, so it does not wait.
Interesting, looks like Java client is different. I tested in 
https://github.com/uber/cadence-java-samples/pull/47
that it will wait for completion. 

also https://github.com/uber/cadence-java-samples/pull/48 for long running activity activity will get canceled right way when reporting heartbeat. 

Yeah I believe pulling activity out is better to read. if having a separate paragraph.  It has different cases than other. Using `()` is friendly to understand.

longquanzheng (approved):
some minor comments, otherwise LGTM

PR #1135, opened 2021-09-29
Add a random UUID to worker identities, to prevent collisions

Groxx:
The fallback worker identity of `{pid}@{host}@{tasklist}` is insufficient in practice.
E.g. within Uber, our docker setup can somewhat regularly lead to multiple instances
of a worker ending up on the same physical host (which is not changed by docker) and
with the same PID (as all the docker containers start up the same way -> all service
processes get the same PIDs).
Other causes are possible too, e.g. if a worker crashes and is restarted it could
share the same host+pid, even though its caches (and anything else we actually care
about for the identity) are lost.

At the least-problematic-but-confusing level, collisions can lead to a misleadingly-short
list of workers on tasklists in the web UI, as the list does not show duplicates.

At the most-problematic level, this can lead to uneven load and sub-par request
routing, as only the first/last/??? request to poll the server will receive data
intended for a specific worker.

Ideally that wouldn't be an issue, e.g. the server could keep better track of polls
and route more precisely, but 1) it's difficult to reliably identify polls since
the server only has the poller-provided data (which mis-identifies itself), and 2)
even if it can be fixed on the server, improving identity uniqueness helps both fix
this now and reduces or eliminates the impact of any future identity confusions.

---

Alternatively, we could add a process-global random UUID that auto-inits, which
would give us IDs that can be correlated across tasklists.  While I think this is
acceptable as well, I'm loathe to add another global variable, and the benefit of
cross-task identity correlating is pretty minor or nonexistent.

If someone actually cares about that, they can pass an explicit ID that they
generate however they like.

### DIFF
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -32,6 +32,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/pborman/uuid"
 	"github.com/uber-go/tally"
 	s "go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/internal/common"
@@ -196,8 +197,12 @@ func newChannelContextHelper(
 }
 
 // GetWorkerIdentity gets a default identity for the worker.
+//
+// This contains a random UUID, generated each time it is called, to prevent identity collisions when workers share
+// other host/pid/etc information.  These alone are not guaranteed to be unique, especially when Docker is involved.
+// Take care to retrieve this only once per worker.
 func getWorkerIdentity(tasklistName string) string {
-	return fmt.Sprintf("%d@%s@%s", os.Getpid(), getHostName(), tasklistName)
+	return fmt.Sprintf("%d@%s@%s@%s", os.Getpid(), getHostName(), tasklistName, uuid.New())
 }
 
 func getHostName() string {


internal/internal_utils.go
```diff
@@ -196,8 +197,12 @@ func newChannelContextHelper(
 }
 
 // GetWorkerIdentity gets a default identity for the worker.
+//
+// This contains a random UUID, generated each time it is called, to prevent identity collisions when workers share
+// other host/pid/etc information.  These alone are not guaranteed to be unique, especially when Docker is involved.
+// Take care to retrieve this only once per worker.
 func getWorkerIdentity(tasklistName string) string {
-	return fmt.Sprintf("%d@%s@%s", os.Getpid(), getHostName(), tasklistName)
+	return fmt.Sprintf("%d@%s@%s@%s", os.Getpid(), getHostName(), tasklistName, uuid.New())
```
Groxx, line 205: Only 3 callers exist currently, all of them are a single call when initializing a client or worker:

https://github.com/uber-go/cadence-client/blob/7c70757e7c7fab5e3f7fc8bc4a9c5a3acef01a8b/internal/client.go#L518-L521

https://github.com/uber-go/cadence-client/blob/7c70757e7c7fab5e3f7fc8bc4a9c5a3acef01a8b/internal/client.go#L565-L568

https://github.com/uber-go/cadence-client/blob/7c70757e7c7fab5e3f7fc8bc4a9c5a3acef01a8b/internal/internal_worker.go#L211-L213
(this one is called in several locations, e.g. `newActivityWorker`, `newAggregatedWorker`, `newSessionWorker`, `newWorkflowTaskWorkerInternal`)

Groxx: Starting it up locally does show both workers on the same UUID, in any case:
Screen Shot 2021-09-28 at 19 08 28

vytautas-karpavicius: Have you considered generating uuid once in the `init()` and then use it? It may be safer a safer way to prevent double generation.

Groxx: tbh I think double-generation may actually be desirable - what if they start the same worker twice in the same process? Either with a partial shutdown and starting it back up, or just running two concurrently.  E.g. the Kafka ingester we have works somewhat like this.

vytautas-karpavicius: Yeah, in that case it makes sense.

vytautas-karpavicius (approved):

PR #1137, opened 2021-10-05
Enhancing docs for selector, from feedback and careful testing

Groxx:
#1131 was merged with some mostly-minor feedback to tackle, and this almost completes it.

I have not yet run integration tests against signal channel behavior when closed, and I'm
not yet entirely happy with the docs around signal channels in general, but I think this
is a net improvement worth merging.

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**


yux0 (approved):

internal/workflow.go
```diff
@@ -83,36 +99,69 @@ type (
 		//   decoding will be attempted, so you can try it yourself.
 		// - for other channels, an interface{} pointer.  All values are interfaces, so this will never fail, and you
 		//   can inspect the type with reflection or type assertions.
-		Receive(ctx Context, valuePtr interface{}) (more bool)
+		Receive(ctx Context, valuePtr interface{}) (ok bool)
```
Groxx, line 102: more consistent ok/more use across the board to match the go language reference's variable names.

internal/workflow.go
```diff
@@ -137,20 +186,16 @@ type (
...
-	//  			counter += i
-	//  		}
-	//  	case i, more := <- chB:
-	//  		if more {
-	//  			counter += i
-	//  		}
+	//  	case x := <- chA:
+	//  		counter += i
+	//  	case y := <- chB:
+	//  		counter += i
```
Groxx, line 198: decided the ok/more/subtle difference in behavior wasn't worth addressing, as it made the code less normal and more complex.

for chans that are never closed and only contain the right values (no Receive errors), which is often the case in real code, both before and after behave identically.

internal/workflow.go
```diff
@@ -83,36 +99,69 @@ type (
...
+		// But if your output values are zero values, this is equivalent to a simpler form:
+		//  select {
+		//  case value, ok := <- aChannel:
+		//  	more = ok
+		//  default:
+		//  	// no value was read
+		//  	ok = false
+		//  	// but the read would have blocked, so the channel is not closed
+		//  	more = true
 		//  }
```
Groxx, line 157: This took a bit of time to nail down precisely :|  it's a bit odd (I don't believe I've ever seen this kind of select statement in normal go code), but may be useful in some cases.

PR #1142, opened 2021-10-11
fmt, removing globals from tests, scoping test logs

Groxx:
... or nearly.  TestWorkerOptionDefaults currently produces a log because
we have an info-level log when no logger is set, and... I think that makes sense.
```
=== RUN   TestWorkerOptionDefaults
{"level":"info","ts":"2021-10-11T16:42:58.778-0700","caller":"internal/internal_worker.go:223","msg":"No logger configured for cadence worker. Created default one."}
{"level":"info","ts":"2021-10-11T16:42:58.779-0700","caller":"internal/internal_worker.go:227","msg":"No metrics scope configured for cadence worker. Use NoopScope as default."}
```
A better behavior may be to wait to produce that log until the worker is started,
but that's a bit out of scope for this PR + we may not want to do that anyway.

I also got rid of a couple suspicious globals that have no need to be globals, and
added a bit of documentation about what the heck they were doing anyway.

### DIFF
--- a/internal/internal_workers_test.go
+++ b/internal/internal_workers_test.go
@@ -510,6 +510,15 @@ func (s *WorkersTestSuite) TestQueryTask_WorkflowCacheEvicted() {
 		DisableActivityWorker: true,
 		Identity:              "test-worker-identity",
 		DataConverter:         dc,
+		// set concurrent decision task execution to 1,
+		// otherwise query task may be polled and start execution
+		// before decision task put created workflowContext into the cache,
+		// resulting in a cache hit for query
+		// by setting concurrent execution size to 1, we ensure when polling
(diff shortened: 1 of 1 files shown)


Groxx:
Well.  Unfortunately this exposes a race due to our incorrect shutdown logic, since we're using `testing.T` logs more :\  In testing.go there's this note:
```
// Do not lock t.done to allow race detector to detect race in case
// the user does not appropriately synchronizes a goroutine.
t.done = true
if t.parent != nil && atomic.LoadInt32(&t.hasSub) == 0 {
	t.setRan()
}
```

If it happens too often, we can revert.  Otherwise tbh I'd like to treat this as motivation to fix our leaky and incorrect shutdown logic.

PR #1149, opened 2021-11-09
fix 0.18.4

Groxx:
- Revert "Added 2-way proto-thrift mapper (#1130)"
    - nil child run-ids are causing panics, just reverting for now.
- fmt, removing globals from tests, scoping test logs (#1142)
    - needed for build, plus broadly useful.
    - conflicted on jwt-authorization stuff as that's not part of this branch, but cleanup was easy.
- Retract 0.18.4, and prep for 0.18.5


Shaddoll (approved):

Groxx:
Merged via cherry-picking (we should see if we can enable a rebase-y PR merge...)

PR #1150, opened 2021-11-10
Fix NPE when canceling workflow without run id

vytautas-karpavicius:
**What changed?**
Fixed NPE in proto mapper.

**Why?**
When canceling current workflow (without run id), this field can be nil.

**How did you test it?**
Added new integration test to reproduce, fix and verify.

**Potential risks**


fenghan89 (approved):

yycptt (approved):
Do we have the same issue for signaling external workflow? I think runID can also be empty there.

vytautas-karpavicius:
> Do we have the same issue for signaling external workflow? I think runID can also be empty there.

RunId can be empty there as well. But it is slightly different. Because workflowId/runId are both with WorkflowExecution, so they map 1:1 for proto/thrift. For cancelation WorkflowID, RunID fields were directly on request type, thus a different mapper being used.

PR #1152, opened 2021-11-11
Use generated proto types from cadence-idl repo

vytautas-karpavicius:
**What changed?**
- `go get github.com/uber/cadence-idl`
- Removed proto file generation from Makefile
- Removed existing generated public API proto types
- Replaced `go.uber.org/cadence/.gen/proto/api/v1` with `github.com/uber/cadence-idl/go/proto/api/v1` everywhere

**Why?**
Public API proto types are generated on cadence-idl repo (https://github.com/uber/cadence-idl/pull/94) so that they could be shared between client/server. Otherwise it results in duplicated type registrations when using both.

**How did you test it?**

**Potential risks**
Unfortunately this is a BREAKING CHANGE for people that started using `compatibility` package. Good thing is that fix is easy - replace one import path `go.uber.org/cadence/.gen/proto/api/v1` -> `github.com/uber/cadence-idl/go/proto/api/v1`

Will need to mention this in the change log of the next release.


Groxx (approved):
SGTM, and nice and simple :+1:

Should we tag a 0.19.x next maybe, to point out the more-likely-to-be-breaking change?

PR #1153, opened 2021-11-22
Introduce CHANGELOG.md

vytautas-karpavicius:
**What changed?**
Introducing changelog based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format.

Collected changes from previous 0.18.5 release.

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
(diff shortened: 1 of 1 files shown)


Groxx (approved):
bleh.  Sorry about the delay - this looks good!  Lets do it.

There are some kinda-easy ways to mechanically transform these kinds of changelogs to mark a version release, think we should add a makefile target for it?  Assuming we can find a script that doesn't look to bad, anyway.

vytautas-karpavicius:
> bleh. Sorry about the delay - this looks good! Lets do it.
> 
> There are some kinda-easy ways to mechanically transform these kinds of changelogs to mark a version release, think we should add a makefile target for it? Assuming we can find a script that doesn't look to bad, anyway.

I'm not against it, but marking the version isn't that hard to do manually either. IMHO we can live without it for now.

PR #1155, opened 2021-12-01
Modernize makefile, to make it stable like github.com/uber/cadence

Groxx:
Builds now work reliably, even in parallel, and it recovers from previous
builds (whether successful or not) without running more than necessary.

We should switch to revive and goimports like the server does, and ideally
both will move to `go test ./...` eventually (with build tags or something
to exclude integration tests) to make the tests run more normally.  We can
also take advantage of Go's built-in coverage merging and reports that way.

But those are for another day.  This is just setting a saner foundation.

---

At a very high level, this brings the client's makefile in line with the
server's, and adds `make build` and `make all` targets for dev-simplicity
(and visibility).
These targets will do a full build / lint _only when necessary_, i.e. after
changes have been made.  They should be more than sufficient for most uses,
as they now ensure thrift/fmt/copyright headers consistently, but you can run
`make lint` to get the output again if desired, and some stages can be run
independently without running later ones (like fmt).

To see current top-level targets, just run `make`.  It'll print help output:
```
❯ make
help                 Prints a help message showing any specially-commented targets
all                  refresh codegen, lint, and ensure the dummy binary builds, if necessary
build                ensure all packages build
copyright            update copyright headers
deps                 Check for dependency updates, for things that are directly imported
deps-all             Check for all dependency updates
errcheck             (re)run errcheck
fmt                  run gofmt
lint                 (re)run golint
staticcheck          (re)run staticcheck
test                 run all tests (requires a running cadence instance)
unit_test            run all unit tests
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,12 +10,10 @@
 *.html
 .tmp/
 /vendor
-bins
-cadence-client
-bins
 test.log
 /dummy
 .build
 .bins
+.bin
 .DS_Store
 .gobincache
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,9 +6,8 @@ This doc is intended for contributors to `cadence-client` (hopefully that's you!
 
 ## Development Environment
 
-* Go. Install on OS X with `brew install go`. The minimum required Go version is 1.10.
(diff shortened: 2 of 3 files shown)


mantas-sidlauskas (approved):
lgtm

PR #1167, opened 2022-05-19
Moving retryable-err checks to errors.As, moving some to not-retryable

Groxx:
We've known for a while that we have a service-busy-error-triggered retry storm,
this should fix that and also reduce a few other "not worth retrying" retries, so
they eat up less of a domain's frontend quota (where applicable to frontends).

Commits are ordered for ease of reading / verifying:
- just adds test
- just moves to errors.As, tests still pass
- changes which types are retried


internal/internal_retry.go
```diff
@@ -62,26 +63,61 @@ func createDynamicServiceRetryPolicy(ctx context.Context) backoff.RetryPolicy {
...
+	if target := (*s.FeatureNotEnabledError)(nil); errors.As(err, &target) {
+		return false
+	}
+	if target := (*s.LimitExceededError)(nil); errors.As(err, &target) {
+		return false
+	}
+	if target := (*s.QueryFailedError)(nil); errors.As(err, &target) {
+		return false
+	}
+	if target := (*s.ServiceBusyError)(nil); errors.As(err, &target) {
```
davidporter-id-au, line 104: this is a transient, retriable error, but needs to be backed off

internal/internal_retry_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	s "go.uber.org/cadence/.gen/go/shared"
+)
+
+func TestErrRetries(t *testing.T) {
+	t.Run("retryable", func(t *testing.T) {
+		for _, err := range []error{
+			&s.CurrentBranchChangedError{},      // unsure, but second attempt may get the latest branch?
```
vytautas-karpavicius, line 14: FYI, `CurrentBranchChangedError`, `RetryTaskV2Error`, `InternalDataInconsistencyError` and `RemoteSyncMatchedError` are all internal errors and should never even appear for client. In protobuf, they were even [moved out](https://github.com/uber/cadence/blob/master/proto/internal/uber/cadence/shared/v1/error.proto) of public IDLs.  Probably does not matter for this PR though.

internal/internal_retry_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+			&s.InternalServiceError{},  // fallback error type from server, "unknown" = "retry might work"
+			errors.New("unrecognized"), // fallback error type elsewhere, "unknown" = "retry might work"
+		} {
+			assert.True(t, isServiceTransientError(err), "%T should be transient", err)
+		}
+	})
+	t.Run("terminal", func(t *testing.T) {
+		for _, err := range []error{
+			&s.AccessDeniedError{},                      // access won't be granted immediately
```
mantas-sidlauskas, line 27: big 👍 for comments here

mantas-sidlauskas, line 27: big 👍 for comments here

internal/internal_retry_test.go
```diff
@@ -0,0 +1,53 @@
...
+	s "go.uber.org/cadence/.gen/go/shared"
+)
+
+func TestErrRetries(t *testing.T) {
+	t.Run("retryable", func(t *testing.T) {
+		for _, err := range []error{
+			// service-busy means "retry later", which is still transient/retryable.
+			// callers with retries MUST detect this separately and delay before retrying,
+			// and ideally we'll return a minimum time-to-wait in errors in the future.
+			&s.ServiceBusyError{},
```
Groxx, line 17: tackled in a followup PR, as I want that a separate commit/review

mantas-sidlauskas (approved):

PR #1169, opened 2022-05-30
Prevent local-activity panics from taking down the worker process

Groxx:
Previously, a local activity that panicked would lead to a panic
exiting the goroutine that performed the call, terminating the whole
worker process immediately.

Obviously this is not good behavior.  So now that's fixed.

Seems like it has been this way since local activities were introduced,
in commit 79c072aea60cbf2ed9bfc47ca85e793a8c07d77b , so it's pretty
surprising that it went unnoticed / unfixed for this long.
Particularly since it stands a decent chance of killing every worker
instance due to decision task retries.

--------

I'm not quite confident that this is done, but I'm putting it up for early
review / eyeballs / tests while I read more of the surrounding code.

In particular, a simple and non-breaking (because this was always broken)
change that may be worth making is to _not_ do retries on local activities
that panic:
```
func (w *workflowExecutionContextImpl) retryLocalActivity(lar *localActivityResult) bool {
	// add IsPanicError to the list
	if lar.task.retryPolicy == nil || lar.err == nil || IsCanceledError(lar.err) || IsPanicError(lar.err) {
		return false
	}
```
Since panics are inherently a bit more dangerous / imply a more significant
violation of expectations, perhaps it's worth treating them as fatal?


vytautas-karpavicius (approved):
Looks reasonable to me.

Groxx:
On second thought: nah, allow retries on panics.  If they have retries configured, they'll have retries, and the onus is on them to include `errRetryPanic` in the non-retryable list.

PR #1170, opened 2022-05-31
Do not retry on LimitExceededError

mantas-sidlauskas:
**What changed?**
Mark `LimitExceededError` as non transient error

**Why?**
There's no need to retry signal operation as error will not go away eventually

**How did you test it?**
Unit test

**Potential risks**


demirkayaender (approved):

davidporter-id-au (approved):

davidporter-id-au:
I reckon the `limitExceeded` error fix is probably semi-urgent I now realise (apologies for missing it before). Maybe we can just merge in this fix and Steven's larger one can happen after?

PR #1171, opened 2022-06-13
Stop retrying get-workflow-history with an impossibly-short timeout

Groxx:
Previously, when a `GetWorkflow(...).Get(...)` call timed out while waiting
for the workflow to complete, this happened:
- N-1 requests of the default long-poll timeout occurred, and it correctly retried
- The final request would have something like 5 seconds left, so it performs that call
- The server gives up the request *slightly before* that timeout so it can return the
  next page token for a future request to use (as happened in the N-1 earlier requests)
- Since no history was received and there's still time left in the context (~50ms
  internally), another impossibly-short request was sent
- This request fails immediately with a "insufficient time for long poll request" error.
- *This error* is what is returned from `Get(...)`

Which is pretty clearly sub-optimal.
Both because we sent a request that is almost guaranteed to fail, and because the error
returned to the caller is fairly generic looking / doesn't describe what happened.

What happened is that we ran out of time.
So this now returns a DeadlineExceeded error, like any other timeout, and does not
cause that final request to occur.

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/internal_workflow_client.go
+++ b/internal/internal_workflow_client.go
@@ -512,12 +512,18 @@ func (wc *workflowClient) GetWorkflowHistory(
 		var err error
 	Loop:
 		for {
+			var isFinalLongPoll bool
 			err = backoff.Retry(ctx,
 				func() error {
 					var err1 error
 					tchCtx, cancel, opt := newChannelContext(ctx, wc.featureFlags, func(builder *contextBuilder) {
 						if isLongPoll {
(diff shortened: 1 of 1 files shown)


Groxx:
I should add a test or three for this, and I think I might need a different error message (since I believe it's not *guaranteed* that it's due to waiting on the workflow, the database could've just failed to respond)... but this works well locally at least.

demirkayaender (approved):

PR #1174, opened 2022-06-22
Retry service-busy errors after a delay

Groxx:
Builds on #1167, but adds delay before retrying service-busy errors.

For now, since our server-side RPS quotas are calculated per second, this delays
at least 1 second per service busy error.
This is in contrast to the previous behavior, which would have retried up to about
a dozen times in the same period, which is the cause of service-busy-based retry
storms that cause lots more service-busy errors.

---

This also gives us an easy way to make use of "retry after" information in errors
we return to the caller, though currently our errors do not contain that.

Eventually this should probably come from the server, which has a global view of
how many requests this service has sent, and can provide a more precise delay to
individual callers.
E.g. currently our server-side ratelimiter works in 1-second slices... but that
isn't something that's guaranteed to stay true.  The server could also detect truly
large floods of requests, and return jittered values larger than 1 second to more
powerfully stop the storm, or to allow prioritizing some requests (like activity
responses) over others simply by returning a lower delay.


davidporter-id-au (approved):

internal/common/backoff/retry.go
```diff
@@ -30,8 +30,10 @@ type (
 	// Operation to retry
 	Operation func() error
 
-	// IsRetryable handler can be used to exclude certain errors during retry
-	IsRetryable func(error) bool
```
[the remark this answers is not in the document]
Groxx: we have `IsRetryable` on the server side, but it's just a bool-returning func like this used to be.  Naming-wise though I think either's fine, and I may have settled on `RetryAfter` due to an earlier abandoned attempt 🤔.  I'm game to switch back to `IsRetryable`, seems fine since it returns a bool.

---

As to wait-time: after re-thinking stuff for a bit and half thinking through a few scenarios where it would be defensible to ignore the server's "back off until" time...  

... yeah, we should just trust the server's value in all cases.  Otherwise it can't protect itself from misbehavior, e.g. if a client-side bug leads to hammering the server for some reason.  The lower we can push that logic in code, the less likely it is to be ignored by accident.

And in cases like heartbeats, where there are "meta" timeouts to consider (e.g. the server won't know about the remaining heartbeat limit, because _it can't_, because it's rejecting the request without bothering to do that expensive lookup), we can resolve that later by passing that kind of information in the request and having the server take that into account when rejecting the request.  E.g. if we send a heartbeat with a "we only have 2s left" field, the server can reduce a retry-after down to 1s rather than 5s to give it another chance, effectively prioritizing those heartbeat requests over others.

So since any client-side delay tweaks are reproducible with server-side code and sending more data if necessary, lets just bake in that the server's value is the best one to use.  That way the server can more reliably influence callers when needed.

internal/common/backoff/retry.go
```diff
@@ -103,16 +106,40 @@ Retry_Loop:
 			return err
 		}
 
-		// Check if the error is retryable
-		if isRetryable != nil && !isRetryable(err) {
```
Groxx, line 107: `isRetryable == nil` was only true in tests, so what was changed and now it's just assumed to exist in all cases.

internal/common/backoff/retry.go
```diff
@@ -123,3 +150,21 @@ Retry_Loop:
...
+// ErrRetryableAfter returns a minimum delay until the next attempt.
+//
+// for most errors this will be 0, and normal backoff logic will determine
+// the full retry period, but e.g. service busy errors (or any case where the
+// server knows a "time until it is not useful to retry") are safe to assume
+// that a literally immediate retry is *not* going to be useful.
+//
+// note that this is only a minimum, however.  longer delays are assumed to
+// be equally valid.
+func ErrRetryableAfter(err error) (retryAfter time.Duration) {
```
Groxx, line 163: decided to move it here, and one thing needs it externally, so it's exposed.  and I kinda like the `backoff.ErrRetryableAfter` package/name, keeps it clear that it's retry-backoff-related.

internal/internal_task_pollers.go
```diff
@@ -765,12 +765,27 @@ func (wtp *workflowTaskPoller) poll(ctx context.Context) (interface{}, error) {
 
 	response, err := wtp.service.PollForDecisionTask(ctx, request, getYarpcCallOptions(wtp.featureFlags)...)
```
shijiesheng, line 766: Do you think we should use `backoff.Retry` like other endpoints? In this way, we don't have to expose `ErrRetryableAfter` to the outside and it's more consistent across all endpoints

davidporter-id-au (approved):
nice!

internal/internal_task_pollers.go
Groxx: there's an exponential backoff at a higher level, this is within that retry.

it may be easier to change that retryer to check the error value though 🤔.  I think this one was ignoring a number of categories, but I'm not 100% sure why.

shijiesheng: right. Let's land it first. This refactoring can come later

shijiesheng (approved):

Groxx:
Merging, will try to follow up this week with a cleanup (if feasible, given the custom behavior I remember... I suspect it won't be, but worth checking on anyway).

internal/internal_task_pollers.go
Groxx: Update!

Nope.  The poller uses a unique kind of ratelimiter and retry loop :|  Changing that is substantial and unifying the two strategies is probably not going to be an improvement, at least without a bigger re-thinking of retries and ratelimiting in general.

PR #1179, opened 2022-07-15
Simplify worker options configuration value propagation

shijiesheng:
**What changed?**

use embedding to simplify assignment

**Why?**

There are duplicate fields in `workerExecutionParameters` and `WorkerOptions` that unnecessarily complicates the logic. 

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -396,7 +396,7 @@ func newWorkflowTaskHandler(
 		dataConverter:                  params.DataConverter,
 		contextPropagators:             params.ContextPropagators,
 		tracer:                         params.Tracer,
-		workflowInterceptors:           params.WorkflowInterceptors,
+		workflowInterceptors:           params.WorkflowInterceptorChainFactories,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


internal/internal_task_handlers.go
```diff
@@ -396,7 +396,7 @@ func newWorkflowTaskHandler(
 		dataConverter:                  params.DataConverter,
 		contextPropagators:             params.ContextPropagators,
 		tracer:                         params.Tracer,
-		workflowInterceptors:           params.WorkflowInterceptors,
+		workflowInterceptors:           params.WorkflowInterceptorChainFactories,
```
davidporter-id-au, line 399: what is a Chain factory?

shijiesheng: It's an option field related to replay. Here is the link. Here it looks like we need a consolidation on the naming. 
https://sourcegraph.com/github.com/uber-go/cadence-client@master/-/blob/internal/worker.go?L200

shijiesheng: Unfortunately, both WorkerOptions and ReplayOptions are exported. But they have different names. 
https://sourcegraph.com/github.com/shijiesheng/cadence-client@master/-/blob/worker/worker.go#L191:43
https://sourcegraph.com/github.com/shijiesheng/cadence-client@master/-/blob/worker/worker.go#L178:37

shijiesheng, line 399: I'll rename this as well

davidporter-id-au (approved):

PR #1181, opened 2022-08-03
Set up go-apidiff github action on PRs

Groxx:
This may also be worth running for all pushes, particularly to `A.B.x` release branches, but we might as well start simple.

(currently testing things out, not sure how this behaves when newly added)


Groxx:
seems totally ineffective :\

Groxx:
ah.  maybe it does work, but golang.org/x/exp/apidiff has.... shall we say "some" gaps.  import changes being one of them.
maybe this can be saved or recreated with a patch... 🤔

Groxx:
does work, though it'd be nice if it were turned into a comment 🤔 
```
go.uber.org/cadence/workflow
  Incompatible changes:
  - ExecuteChildWorkflow: changed from func(go.uber.org/cadence/internal.Context, interface{}, ...interface{}) go.uber.org/cadence/internal.ChildWorkflowFuture to func(go.uber.org/cadence/internal.Context, string, interface{}, ...interface{}) go.uber.org/cadence/internal.ChildWorkflowFuture
  - ExecuteLocalActivity: removed
  - GetInfo: changed from func(go.uber.org/cadence/internal.Context) *go.uber.org/cadence/internal.WorkflowInfo to func(go.uber.org/cadence/internal.Context, ...string) *go.uber.org/cadence/internal.WorkflowInfo
  Compatible changes:
  - AddedCompatibleFunc: added
Error: Process completed with exit code 1.
```

Groxx:
So with https://github.com/Groxx/exp/commit/6126f99dd2b900ba63ab4182e3742407a1e78d12 this is likely to catch more... but it's still missing large categories of behavior :\  not sure what may need to be done to tackle that.

PR #1184, opened 2022-08-20
Add poller autoscaler

shijiesheng:
**What changed?**

* added autoscaler package with interface and implementation of linearRecommender
* added pollerAutoscaler implementation

**Why?**

Pollers constantly poll the cadence-frontend for tasks even though there are no tasks. 
To reduce unnecessary polling, an autoscaler is used to limit the number of concurrent polls. This PR addresses the implementation of polleraAutoscaler and there will be a follow-up PR to add the autoscaler on pollers. 

As for the design of autoscaler, a resizable semaphore is used to limit concurrency. 

**How did you test it?**

unit tests

**Potential risks**


internal/common/autoscaler/recommender_test.go
```diff
@@ -0,0 +1,125 @@
...
+			args: args{
+				currentResource: 10,
+				currentUsages: map[UsageType]UsageInMilli{
+					PollerUtilizationRate: 600,
+				},
+			},
+			want: Resource(12),
+		},
+		{
+			name:   "over utilized, scale up",
```
ZackLK, line 95: Duplicates test above?

internal/common/autoscaler/recommender_test.go
```diff
@@ -0,0 +1,125 @@
...
+			args: args{
+				currentResource: 10,
+				currentUsages: map[UsageType]UsageInMilli{
+					PollerUtilizationRate: 500,
+				},
+			},
+			want: Resource(10),
+		},
+		{
+			name:   "under utilized, scale down",
```
ZackLK, line 62: Worth adding a test where usage is 0?

internal/internal_poller_autoscaler.go
```diff
@@ -0,0 +1,168 @@
...
+	return autoscaler.Resource(p.sem.GetLimit())
+}
+
+// Start an auto-scaler go routine and returns a done to stop it
+func (p *pollerAutoScaler) Start() autoscaler.DoneFunc {
+	logger := p.logger.Sugar()
+	go func() {
+		for {
+			select {
+			case <-p.ctx.Done():
```
ZackLK, line 112: This should return when cancelled so the goroutine doesn't leak?

internal/internal_poller_autoscaler_test.go
```diff
@@ -0,0 +1,248 @@
...
+					defer wg.Done()
+					for pollResult := range pollChan {
+						pollerScaler.Acquire(1)
+						pollerScaler.CollectUsage(pollResult)
+						pollerScaler.Release(1)
+					}
+				}()
+			}
+			wg.Wait()
+			time.Sleep(time.Millisecond * 500)
```
ZackLK, line 154: Instead of sleeping yourself, better would be to use assert.Eventually() - less chance of race condition and should test faster.

internal/common/autoscaler/types.go
```diff
@@ -0,0 +1,50 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package autoscaler
+
+type (
+	// Resource is the unit of scalable resources
+	Resource uint
```
ZackLK, line 25: Should this be called ResourceCount, or something that signifies that it's an int / unit? The way "Resource" reads to me makes it sound like its more than an int.

internal/common/autoscaler/types.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// UsageInMilli is the unit of Resource usage times 1000
+	UsageInMilli uint64
```
ZackLK, line 28: And just to clarify the naming of this - its got nothing to do with milliseconds, right? As far as I can tell, its just resource usage * 1000. Milli in the name makes it sound like milliseconds, but maybe thats just me. Would this not work as a float?

internal/common/autoscaler/recommender_test.go
shijiesheng: removed

internal/common/autoscaler/recommender_test.go
shijiesheng: added

internal/common/autoscaler/types.go
shijiesheng: Initially I was more ambitious in thinking a single autoscaler can handle multiple resources. But it doesn't seem we will need it any time soon. So yeah, I've changed it to ResourceUnit.

internal/common/autoscaler/types.go
shijiesheng: I've changed to MilliUsage similar to how kubernetes autoscaler named Millicores. https://sourcegraph.com/github.com/kubernetes/autoscaler@master/-/blob/vertical-pod-autoscaler/pkg/recommender/logic/recommender.go#L27:5 

I was trying to avoid using float to avoid negative numbers from the beginning. Otherwise, I'll need to handle negative cases in a lot of places unnecessarily.

internal/internal_poller_autoscaler_test.go
shijiesheng: TIL. Thank you for the advice. I've added optional hook functions in the pollerAutoscaler implementation to simply the testing.

internal/internal_poller_autoscaler.go
shijiesheng: good catch!

ZackLK (approved):
Looks good, thanks for making those changes!

PR #1186, opened 2022-08-30
add poller autoscaling in activity and decision workers

shijiesheng:
**What changed?**

* add pollerAutoScalerOptions
* surface user set options and the feature flag to enable poller autoscaling
* start autoscaling in the workers

**Why?**

**How did you test it?**

**Potential risks**


internal/worker.go
```diff
@@ -95,6 +102,27 @@ type (
...
 
+		// optional: Sets the minimum number of goroutines that will concurrently poll the
+		// cadence-server to retrieve decision tasks. If FeatureFlags.PollerAutoScalerEnabled is set to true,
+		// changing this value will NOT affect the rate at which the worker is able to consume tasks from a task list.
+		// Default value is 1
+		MinConcurrentDecisionTaskPollers int
+
+		// optional: Sets the interval of poller autoscaling, between which poller autoscaler changes the poller count
+		// based on poll result. It takes effect if FeatureFlags.PollerAutoScalerEnabled is set to true.
+		// Default value is 1 min
```
ZackLK, line 113: Is 1 minute fast enough?

ZackLK (approved):

internal/worker.go
shijiesheng: I'm still experimenting on this. Since we ensure at least one poller is always working, the impact of this parameter should be low.

internal/internal_poller_autoscaler.go
```diff
@@ -60,31 +62,40 @@ type (
...
-	initialPollerCount int,
-	minPollerCount int,
-	maxPollerCount int,
-	targetMilliUsage uint64,
-	isDryRun bool,
-	cooldownTime time.Duration,
+	options pollerAutoScalerOptions,
 	logger *zap.Logger,
 	hooks ...func()) *pollerAutoScaler {
 	ctx, cancel := context.WithCancel(context.Background())
```
Groxx, line 81: is this a sign we should be passing in a context?  e.g. to propagate shutdown information maybe.

internal/internal_poller_autoscaler.go
```diff
@@ -139,9 +150,12 @@ func (p *pollerAutoScaler) Start() autoscaler.DoneFunc {
...
 	}()
-	return func() {
-		p.cancel()
-	}
+	return
+}
+
+// Stop stops the poller autoscaler
+func (p *pollerAutoScaler) Stop() {
+	p.cancel()
```
Groxx, line 158: not relevant to this PR, but in general: we probably want a (time-bounded?) blocking "... and wait for shutdown to complete" behavior here.
otherwise it's impossible to rule out dangling goroutines (e.g. goleak tests) or race conditions (background loop triggering at the same time as cancelation).

internal/internal_worker_base.go
```diff
@@ -260,6 +276,14 @@ func (bw *baseWorker) pollTask() {
 	var task interface{}
 	bw.retrier.Throttle()
 	if bw.pollLimiter == nil || bw.pollLimiter.Wait(bw.limiterContext) == nil {
+		if bw.pollerAutoScaler != nil {
+			if pErr := bw.pollerAutoScaler.Acquire(1); pErr == nil {
+				defer bw.pollerAutoScaler.Release(1)
+			} else {
+				bw.logger.Warn("poller auto scaler acquire error", zap.Error(pErr))
+			}
+		}
```
Groxx, line 285: I'm pretty sure this ordering means this is possible:
- autoscale from N to 1
- allow N to get through pollLimiter, while autoscaler allows one long request and blocks N-1
- long request finishes, N-1 run through as fast as possible, ignoring ratelimter
  - sequentially, as autoscaler is at 1.  or 2 at once if at 2, etc.

swapping the order seems better, since "waiting on the per-second request limiter" feels like "a concurrent poller" to me.  the request itself isn't running, but we've reserved space for it to run "on time".  so we'll degrade "concurrent requests on the network" (allowing it to go lower than specified) in favor of preserving "requests per second".

Groxx (commented):
A minor thing worth changing, but otherwise this all looks reasonable / is mostly "just" a (good) refactor.

so 👍 structurally LGTM.  I should read more of the autoscaler before reviewing in more depth though.

---

unrelated / not a problem for any realistic valuesfor us, but the semaphore library doesn't detect overflows: https://github.com/marusama/semaphore/blob/master/semaphore.go#L98
we should probably make a PR for that.  it'd be easy to fix.

PR #1194, opened 2022-11-03
Cdnc 2847 Add version metrics to the clients

allenchen2244:
Added a metric for the workflow client and domain client versions

So we can have graphs to see which versions are running

Was able to deploy to canary staging without issue and query and see the metrics in Grafana

Just a metric shouldn't be any risk


allenchen2244:
Moving into cadencefx

PR #1197, opened 2022-11-15
changing dynamic poller scaling strategy.

mindaugasbarcauskas:
**What changed?**

Updating dynamic poller scaling strategy in the case that poller utilisation is 100%

**Why?**
currently how linear scaler works is that if utilisation is at 100% and the target utilisation is 50% then the number of pollers will double

so in case we have one poller and a large backlog of tasks to process our poller scaling will look like this

1 -> 2 -> 4 -> 8 -> 10 (maximum)

However this would mean that the latency to pick up tasks in case of bursty traffic would increase.

**How did you test it?**

Unit test.

**Potential risks**

Amount of pollers may increase too suddenly and cause load issues for the worker (worst case)


shijiesheng:
One unit test failed FAIL: Test_pollerAutoscaler/over_utilized,_scale_up (0.05s)

PR #1201, opened 2022-12-02
Fix flaky test

mindaugasbarcauskas:
**What changed?**

Fixing flaky integration test since master build was failing. Test was flaky because it was testing that child workflow gets terminated when parent workflow finishes. The test was running into race condition.

**Why?**

**How did you test it?**
pipeline run.

**Potential risks**

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -421,7 +421,8 @@ func (ts *IntegrationTestSuite) TestChildWFWithParentClosePolicyTerminate() {
 	ts.NoError(err)
 	resp, err := ts.libClient.DescribeWorkflowExecution(context.Background(), childWorkflowID, "")
 	ts.NoError(err)
-	err = ts.getWorkflow(resp.WorkflowExecutionInfo.Execution.GetWorkflowId(), resp.WorkflowExecutionInfo.Execution.GetRunId())
+	// Need to wait for child workflow to finish as well otherwise test becomes flaky
+	err = ts.waitForWorkflowFinish(resp.WorkflowExecutionInfo.Execution.GetWorkflowId(), resp.WorkflowExecutionInfo.Execution.GetRunId())
 	resp, err = ts.libClient.DescribeWorkflowExecution(context.Background(), childWorkflowID, "")
 	ts.NoError(err)
 	ts.True(resp.WorkflowExecutionInfo.GetCloseTime() > 0)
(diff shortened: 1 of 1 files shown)


test/integration_test.go
```diff
@@ -581,6 +586,15 @@ func (ts *IntegrationTestSuite) registerWorkflowsAndActivities(w worker.Worker)
 	ts.activities.register(w)
 }
 
+func (ts *IntegrationTestSuite) waitForWorkflowFinish(wid string, runId string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
+	defer cancel()
+	wfRun := ts.libClient.GetWorkflow(ctx, wid, runId)
+	var val interface{}
+	err := wfRun.Get(ctx, &val)
+	return err
```
mantas-sidlauskas, line 595:
```suggestion
	return wfRun.Get(ctx, &val)
```

test/integration_test.go
```diff
@@ -421,7 +421,12 @@ func (ts *IntegrationTestSuite) TestChildWFWithParentClosePolicyTerminate() {
 	ts.NoError(err)
 	resp, err := ts.libClient.DescribeWorkflowExecution(context.Background(), childWorkflowID, "")
 	ts.NoError(err)
+	// Need to wait for child workflow to finish as well otherwise test becomes flaky
+	err = ts.waitForWorkflowFinish(resp.WorkflowExecutionInfo.Execution.GetWorkflowId(), resp.WorkflowExecutionInfo.Execution.GetRunId())
```
mantas-sidlauskas, line 425: this error is not checked

mindaugasbarcauskas: Not checked by design. since this call can throw terminated error which we do not care about.

test/integration_test.go
mindaugasbarcauskas: updated.

test/integration_test.go
```diff
@@ -581,6 +586,14 @@ func (ts *IntegrationTestSuite) registerWorkflowsAndActivities(w worker.Worker)
 	ts.activities.register(w)
 }
 
+func (ts *IntegrationTestSuite) waitForWorkflowFinish(wid string, runId string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), ctxTimeout)
+	defer cancel()
+	wfRun := ts.libClient.GetWorkflow(ctx, wid, runId)
+	var val interface{}
+	return wfRun.Get(ctx, &val)
```
Groxx, line 594:
```suggestion
	return wfRun.Get(ctx, nil)
```

test/integration_test.go
```diff
@@ -421,7 +421,12 @@ func (ts *IntegrationTestSuite) TestChildWFWithParentClosePolicyTerminate() {
 	ts.NoError(err)
 	resp, err := ts.libClient.DescribeWorkflowExecution(context.Background(), childWorkflowID, "")
 	ts.NoError(err)
```
Groxx, line 423: probably not necessary any more?  I don't think GetWorkflow needs a runid anyway...

Groxx (approved):
Yep, makes sense that this'd be racy since child-wf-close isn't transactional.
I'd do some small cleanups, but they're pretty foolproof, so 👍-ing now.

PR #1202, opened 2022-12-08
Export GetRegisteredWorkflowTypes so I can use in shadowtest.

ZackLK:
Exported the GetRegisteredWorkflowTypes() function for public use.

In order to know what function types have been registered and be able to iterate them, and run the
WorkflowShadower on each exported function type.

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -789,7 +789,7 @@ func (aw *aggregatedWorker) Start() error {
 	}
 
 	if aw.workflowWorker != nil {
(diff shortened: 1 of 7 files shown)


Groxx:
👍 nice and easy after the cleanup.  thanks!
I think there's a good chance we'll want an activity version eventually, and probably some per-worker/registry versions as well... but I'm not sure how far we can take this without making some big breaking API changes.  The rest is probably better to wait for v2, though in general this seems like a useful thing to expose.

Groxx (approved):

PR #1206, opened 2023-01-16
ran "make fmt"

mindaugasbarcauskas:
**What changed?**

Just ran "make fmt" so that files are formatted correctly.

**Why?**

These changes come in when running "make build" and interfere when making changes.

**How did you test it?**
Should be no functional impact.

**Potential risks**
Not applicable.

### DIFF
--- a/.gen/go/cadence/cadence.go
+++ b/.gen/go/cadence/cadence.go
@@ -62,14 +62,14 @@ type WorkflowService_CountWorkflowExecutions_Args struct {
 // An error is returned if the struct or any of its fields failed to
 // validate.
 //
-//   x, err := v.ToWire()
-//   if err != nil {
-//     return err
-//   }
+//	x, err := v.ToWire()
+//	if err != nil {
+//	  return err
+//	}
 //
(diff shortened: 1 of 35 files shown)


Groxx:
We should probably:
- make sure our docker containers are updated, if we have any that are doing a build and looking for changes (iirc we have that on server, I forget about client)
- bake in the same kind of go version check as server has: https://github.com/uber/cadence/blob/master/Makefile#L31
- make sure you've done fmt with that version of Go

I'm assuming you did this with 1.19, but double check?  `>=1.19` and `<1.19` produce different output, which is the underlying issue you're hitting.

Groxx:
> I did it with 1.19 go version. Yeah I was wondering if gofmt might depend on local configuration. I will check on those points. Thanks Steven.

Normally it doesn't, but 1.19 made (good!) enhancements to godoc's capabilities and formatting... and unfortunately it thrashes with older versions :|  Normally they're fairly careful to make changes that are backwards compatible, but not this time.

davidporter-id-au (approved):

PR #1208, opened 2023-01-16
Making cancelWorkflow call extendable with options.

mindaugasbarcauskas:
**What changed?**

Adding ability to provide workflow cancel reason by making cancelWorkflow call extendable with options.
note: targeting updated-idl branch since those changes are not merged yet and it is stacked on top of that.

**Why?**

**How did you test it?**
Added unit tests.

**Potential risks**

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -46,6 +46,8 @@ const (
 )
 
 type (
+	Option func(r *s.RequestCancelWorkflowExecutionRequest)
+
 	// Client is the client for starting and getting information about a workflow executions as well as
 	// completing activities asynchronously.
 	Client interface {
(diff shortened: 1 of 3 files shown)


internal/client.go
```diff
@@ -139,7 +141,7 @@ type (
 		//	- BadRequestError
 		//	- InternalServiceError
 		//	- WorkflowExecutionAlreadyCompletedError
-		CancelWorkflow(ctx context.Context, workflowID string, runID string) error
+		CancelWorkflow(ctx context.Context, workflowID string, runID string, opts ...Option) error
```
Groxx, line 144: unfortunately changes to interfaces are always breaking :\  it's one of the things we need to address in the next api version, since it's so crippling.

Groxx:
Covered IRL, but to tl;dr:
- yea, we've been breaking APIs in pre-release for a while, this is a rather minor one (and most code should upgrade transparently).  let's merge it.
- no func closures for functional options, they're _far_ too problematic.  just use a regular type and a privately-implementing interface.

mindaugasbarcauskas:
closing created a different pull request

PR #1209, opened 2023-01-18
Adding go version check when building locally.

mindaugasbarcauskas:
**What changed?**

Adding check to print a warning when building with a different go version locally.

**Why?**
For example, when building locally formatting changes happen with go1.19 version

**How did you test it?**

Tested locally.

**Potential risks**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ BIN := .bin
 #
 # this can _likely_ remain a major version, as fmt output does not tend to change in minor versions,
 # which will allow findstring to match any minor version.
-EXPECTED_GO_VERSION := go1.16.3
+EXPECTED_GO_VERSION := go1.16
 CURRENT_GO_VERSION := $(shell go version)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,6 +23,19 @@ BUILD := .build
 # usually unnecessary to clean, and may require downloads to restore, so this folder is not automatically cleaned.
 BIN := .bin
 
+# current (when committed) version of Go used in CI, and ideally also our docker images.
+# this generally does not matter, but can impact goimports or formatting output.
+# for maximum stability, make sure you use the same version as CI uses.
+#
+# this can _likely_ remain a major version, as fmt output does not tend to change in minor versions,
+# which will allow findstring to match any minor version.
+EXPECTED_GO_VERSION := go1.16.3
```
Groxx, line 32: can probably change this to `go1.16` so it allows all 1.16 minor versions.  unless other 1.16 versions behave differently?

Groxx (approved):

PR #1210, opened 2023-01-19
update CI go version.

mindaugasbarcauskas:
**What changed?**
Updating docker files to use 1.19 go version in CI pipelines

**Why?**

**How did you test it?**

CI pipeline passes.

**Potential risks**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ BIN := .bin
 #
 # this can _likely_ remain a major version, as fmt output does not tend to change in minor versions,
 # which will allow findstring to match any minor version.
-EXPECTED_GO_VERSION := go1.16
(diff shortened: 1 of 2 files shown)


davidporter-id-au (approved):
LGTM, but maybe run it by @Groxx , I have less strong opinions here

Groxx:
yea, I think we can do this.  it'll let us start formatting for 1.19 and then we can get the godoc improvements too.

I believe the go.mod value will still prevent us from using generics and stuff that doesn't exist in 1.13 (which is good).  we should probably consider bumping that to at least 1.18 so we can use generics at some point.  1.18 is about 9 months old now, and 1.13 is ancient, seems probably safe to update and force that on our users.

Groxx (approved):

PR #1211, opened 2023-01-30
Updating IDL version for go client.

mindaugasbarcauskas:
**What changed?**

Updated idl version

**Why?**

**How did you test it?**

CI pipeline.

**Potential risks**

### DIFF
--- a/internal/compatibility/proto/request.go
+++ b/internal/compatibility/proto/request.go
@@ -632,6 +632,7 @@ func RestartWorkflowExecutionRequest(r *shared.RestartWorkflowExecutionRequest) 
 		Domain:            r.GetDomain(),
 		WorkflowExecution: WorkflowExecution(r.GetWorkflowExecution()),
 		Identity:          r.GetIdentity(),
+		Reason:            r.GetReason(),
 	}
 
 	return &request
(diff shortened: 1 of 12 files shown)


internal/compatibility/proto/request.go
```diff
@@ -623,3 +623,16 @@ func RefreshWorkflowTasksRequest(r *shared.RefreshWorkflowTasksRequest) *apiv1.R
...
+
+func RestartWorkflowExecutionRequest(r *shared.RestartWorkflowExecutionRequest) *apiv1.RestartWorkflowExecutionRequest {
+	if r == nil {
+		return nil
+	}
+	request := apiv1.RestartWorkflowExecutionRequest{
+		Domain:            r.GetDomain(),
+		WorkflowExecution: WorkflowExecution(r.GetWorkflowExecution()),
+		Identity:          r.GetIdentity(),
+	}
```
Groxx, line 635: seems like proto is missing the `reason` field?  is that worth adding in now so it happens all at once?

Groxx: ah, no, this proto API is missing a field that's in the thrift API.  we should fix that.

Groxx: Tackling this in https://github.com/uber/cadence-idl/pull/128

mindaugasbarcauskas: thanks Steven, very nice catch! I updated to latest idl where you included the reason and updated the mapper.

Groxx (approved):
LGTM 👍 

Probably worth choosing a more useful commit message though (e.g. it adds restart, it's not just a minor internal addition).

PR #1212, opened 2023-01-31
adding serviceBusy tag for transient-poller-failure counter metric.

mindaugasbarcauskas:
**What changed?**

Adding cause tag for poller transient failure metric to determine if the cause of the error is serviceBusy error which means that the pollers are being ratelimited.

**Why?**
With this it will be now possible to configure alerts in case a service is severely rate limited.

**How did you test it?**
Existing tests since only added tag to metric.

**Potential risks**

low risk change.


internal/internal_task_pollers.go
```diff
@@ -47,6 +48,9 @@ const (
 	stickyDecisionScheduleToStartTimeoutSeconds = 5
 
 	ratioToForceCompleteDecisionTaskComplete = 0.8
+
+	causeTag    = "Cause"
```
shijiesheng, line 52: might be better to put it into `internal_logging_tags.go`

internal/internal_task_pollers.go
```diff
@@ -767,8 +771,15 @@ func (wtp *workflowTaskPoller) poll(ctx context.Context) (interface{}, error) {
 	response, err := wtp.service.PollForDecisionTask(ctx, request, getYarpcCallOptions(wtp.featureFlags)...)
 	if err != nil {
 		retryable := isServiceTransientError(err)
+
 		if retryable {
-			wtp.metricsScope.Counter(metrics.DecisionPollTransientFailedCounter).Inc(1)
+			if target := (*s.ServiceBusyError)(nil); errors.As(err, &target) {
+				tagsMap := map[string]string{}
+				tagsMap[causeTag] = serviceBusy
+				wtp.metricsScope.Tagged(tagsMap).Counter(metrics.DecisionPollTransientFailedCounter).Inc(1)
```
shijiesheng, line 779: `Tagged(map[string]string{causeTag: serviceBusy})` for simplification

internal/internal_task_pollers.go
Groxx: I suppose we don't have a shared metric-tag-keys definition location, do we 🤔 

looking at e.g. internal_worker.go though:
- yea, a constant here looks fine.  we can consolidate them later, but I don't see a "cause" elsewhere so I don't think this will be confused with / clobber anything.
- lowercase it like all other metric keys.

internal/internal_task_pollers.go
Groxx: +1 for init-with-data whenever possible.  it's a safer habit, e.g. detects collisions.

internal/internal_task_pollers.go
```diff
@@ -1008,7 +1019,14 @@ func (atp *activityTaskPoller) poll(ctx context.Context) (*s.PollForActivityTask
 	if err != nil {
 		retryable := isServiceTransientError(err)
 		if retryable {
-			atp.metricsScope.Counter(metrics.ActivityPollTransientFailedCounter).Inc(1)
+
+			if target := (*s.ServiceBusyError)(nil); errors.As(err, &target) {
```
Groxx, line 1023: I kinda wonder if it'd be worth making a helper specifically for service-busy, as it's handled specially in a few locations...

but not at all requiring that now, just thinking out loud.

Groxx (changes requested):
I'll stick myself in the re-review list to make things clear.  just minor stuff, but poke me when done and let's land it :)

internal/internal_task_pollers.go
mindaugasbarcauskas: @Groxx addressed the comment, but all the keys are camelCase in internal_logging_tags, I think they get lowercased by the MetricsScope.

Groxx: all metrics are lowercased: https://github.com/uber-go/cadence-client/blob/master/internal/common/metrics/constants.go

tbh we should probably lowercase logs too, uppercase is annoying to deal with when using a case-sensitive system.

Groxx (approved):
👍 to unblock (probably should've to begin with tbh), but metrics are all lower_snake_cased.  works with all metrics systems, and uppercase is often worth avoiding (sadly).

log tags are different because ¯\_(ツ)_/¯.  tbh it'd probably be easier to use them if we snake_cased them too.

PR #1213, opened 2023-01-31
Adding ability to provide cancellation reason to cancelWorkflow API

mindaugasbarcauskas:
**What changed?**

Adding ability to provide options to cancel workflow API.

**Why?**

In order to allow populating cancel reason.

**How did you test it?**

Added unit tests.

**Potential risks**


Groxx (approved):
👍 
I wonder if the internal package references will make this mock more annoying to use in IDEs....

... but tbh I'm fine with that temporarily.  Getting rid of the internal alias is at the top of my list of things to change for v2.

PR #1214, opened 2023-02-04
Expose WithCancelReason and related types publicly, as originally intended

Groxx:
Missed this during review of #1213, sorry!
I'm not entirely sure how to catch things like this in tests, without a separate module...


mindaugasbarcauskas (approved):
thanks Steven, looks good.

PR #1215, opened 2023-02-04
Modernize makefile like server, split tools into their own module

Groxx:
Long needed, should help stabilize some things.  And make it easier to keep them in sync.

I'm fairly sure adding this go-version info to the build path will solve a couple things:
1. Our fmt-thrashing around 1.19, and probably future fmt changes, because tools won't be able to "leak" across Go versions.
2. In-docker-container architecture conflicts (maybe only an issue in server?).

Next up: switch to some of the better tools, like revive, and update them where possible.


Groxx:
not sure why coveralls isn't leaving a comment.  it uploaded the coveralls run just fine 🤔

davidporter-id-au (approved):

davidporter-id-au:
Not super familiar with the placement of the Go.mod you on this, but I'll just trust it

PR #1216, opened 2023-02-10
Add missing activity logger fields for local activities

Groxx:
Apparently these were never added, so local activity logs are just (dramatically) less useful.
That should now be fixed, I think these are the only places that need changing.

### DIFF
--- a/internal/internal_logging_tags.go
+++ b/internal/internal_logging_tags.go
@@ -36,6 +36,7 @@ const (
 	tagSideEffectID      = "SideEffectID"
 	tagChildWorkflowID   = "ChildWorkflowID"
 	tagLocalActivityType = "LocalActivityType"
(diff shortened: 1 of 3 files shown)


internal/internal_logging_tags.go
```diff
@@ -36,6 +36,7 @@ const (
 	tagSideEffectID      = "SideEffectID"
 	tagChildWorkflowID   = "ChildWorkflowID"
 	tagLocalActivityType = "LocalActivityType"
+	tagLocalActivityID   = "LocalActivityID"
```
mindaugasbarcauskas, line 39: Will this tag not cause cardinality issues?

mindaugasbarcauskas (approved):
In general, looks good, just one small question.

PR #1221, opened 2023-03-01
Prototype of a rate-limiter intended to favor workflows getting history over polling for new workflows.

Groxx:
This is to address an explosion of GetWorkflowExecutionHistory requests in one of our internal domains.
"Explosion" to the tune of: normally a couple hundred per second, but during this issue we saw up to ~100,000/s.

A larger description will come after I get some more sleep, but the quick and dirty summary is:
- they had many "live" workflows
- they started to build up a decision-schedule queue
- slowing them down
- overloading caches, causing a lot of un-cached decisions
- ... leading to a lot of history iterators in new workflows looping, trying to load history, and getting ratelimited...
- ... causing more to loop and try to load history...
- ... slowing things down further and making it worse.

Decision tasks were regularly >10 minutes, just trying to load history.

So this is an attempt to prevent that from happening.
It's not yet complete, just contains the limiter I'm planning, and tests.


internal/internal_worker_base.go
```diff
@@ -288,8 +291,9 @@ func (bw *baseWorker) pollTask() {
 		}
 	}
 
-	bw.retrier.Throttle()
+	bw.retrier.Throttle() // sleeps if retry policy determines it should sleep after failures
 	if bw.pollLimiter == nil || bw.pollLimiter.Wait(bw.limiterContext) == nil {
+		// TODO: block here on a shared semaphore with history-loading?
```
Groxx, line 296: this is where I'm planning on putting it.

at this point the error-throttler has already delayed things, and we're inside both the auto-scaler and per-second limiter, so an _immediate_ poll is correct, and a _delayed_ poll is... potentially exceeding per-second, but not exceeding concurrent.

this is not the most hygienic location, I think this whole thing might be worth revamping.  but I don't believe it will make it _behave_ worse in the meantime.

internal/pahlimiter/limiter_test.go
```diff
@@ -0,0 +1,329 @@
...
+		test := func() (int64, int64) {
+			l, err := NewWeighted(2, 1, 10)
+			require.NoError(t, err, "should be impossible")
+			defer l.Close()
+
+			return thrashForFairness(l)
+		}
+
+		// run N times in parallel to reduce noise.
+		// none may have zeros, and it should be heavily skewed in favor of history requests
```
Groxx, line 242: I've used this kind of "run, measure, fail if I can't find what I expect" thing in other tests I've written (not in this repo), it has worked out fairly well.

I'm definitely open to ideas to improve it though, this was sorta a blind duplication because I knew it'd be an issue because it's inherently noisy.  There's probably some better ways.

internal/pahlimiter/limiter.go
```diff
@@ -0,0 +1,305 @@
...
+// are allowed through this limiter to actually perform that request.
+//
+// In this situation it will still limit the number of actual concurrent requests to load history, but with a very
+// large increase in complexity.  If you want this, strongly consider just using NewHistoryLimited.
+//
+// ---
+//
+// All that said: this is NOT built to be a reliable blocker of polls for at least two reasons:
+//   - History iterators do not hold their resources between loading (and executing) pages of history, causing a gap
+//     where a poller could claim resources despite the service being "too busy" loading history from a human's view.
```
emrahs, line 137: When this happens, it'll lead to an increased number of decision tasks that timeout, because the bandwidth will be split across many workflows that are all trying to fetch history concurrently. Given a fixed bandwidth, when the number of actors increases the average latency increases. Once latency is high enough to cause the tasks to timeout before even the replay is done, the worker throughput will drop to zero since the Respond() calls to the cadence server will be rejected due to task timeout. 

In order to prevent that, we need to guarantee that a decision task that started fetching history is prioritized over a decision task that hasn't started fetching history. So, I think what that means is that the replayer should hold the resources until it fetches all existing history pages rather than releasing & re-acquiring for each page.

davidporter-id-au:
I would add a brief few lines about why this code is client-side and not server-side similar to how we discussed offline here

internal/pahlimiter/limiter_test.go
```diff
@@ -0,0 +1,329 @@
...
+		started := countWhileLeaking(l.Poll, fastTest)
+		assert.Equal(t, 100, started, "all requests should have been allowed")
+	})
+}
+
+func TestHistoryLimited(t *testing.T) {
+	t.Parallel()
+
+	t.Run("should not limit poll requests", func(t *testing.T) {
+		t.Parallel()
```
davidporter-id-au, line 57: is the duplicate use of `t.Parallel` here intentional?

internal/pahlimiter/limiter.go
```diff
@@ -0,0 +1,305 @@
+// Package pahlimiter contains a PollAndHistoryLimiter, used to share resources between polls and history loading,
+// to prevent flooding the server with history requests that will not complete in a reasonable time.
+package pahlimiter
```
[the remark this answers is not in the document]
Groxx: to make selects _flexible in quantity_ involves some rather annoying reflection (or codegen)... but yeah, I think this is relatively generalizable.  I'll think about it anyway, may take some fiddling to make something "good".

internal/pahlimiter/limiter.go
```diff
@@ -0,0 +1,305 @@
...
+		// only the first will have an effect.
+		// TODO: see if this is necessary... but it's easy and safe.
+		Poll(context.Context) (ok bool, done func())
+		// GetHistory will try to acquire a history-downloading resource,
+		// blocking until it succeeds or the context is canceled.
+		//
+		// The done func will release the resource - it will always be returned and can be called multiple times,
+		// only the first will have an effect.
+		// TODO: see if this is necessary... but it's easy and safe.
+		GetHistory(context.Context) (ok bool, done func())
```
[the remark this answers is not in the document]
Groxx: not entirely decided tbh.
- a trivial answer is "inside the get-history backoff callback", but that leaves substantial gaps for polls to occur.
- around the whole backoff callback allows us to consume this resource _during_ retries, which seems reasonable, though might lead to confusing metrics (pollers dropping without apparent cause, etc).
- around the whole "process a decision task" is potentially difficult, but would be the most-complete protection...  but I'm not confident that this would be _safe_, as we leak some workflow goroutines in unknown scenarios (I suspect corrupt histories or panics) and that could lead to permanently consumed resources.  probably desirable, but IMO too risky without some kind of detection.

internal/pahlimiter/limiter_test.go
Groxx: yep.  the "outer" one allows this top-level test to run concurrently with other top-level tests.  the "inner" ones allow each sub-test to run concurrently with tests inside this test.

both are needed to allow _all_ sub-tests in this file to run simultaneously.  e.g. remove the outer one and `TestHistoryLimited` will start, run all sub-tests, and complete them all before `TestUnlimited` is allowed to unblock its parallel tests.

PR #1224, opened 2023-03-09
Corrected error messages in getValidatedActivityOptions

jakobht:
**What changed?**
* Changed the error messages in getValidatedActivityOptions so they are correct
* Removed redundant line of code 
* Added tests for the new error messages

**Why?**
The error messages were slightly incorrect, the redundant line did not do anything

**How did you test it?**
I made new unit tests which checks the error messages

**Potential risks**
* The redundant line might have some strange side effect I didn't catch, but I'm pretty sure it's just a leftover from a refactor as it is the first thing called in the function called on the line below
* Changing the error message changes the interface, so if some users are checking against the strings their code could break.

### DIFF
--- a/internal/internal_activity.go
+++ b/internal/internal_activity.go
@@ -181,13 +181,13 @@ func getValidatedActivityOptions(ctx Context) (*activityOptions, error) {
 		p.TaskListName = p.OriginalTaskListName
 	}
 	if p.ScheduleToStartTimeoutSeconds <= 0 {
-		return nil, errors.New("missing or negative ScheduleToStartTimeoutSeconds")
+		return nil, errors.New("invalid non-positive ScheduleToStartTimeoutSeconds")
 	}
 	if p.StartToCloseTimeoutSeconds <= 0 {
-		return nil, errors.New("missing or negative StartToCloseTimeoutSeconds")
+		return nil, errors.New("invalid non-positive StartToCloseTimeoutSeconds")
 	}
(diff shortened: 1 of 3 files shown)


internal/internal_activity.go
```diff
@@ -181,13 +181,13 @@ func getValidatedActivityOptions(ctx Context) (*activityOptions, error) {
 		p.TaskListName = p.OriginalTaskListName
 	}
 	if p.ScheduleToStartTimeoutSeconds <= 0 {
-		return nil, errors.New("missing or negative ScheduleToStartTimeoutSeconds")
+		return nil, errors.New("invalid non-positive ScheduleToStartTimeoutSeconds")
```
dkrotx, line 184: I think you shouldn't change this. It's only ScheduleToCloseTimeoutSeconds' message that is incorrect.

internal/internal_activity.go
```diff
@@ -181,13 +181,13 @@ func getValidatedActivityOptions(ctx Context) (*activityOptions, error) {
...
 	}
 	if p.StartToCloseTimeoutSeconds <= 0 {
-		return nil, errors.New("missing or negative StartToCloseTimeoutSeconds")
+		return nil, errors.New("invalid non-positive StartToCloseTimeoutSeconds")
```
dkrotx, line 187: same as the above

dkrotx (approved):
Looks good

PR #1225, opened 2023-03-13
Merged the timeout logic for the tests in internal_workers_test.go

jakobht:
**What changed?**
I have merged the timeout logic for the tests in internal_workers_test, this raises the timeout to 10s for all tests, and treats timing out as a test failure.

**Why?**
The tests were flaky, sometimes timing out sometimes not, also it was hard to find the reason they failed as timing out was not treated as a failure.

**How did you test it?**
I ran all unit tests, this only changes the structure of the tests. 

**Potential risks**
This only changes the structure of the tests so it should not pose any risks.


internal/internal_workers_test.go
```diff
@@ -36,6 +32,9 @@ import (
 	"go.uber.org/yarpc"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
+	"sync/atomic"
```
dkrotx, line 35: Check your goimports/gofmt settings since the original version was correct - standard Go packages should go first.

internal/internal_workers_test.go
```diff
@@ -899,3 +877,17 @@ func (s *WorkersTestSuite) TestMultipleLocallyDispatchedActivity() {
 	s.True(activityResponseCompletedCount > 0)
 	s.True(activityCalledCount > 0)
 }
+
+// wait for test to complete - timeout and fail after 10 seconds to not block execution of other tests
+func startWorkerWaitForChannelAndTimeoutStopWorker(s *WorkersTestSuite, worker *aggregatedWorker, doneCh *chan struct{}) {
```
dkrotx, line 882: nit: just `startWorkerAndWait` would be OK. Currently the name captures **everything** that function does, which makes it too long.

dkrotx (approved):

PR #1228, opened 2023-03-14
Add more detail to how workflow.Now behaves

Groxx:
Docs are currently a bit vague on what time exactly it represents, which brings up questions like "is this the same as the cron schedule start time" or "what time does it represent if my workers are down for an hour".

So this description was gathered experimentally.
It may be a good idea to track exactly what updates it and when, but I suspect that the "recorded event" is a complete description.  Or it probably should be.


davidporter-id-au (approved):
<3

PR #1231, opened 2023-03-16
Added the Activity Registration required failure scenario to replayer test suite

agautam478:
**What changed?**
Added a new workflow and a test case in the replayer test suite.
The new workflow demonstrates how failing to register a new activity throws a non deterministic error.
This reinforces the requirement of RegisterActivityWithOption option in the replayer.

**Why?**
Made these changes to remember that the replayer requires a new feature.

**How did you test it?**
tested locally.

**Potential risks**
NA


test/replaytests/replay_test.go
```diff
@@ -53,6 +53,14 @@ func TestReplayChildWorkflowBugBackport(t *testing.T) {
 	require.NoError(t, err)
 }
 
+// Gives a non-deterministic-error because the getNameActivity3 was not registered on the replayer.
```
Groxx, line 56: `getNameActivity3` isn't in the greetings.json history or in the workflow code.  what's this testing?

agautam478: updated the documentation. this is testing the scenario where we get a non deterministic error if we don't register the activities in the replay test code.

Groxx (changes requested):
I've got some changes coming in another PR (not yet up, tackling other things first), but to follow those changes:
- no mixedCase files, use snake_case.  both because it's a Go convention and because it consistently avoids case-(in)sensitive-filesystem issues.
- make sure you actually run these, apparently they're not run by our `Makefile` :|  oops.

Groxx (approved):
sounds good for now 👍 
I've got a PR or two to come with some hopeful improvements for these, but let's get this merged in the meantime.

PR #1232, opened 2023-03-17
Expose activity registry on workflow replayer

Groxx:
This is necessary to use worker-instance-registered activities, as otherwise they are unrecognized and may not be able to replay successfully (as a changed activity name is a non-backwards-compatible change).

### DIFF
--- a/internal/workflow_replayer.go
+++ b/internal/workflow_replayer.go
@@ -26,6 +26,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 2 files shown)


test/replaytests/signatureBugWorkflow.go
```diff
@@ -1,3 +1,23 @@
+// Copyright (c) 2017-2021 Uber Technologies Inc.
```
Groxx, line 1: build failure because this was missing

davidporter-id-au (approved):

PR #1233, opened 2023-03-17
Switched to revive, goimports, re-formatted everything

Groxx:
Works pretty well!
This is using the latest versions of everything in internal/tools/go.mod (except thrift), and using the server's revive.toml as a base.

Revive is a bit noisy due to some `Id` -> `ID` recommendations, but we probably should actually do that for v2.


revive.toml
```diff
@@ -0,0 +1,104 @@
...
+[rule.dot-imports]
+[rule.error-naming]
+[rule.error-return]
+[rule.error-strings]
+[rule.errorf]
+# [rule.exported] # disabled due to lack of value / encouraging bad habits.  it complains about internal packages.
+[rule.if-return]
+[rule.increment-decrement]
+[rule.indent-error-flow]
+# [rule.package-comments] # disabled due to lack of value / encouraging bad habits.  it complains about internal packages.
```
davidporter-id-au, line 28: nit, comment refers to package exports, I think you meant requiring comments on packages

revive.toml
```diff
@@ -0,0 +1,104 @@
...
+#### disabled but maybe desirable
+
+# [rule.bare-return] # probably beneficial as it's slightly error-prone, but 2,000 failures
+# [rule.bool-literal-in-expr] # minor code simplifier, few failures
+# [rule.confusing-results] # maybe beneficial, only a few failures
+# [rule.deep-exit] # probably a good idea in most code, some failures, but not trivial to adopt
+# [rule.duplicated-imports] # minor, but may be worthwhile.  failures are weird but harmless
+# [rule.early-return] # minor code simplifier, a handful of failures
+# [rule.get-return] # existing failures are intentional + desirable, but in principle it's a fine idea
+# [rule.import-shadowing] # probably beneficial, but 750 failures
```
davidporter-id-au, line 83: :/ 

the shadowing rules continue to cause me pain over and over again. Hopefully one-day we can enable

mocks/WorkflowRun.go
```diff
@@ -21,9 +21,11 @@
 // Code generated by mockery v1.0.0. DO NOT EDIT.
 package mocks
 
-import context "context"
+import (
```
davidporter-id-au, line 24: did you re-run mocks?

davidporter-id-au (approved):

mocks/WorkflowRun.go
Groxx: goimports rewrites the mocks I think.  which should be fine, mockery just can't produce it correctly from scratch.

but I'll double check.

Groxx: yeah, it's just goimports.
#1239 may give us a standard way to generate them (though I'd like to remove them for v2), which will also give us an easy way to format them up-front so it's not forgotten

revive.toml
Groxx: definitely agreed.  it makes cmd-f a lot less useful too.

revive.toml
Groxx: both complain about internals, e.g. package-comments has this:
```
internal/common/backoff/retry.go
  (21, 1)  https://revive.run/r#package-comments  should have a package comment
```

I'd be game to enable package-comments tbh, I think they have frequent-enough value even in internals.  There's just like 15 of these at the moment and I don't think we particularly care.

PR #1234, opened 2023-03-18
Shift replayer to prefer io.Reader rather than filenames

Groxx:
This allows us to move to `//go:embed` or use hard-coded strings (or any other data-source).
Short-term plans are to use go:embed to read files once / not require runtime access to files, only build.


davidporter-id-au (approved):
nice!

PR #1235, opened 2023-03-23
[error] surface more fields in ContinueAsNew error

shijiesheng:
**What changed?**
To start a new workflow correctly, we actually need more fields:
1. serialized input
2. header info

**Potential risks**
 low risk to add NEW receive methods on public struct


Groxx (approved):
nice and easy.

PR #1236, opened 2023-03-23
Add the change in branch number case(test) to replayersuite

agautam478:
**What changed?**
- Added the workflow that creates an x number of branched workflows based on a passed parameter to replayer suite.
- Added two test cases demonstrating the change in number of spawned branched.

**Why?**
- These changes were added as a part of test cases build to test replayer.

**How did you test it?**
yes, locally

**Potential risks**
NA


Groxx:
Since these don't have branches, nor do they have arguments that change the number of branches:
what does this PR do?

agautam478:
> Merge

If the number of expected branches is changed the replayer should catch it as a non deterministic error. `TestBranchWorkflowWithExtraBranch` tests this case. 

sampleBranchWorkflow2 has the logic where an extra branch in added in :

for i := 1; i <= 4; i++ {
		activityInput := fmt.Sprintf("branch %d of 4", i)
		future := workflow.ExecuteActivity(ctx, sampleActivity, activityInput)
		futures = append(futures, future)
	}
	
The value was 3 when the history was recorded. Refer to https://docs.google.com/document/d/1lwADEFAkaxOiu0bbNXeMmdrPJYCnOBXyw8VJ-aJNqzM/edit
point 1.

Groxx:
ah, right, this is the "branch" workflow from samples, which... also doesn't really have branches 🤦 

If we wanted to make this one clearer about the difference, we could register a more dynamic workflow instead.  e.g.
```go
func makeworkflow(count) func(workflow.Context) error {
  return func(workflow.Context) error {
    // ...
    for i := 0; i < count; i++ {
    // ...
  }
}

// and in the test:
replayer.RegisterWorkflowWithOptions(makeworkflow(3), ...) // this one should work
replayer.RegisterWorkflowWithOptions(makeworkflow(4), ...) // this one should fail
```
since that way it's clear what changed, which is the only relevant part of the test.

Groxx:
alrighty.  given the similarity to the sample terminology and code, how about:
- let me know if you want to try the cleanup.  it's fine if no.
- this doesn't build, try it locally
- let's start using more specific error checks.  "an error" doesn't assert "non-determinism error", but checking error strings can.  (ideally this would be an `errors.As` check, but these errors are not yet typed.  that'll be a v2(+?) thing probably.)

Groxx (approved):

PR #1237, opened 2023-03-23
Add and emulate the issues found in the workflows involving coroutines into the replayersuite.

agautam478:
**What changed?**
- Added a sample workflow executes multiple branches in parallel using workflow.Go() method.
- Added new tests which run the workflow with coroutines.
- The test that should have failed still passes.

**Why?**
To record the cases missed by the replayer.

**How did you test it?**
yes, locally

**Potential risks**
NA


Groxx (approved):
It would probably be good to clean these up so it's clear what's relevant, but iirc these are essentially copies from the samples repo and they're straightforward enough 👍

PR #1238, opened 2023-03-23
Add the case where changing the activities (addition/subtraction/modification in current behavior) in the switch case has no effect on replayer.

agautam478:
**What changed?**
- Added a workflow that incorporates a switch case on the basis of a randomly generated key.
- The history recorded the activity for only one of the cases.
- The replayer seems to not care about any of the activities that were executed.
- Even if you execute an activity that's not registered in the history or register a different activity on the replayer the test passes.
- The workflow not waiting for the activity so it doesn't notice the "called an unregistered activity" case.
- The history has activity scheduled but it's never noticed because the workflow ends with a finished status.

**Why?**
- This is one of the cases that requires investigation into the capabilities of replayer.

**How did you test it?**
yes, locally

**Potential risks**
NA


test/replaytests/replay_test.go
```diff
@@ -98,3 +98,14 @@ func TestGreetingsWorkflow2(t *testing.T) {
...
 	require.Error(t, err)
 }
+
+// this test registers Cherry Activity as the activity but executes a random Activity in the workflow. Ideally replayer doesn't concerns itself with the change in the
+// activity content until it matches the expected output type.History has recorded the output of banana activity instead.
+func TestExclusiveChoiceWorkflow(t *testing.T) {
+	replayer := worker.NewWorkflowReplayer()
+
+	replayer.RegisterWorkflowWithOptions(exclusiveChoiceWorkflow, workflow.RegisterOptions{Name: "choice"})
+	activity.RegisterWithOptions(orderCherryActivity, activity.RegisterOptions{Name: "testactivity", DisableAlreadyRegisteredCheck: true})
```
Groxx, line 108: not sure what the goal is with this registration / comment on the test tbh.
registering cherry as an unused activity / activity name should be fine, since it executes `banana`.  it's the same as "there is another activity registered", which is true for all of these tests since there are like 10 activities registered.

agautam478: the order selection change also makes no difference. I think I will add more details in this test case.

When Registered Activity was Cherry but order selection was Banana the test passes.

Groxx: if you mean the order of the cases in the `switch` statement, that's expected - they're exclusive, so order doesn't matter.

which activity is registered also shouldn't matter... normally, as long as the executed one is also registered.

---

Surprisingly the activity is scheduled:
https://github.com/uber-go/cadence-client/pull/1238/files#diff-ef2faa6c08bffac1dce928695bd84dabd73d4f3b8ae614c0be7cfa9d69b0a637R155-R164

Even though there's no blocking call, so it would never be started:
https://github.com/uber-go/cadence-client/pull/1238/files#diff-0e44a72223317107830eacc47b55840ac19e84f35d710475bab01add8e3a6e05R67-R81

so it kinda feels like:
1. the workflow is just irrational, it's not waiting for the activity so it doesn't notice the "called an unregistered activity" case.
    - looks like the sample code is the same: https://github.com/uber-common/cadence-samples/blob/ce920bd6e917aec5fab3c66057a37a51b23c51ef/cmd/samples/recipes/choice/exclusive_choice_workflow.go#L47
    - tbh this probably is worth a test.  this one could work, but a dramatically simpler one would be possible too.
2. so this is currently falling into the "both finished, so it didn't notice the history changed" bug, though for a possibly-different reason (didn't wait for the activity).
    - I'm not sure if this is possible to test separately from 1... but a test does seem useful, yea.  it's sufficiently different from some of the other causes.

Groxx: so I guess: clarify the goal of the commit and test, and we can include it?  it's showing misbehavior, but the reason doesn't match anything that's written down.

agautam478: I racked my brain to figure out how to represent it best but couldn't come up with anything better than simply registering a different activity within the test case. 

currently it looks like the replayer is not bothered by anything that happens inside the switch case.

test/replaytests/exclusive_choice_workflow.go
```diff
@@ -0,0 +1,126 @@
...
+
+func getOrderActivity() (string, error) {
+	fmt.Printf("Order is for Cherry")
+	return "cherry", nil
+}
+
+func getAppleOrderActivity() (string, error) {
+	fmt.Printf("Order is for Cherry")
+	return "apple", nil
+}
```
Groxx, line 111: since they're the same `func() (string, error)` signature, you only actually need one of these.  but both is fine for easier recording / to show history differences I think.

but:
- getAppleOrderActivity's result isn't actually used anywhere (there's only one history, and it contains banana)
- fmt.printf is the wrong string :)

Groxx:
double check stuff, there's quite a lot out of sync with code/comments/history/etc :)

test/replaytests/replay_test.go
```diff
@@ -99,6 +99,33 @@ func TestGreetingsWorkflow2(t *testing.T) {
...
+// History has recorded the output of banana activity instead. Here, The workflow is not waiting for the activity so it doesn't notice
+// that activity id different from the one recorded in history.
+func TestExclusiveChoiceWorkflowWithUnregisteredActivity(t *testing.T) {
+	replayer := worker.NewWorkflowReplayer()
+
+	replayer.RegisterWorkflowWithOptions(exclusiveChoiceWorkflow, workflow.RegisterOptions{Name: "choice"})
+	replayer.RegisterActivityWithOptions(orderCherryActivity, activity.RegisterOptions{Name: "testactivity", DisableAlreadyRegisteredCheck: true})
+	err := replayer.ReplayWorkflowHistoryFromJSONFile(zaptest.NewLogger(t), "choice.json")
+	require.NoError(t, err)
+}
```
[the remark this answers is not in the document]
agautam478: getOrderActivity is the default activity that was recorded in the History so we don't need to register is separately I think.
Yes, the recorded history is banana but we are registering cherry.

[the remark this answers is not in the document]
Groxx: > Yes, the recorded history is banana but we are registering cherry.

and

> and executes the same activity in the workflow.

disagree, because it's executing banana.

---

re registering separately: given other tests / etc, I think we kinda have two options for how we think replayer should work, and the tests are currently claiming inconsistent things:
1. we should make sure activities are registered, and that the workflow calls the same activities.
    - in this case, the `getOrderActivity` call should fail.  currently it succeeds.
2. we shouldn't care, because it doesn't actually matter for _replaying a workflow_.  the data can still be read, because it has nothing to do with the activity, only the `.Get(ctx, &out)` out-param's type.
    - in this case, `TestGreetingsWorkflowforActivity` should be fine because it's not registered, but currently it fails.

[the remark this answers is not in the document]
Groxx: I think we agree that it should be 1, but in that case these tests should be "correct" up to the point that they're not.  otherwise there's N things that are wrong and they're claiming it's Z when it could / should be X or Y which happened earlier.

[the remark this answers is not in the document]
Groxx: (I think some of this is also that I'm getting confused by the blend of "test passes but it should not" when I'm rereading, so I'll double check all this later with a fresh brain and see if I'm missing something.)

test/replaytests/replay_test.go
```diff
@@ -100,6 +100,34 @@ func TestGreetingsWorkflow2(t *testing.T) {
...
+
+	replayer.RegisterWorkflowWithOptions(exclusiveChoiceWorkflow, workflow.RegisterOptions{Name: "choice"})
+	err := replayer.ReplayWorkflowHistoryFromJSONFile(zaptest.NewLogger(t), "choice.json")
+	require.NoError(t, err)
+}
+
+// This test registers Cherry Activity as the activity but calls Apple activity in the workflow code. Infact, Cherry and Banana
+// activities are not even a part of the workflow code in question.
+// History has recorded the output of banana activity. Here, The workflow is not waiting for the activity so it doesn't notice
+// that registered activity is different from executed activity and the activity name has changed in case of getAppleOrderActivity.
```
Groxx, line 118: > activity name has changed in case of getAppleOrderActivity
^ not relevant since it's registered as `main.GetOrderActivity`

Groxx (approved):
yea, let's do it.
there's some cleanup to do in this whole folder soonish I think, but the tests are useful for tracking fixes to replayer/shadower in the meantime :)

PR #1240, opened 2023-03-26
Locally-dispatched activity test flakiness hopefully resolved

Groxx:
At least, I finally ran it more than 5x in a row without failure.  yay!
We should probably just ban bare atomics.

There are occasionally test races between test-completion and zap logging,
but those are sorta un-resolvable unless we reliably stop everything in
tests.  Which would be excellent, but I think we're fairly far from
achieving that at the moment.

### DIFF
--- a/internal/internal_workers_test.go
+++ b/internal/internal_workers_test.go
@@ -23,13 +23,13 @@ package internal
 
 import (
 	"context"
-	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/golang/mock/gomock"
 	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/suite"
+	"go.uber.org/atomic"
 	"go.uber.org/cadence/.gen/go/cadence/workflowservicetest"
 	m "go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/internal/common"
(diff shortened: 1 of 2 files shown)


davidporter-id-au (approved):
Can't fault any of this, a more general question for racy-tests, should we be running more than a single iteration?

Groxx:
racy tests: if you mean to let the race detector catch things, it's mostly unlikely to catch anything it's not already catching.

if you mean to reduce flakiness: possibly, though I'm not sure if this fully resolves the flakiness in this.  it did change it from "almost always fails" to "passed multiple times in a row" though, and these were clearly incorrect, so I think it's a true _improvement_ in any case.

Groxx:
> oh yes, this is a strict fix, I was just wondering in the general case if we should increase certainty where there was a lot of concurrency involved, but, as you point out, that's what the race-detector does and I just forgot about it

yea.  and I don't think we have data races in tests except for some cases where code does not shut down in time, and it produces a log after the test has completed (we have incorrect shutdowns _all over_, which is why a lot of tests can't be converted to use `zaptest.NewLogger(t)`).  or at least all the ones I've seen have been just ^ that logger-race-due-to-shutdown.

PR #1247, opened 2023-04-27
Test-suite bugfix: local activity errors were not encoded correctly

Groxx:
The added test hopefully reveals the issue - prior to this fix,
that `err := f.Get(...)` -> `errors.Is(err, ...)` check passed, because
the error value was not ever encoded into a `*GenericError`.

Thankfully this was only a test environment bug, and a user discovered
it due to writing tests.

---

The non-test behavior can be seen split between these two locations:
- encoding error reason / data: https://github.com/uber-go/cadence-client/blob/ad91fc718330cb40ff26627fa1d1a6c78c2ceaca/internal/internal_event_handlers.go#L1159-L1163
- constructing the generic error in handleLocalActivityMarker: https://github.com/uber-go/cadence-client/blob/ad91fc718330cb40ff26627fa1d1a6c78c2ceaca/internal/internal_event_handlers.go#L1132-L1135

... which also show incorrect / pointless string/byte conversions.
I've left comments there just to prevent confusion, but these fields are
probably worth changing at some point.

### DIFF
--- a/internal/internal_event_handlers.go
+++ b/internal/internal_event_handlers.go
@@ -1159,10 +1159,10 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessLocalActivityResult(lar *lo
 	if lar.err != nil {
 		errReason, errDetails := getErrorDetails(lar.err, weh.GetDataConverter())
 		lamd.ErrReason = errReason
-		lamd.ErrJSON = string(errDetails)
+		lamd.ErrJSON = string(errDetails) // TODO: this is not json, nor is it necessarily a valid string
 		lamd.Backoff = lar.backoff
 	} else {
-		lamd.ResultJSON = string(lar.result)
+		lamd.ResultJSON = string(lar.result) // TODO: same, not json nor a string
 	}
 
 	// encode marker data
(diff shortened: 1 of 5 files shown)


internal/internal_event_handlers.go
```diff
@@ -1159,10 +1159,10 @@ func (weh *workflowExecutionEventHandlerImpl) ProcessLocalActivityResult(lar *lo
 	if lar.err != nil {
 		errReason, errDetails := getErrorDetails(lar.err, weh.GetDataConverter())
 		lamd.ErrReason = errReason
-		lamd.ErrJSON = string(errDetails)
+		lamd.ErrJSON = string(errDetails) // TODO: this is not json, nor is it necessarily a valid string
```
davidporter-id-au, line 1162: clarification, not *validated* as being json or even a string? or *never* json?

davidporter-id-au (approved):

internal/internal_event_handlers.go
Groxx: it's arbitrary user-encoded data (because it uses their dataconverter).

json is the default, but since it's arbitrary IMO that means "never json".

PR #1252, opened 2023-05-31
Adding in additional header to determine a more stable isolation-group

davidporter-id-au:
**What changed?**

- Adds a new header for downstream which is expected to be more stable to determine the origin of the worker
- Repurposes the auth / identity wrapper a bit to shim this in there

**Why?**

**How did you test it?**

**Potential risks**


internal/client.go
```diff
@@ -578,6 +581,7 @@ func NewClient(service workflowserviceclient.Interface, domain string, options *
 	if options != nil && options.Authorization != nil {
 		service = auth.NewWorkflowServiceWrapper(service, options.Authorization)
 	}
+	service = isolationgroup.NewWorkflowServiceWrapper(service, options.IsolationGroup)
```
Shaddoll, line 584: check if `options.IsolationGroup` is empty string?

internal/common/isolationgroup/service_wrapper.go
```diff
@@ -0,0 +1,294 @@
...
+	"context"
+
+	"go.uber.org/yarpc"
+
+	"go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
+	"go.uber.org/cadence/.gen/go/shared"
+)
+
+const (
+	isolationGroupHeader = "isolation-group"
```
Shaddoll, line 33: to make it not conflict with some random headers, consider prefixing it with `cadence-` or `cadence-client-`

internal/common/isolationgroup/service_wrapper.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (w *workflowServiceIsolationGroupWrapper) RespondQueryTaskCompleted(ctx context.Context, request *shared.RespondQueryTaskCompletedRequest, opts ...yarpc.CallOption) error {
+	opts = append(opts, w.getIsolationGroupIdentifier())
+	err := w.service.RespondQueryTaskCompleted(ctx, request, opts...)
+	return err
+}
+
+func (w *workflowServiceIsolationGroupWrapper) GetSearchAttributes(ctx context.Context, opts ...yarpc.CallOption) (*shared.GetSearchAttributesResponse, error) {
+	result, err := w.service.GetSearchAttributes(ctx, opts...)
```
Shaddoll, line 263: header is missing

internal/client.go
davidporter-id-au: good call

internal/client.go
```diff
@@ -26,6 +26,8 @@ import (
 	"fmt"
 	"time"
 
+	"go.uber.org/cadence/internal/common/isolationgroup"
```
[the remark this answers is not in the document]
davidporter-id-au: I tried to, not entirely sure if/if there's still a format issue

internal/common/isolationgroup/service_wrapper_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+package isolationgroup
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/suite"
```
Groxx, line 29: tbh I think we should just never write any new code that touches this :\  it's quite easy to avoid it for new stuff.

internal/client.go
```diff
@@ -578,6 +581,9 @@ func NewClient(service workflowserviceclient.Interface, domain string, options *
 	if options != nil && options.Authorization != nil {
 		service = auth.NewWorkflowServiceWrapper(service, options.Authorization)
 	}
+	if options.IsolationGroup != "" {
```
Groxx, line 584: check for nil, since it's exposed + was allowed before.  plus there's not really any reason to require people to pass `&ClientOptions{}` everywhere.

Groxx, line 584: huh.  I'd say keep allowing nil, since it's fine if it's empty... but I think this is entirely internal, so we _could_ start requiring it 🤔.  we'd just have to pass an empty value when null in the public API, which is easy.

if we're going to start requiring it though, there's a lot of other `options != nil` checks that can be removed, since they'll just panic here if they aren't true.

internal/common/isolationgroup/service_wrapper.go
```diff
@@ -0,0 +1,295 @@
...
+	return yarpc.WithHeader(isolationGroupHeader, w.isolationGroupIdentifier)
+}
+
+func (w *workflowServiceIsolationGroupWrapper) DeprecateDomain(ctx context.Context, request *shared.DeprecateDomainRequest, opts ...yarpc.CallOption) error {
+	opts = append(opts, w.getIsolationGroupIdentifier())
+	return w.service.DeprecateDomain(ctx, request, opts...)
+}
+
+func (w *workflowServiceIsolationGroupWrapper) ListDomains(ctx context.Context, request *shared.ListDomainsRequest, opts ...yarpc.CallOption) (*shared.ListDomainsResponse, error) {
+	opts = append(opts, w.getIsolationGroupIdentifier())
```
Groxx, line 59: given how regular this pattern is, might be worth adopting gowrap to reduce minor mistakes now and in maintenance.

we already have it in the server, it'd be fairly easy to add here: https://github.com/uber/cadence/blob/e2a2a940030ec62a0749a4f4d8c86100305220c2/client/frontend/interface.go#L32

evictiontest/workflow_cache_eviction_test.go
```diff
@@ -176,6 +177,7 @@ func (s *CacheEvictionSuite) TestResetStickyOnEviction() {
 	workflowWorker := internal.NewWorker(s.service, "test-domain", "tasklist", worker.Options{
 		DisableActivityWorker: true,
 		Logger:                zaptest.NewLogger(s.T()),
+		IsolationGroup:        "zone-1",
```
Groxx, line 180: is this necessary for this test, or just adding a value to have it in a test somewhere?

internal/internal_worker_test.go
```diff
@@ -445,7 +445,7 @@ func (s *internalWorkerTestSuite) testCompleteActivityHelper(opt *ClientOptions)
 }
 
 func (s *internalWorkerTestSuite) TestCompleteActivity() {
-	s.testCompleteActivityHelper(nil)
+	s.testCompleteActivityHelper(&ClientOptions{IsolationGroup: ""})
```
Groxx, line 448: should probably have a value or not be a field at all, since this is indistinguishable from `&ClientOptions{}`

Groxx (commented):
broadly: might be some stuff to clean up, but structurally sound I think.  we already have wrappers, one more for another optional feature seems fine.

I'm not all that sold on these wrappers to begin with, but they're already here ¯\\\_(ツ)\_/¯

internal/common/isolationgroup/service_wrapper_test.go
davidporter-id-au: Good call, I also sloppily copied and pasted some errors, let me fix it up

internal/internal_worker_test.go
davidporter-id-au: I think I just added it beccause some tests were using methods on the options and panicing

evictiontest/workflow_cache_eviction_test.go
davidporter-id-au: the wrapper code-path is checking if this field is populated before using the additional wrapper:
```
	if options.IsolationGroup != "" {
		service = isolationgroup.NewWorkflowServiceWrapper(service, options.IsolationGroup)
	}

```
. So I figured to give a more realistic test in a prod-ish env we can excecise this additional codepath

internal/client.go
davidporter-id-au: Not entirely sure I follow, I don't feel strongly about this check though, but I'm pretty sure we'll never require this field.  It's unlikely most clients will care about this feature.

davidporter-id-au:
> broadly: might be some stuff to clean up, but structurally sound I think. we already have wrappers, one more for another optional feature seems fine.
> 
> I'm not all that sold on these wrappers to begin with, but they're already here ¯\_(ツ)_/¯

I'm not going to investigate codegen in this PR, though it certainly feel's like it should have it at this point. Cleaned up the test though

internal/client.go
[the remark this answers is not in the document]
3vilhamster: Unfortunately, neither go imports nor go fmt will not join import groups, so if you have them separated by a new line, they will still be in separate groups. Also, Goland's default behavior is to add imports in a new group separating with a new line, which is unfortunate. I think monorepo folks have recently introduced a change that should enforce imports to follow Go Uber style guide. We could adjust later on.

internal/client.go
Groxx: all other lines here include
```go
if options != nil && options.Authorization != nil {
   ^ this nil check
```
which allows options to be nil.

this one doesn't, so either all those != nil checks are unnecessary because we always pass options, or it's changing the function to require options all the time.

so I'd add the nil check to this one too.  or clean up if it actually is always passed.

davidporter-id-au: oh 🤦  sorry, good catch

Shaddoll (approved):

internal/client.go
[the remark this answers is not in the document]
Groxx: goimports _might_ have a way to force that 🤔.  if it does, I'd totally be in favor of it - special-case imports can still be commented, and if we have order-dependent imports then I'm happy to say "absolutely not, remove those init funcs and replace them with something sane"

Groxx (approved):
yea, looks good.  also 👍 for that name change, `-zone` is too context-specific and doesn't match the rest of the code.

Issue #1255, opened 2023-06-12
Implicit registration for Activities and Workflow when mocking them in the testsuite


edmondop:
**Is your feature request related to a problem? Please describe.**

When mocking an Activity or a Workflow using the `TestWorkflowEnvironment`, no-one is interested in mocking Workflows or Activities that will not be executed! And to be executed, they need to be registered. So why don't register them upon mocking, and ignore a double registration problem?

**Proposed Solution**

Registering the activities and workflows when mocking them, and ignore if they are already registered. 

**Additional context**

PR #1256, opened 2023-06-12
Making Workflow and Activity registration optional when they are mocked

edmondop:
**What changed?**

When using the TestSuite and mocking Workflow and Activities, they are registered if they weren't registered before

**Why?**

Solves #1255 

**How did you test it?**

**Potential risks**


agautam478 (approved):

edmondop:
@agautam478 I think we can mark closed the #1255

Issue #1125, opened 2021-09-08
`TestWorkflowEnvironment` doesn't invoke interceptors in the context of child workflows


seiffert:
**Describe the bug**
`TestWorkflowEnvironment`s with custom interceptors don't seem to execute them on child workflows at the moment. I can see the `ExecuteWorkflow` hook of my interceptor being called for the tested main workflow, but if that workflow calls `workflow.ExecuteChildWorkflow` internally, the interceptor's `ExecuteWorkflow` hook isn't called in the context of the child workflow. 

In production code, I can observe how when calling `ExecuteChildWorkflow` the identically-named interceptor method is called and then once the child workflow was scheduled, the interceptor's `ExecuteWorkflow` is called in the context of the child workflow.

**To Reproduce**
Is the issue reproducible? **Yes**

**Steps to reproduce the behavior:**
* Copy `cadence_test.go` from [this Gist](https://gist.github.com/seiffert/b4cae0bcea3658a7c833c19f23a22b6b) to a file.
* Run the test suite `go test -v . -run=TestInterceptorTestSuite`
* Check whether the output contains the lines `start Interceptor.ExecuteWorkflow "child"` and `end Interceptor.ExecuteWorkflow "child"`.

**Expected behavior**
Workflow Interceptors' `ExecuteWorkflow` method is also invoked for child-workflows.

**Additional context**
* [Gist with test case and the expected vs actual output.](https://gist.github.com/seiffert/b4cae0bcea3658a7c833c19f23a22b6b)

PR #1257, opened 2023-06-16
ISSUE 1125: Addressing difference in workflow interceptors when using the testkit

edmondop:
**What changed?**

As identified by https://github.com/uber-go/cadence-client/issues/1125, a bug in the workflow testsuite resulted in a different behavior when running workflows using the real worker (the `ExecuteWorkflow` interceptor would be executed both on parent and children workflow) and the testsuite.

**Why?**

**How did you test it?**

See unit test

**Potential risks**


agautam478 (approved):

edmondop:
@agautam478 I think we can close #1125

PR #1259, opened 2023-06-21
Adding WorkflowType to "Workflow panic" log-message

dkrotx:
It is very expected to have this field, but customers have to walk
through the stack trace, which is not always convinient especially with
wrappers.

**What changed?**
WorkflowType is logged among other fields in "Workflow panic" log-message.

**Why?**
Seeing this field makes debugging easier.
Especially when you have a log-view (Like Kibana or uMonitor) - you can add WorkflowType field as a column.

**How did you test it?**
Verified by writing a workflow that panics + with the unit-test

**Potential risks**
No risks - retrieving workflow-name is lock-free, and at least returns an empty string.


Groxx (approved):
yea - simple enough.
are there other fields we should consider adding?  e.g. tasklist is probably available too, and any others that are commonly added

PR #1260, opened 2023-06-26
Emit cadence worker's hardware utilization inside worker once per host

timl3136:
**What changed?**
We can now emit cadence worker's hardware utilization while polling for decision/activity task

**Why?**
This enable us to monitor worker's utilization

**How did you test it?**
Tested on dev environment

**Potential risks**
No risk since it does not affect overall logic and are just added read-only functions


internal/internal_task_pollers.go
```diff
@@ -1369,3 +1376,21 @@ func convertActivityResultToRespondRequestByID(identity, domain, workflowID, run
 		Details:    details,
 		Identity:   common.StringPtr(identity)}
 }
+
+func emitHardwareMetricsTaggedScope(scope *metrics.TaggedScope) {
+	if scope != nil {
+		cpuPercent, _ := cpu.Percent(0, false)
+		cpuCores, _ := cpu.Counts(false)
+
+		scope.Gauge(metrics.NumCPUCores).Update(float64(cpuCores))
```
davidporter-id-au, line 1385: Won't a guage get overridden unless it has a host-level dimension? If it does have a host-level dimension won't the cardinality be too much?

davidporter-id-au: @allenchen2244

timl3136: It is on datacenter-level dimension, I am currently looking at method to make it host-level

internal/internal_worker_base.go
```diff
@@ -209,6 +215,9 @@ func (bw *baseWorker) Start() {
 	bw.shutdownWG.Add(1)
 	go bw.runTaskDispatcher()
 
+	bw.shutdownWG.Add(1)
+	bw.EmitHardwareUsage()
```
timl3136, line 219: Not sure if that's the best practice of using waitgroup and go routine (since the go routine is started inside this function) but the test is not reporting and goleak.

internal/internal_worker_base.go
```diff
@@ -401,3 +410,38 @@ func (bw *baseWorker) Stop() {
 	}
 	return
 }
+
+func (bw *baseWorker) EmitHardwareUsage() {
+	defer bw.shutdownWG.Done()
+
+	emitOnce.Do(func() {
+		go func() {
+			ticker := time.NewTicker(hardwareEmitInterval)
```
davidporter-id-au, line 419: This needs a panic recover https://go.dev/blog/defer-panic-and-recover

internal/internal_worker_base.go
```diff
@@ -401,3 +410,38 @@ func (bw *baseWorker) Stop() {
 	}
 	return
 }
+
+func (bw *baseWorker) EmitHardwareUsage() {
+	defer bw.shutdownWG.Done()
+
+	emitOnce.Do(func() {
```
davidporter-id-au, line 417: Mind adding a comment as to the sync.Once ussage? I am guessing it's because you don't want duplicated metrics but that's kinda non-obvious to the casual reader

internal/internal_worker_base.go
```diff
@@ -401,3 +410,38 @@ func (bw *baseWorker) Stop() {
 	}
 	return
 }
+
+func (bw *baseWorker) EmitHardwareUsage() {
+	defer bw.shutdownWG.Done()
```
davidporter-id-au, line 415: I think this needs to be in the goroutine

timl3136: wrapped entire function in a goroutine now

internal/internal_worker_base.go
timl3136: Done

internal/internal_worker_base.go
timl3136: Added recover

internal/internal_worker_base.go
```diff
@@ -27,7 +27,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/shirou/gopsutil/cpu"
```
3vilhamster, line 30: Standart imports should be separated from external. I think goimports should do the trick

internal/internal_worker_base.go
```diff
@@ -54,6 +57,8 @@ var (
 
 var errShutdown = errors.New("worker shutting down")
 
+var emitOnce sync.Once
```
3vilhamster, line 60: I would suggest extracting this into WorkerResult as a separate dependency, instead of trying to fit into the workers with sync.Once.

timl3136: What do you mean by WorkerResult? Are you referring to the baseWorker struct?

3vilhamster: Oh well, I thought we provided fx.Module for this as we do in the monorepo. There is a NewCadenceWorkers that returns a single result that can be a good place to have this component.
If we want to care about monorepo - then it should be part of monorepo.
If we want to have this in OSS, then, I think, it should be a separated component that is optional and could be used by developers as an extra option.

3vilhamster: I'm not even sure that we want to have this in OSS, since a developer could use something like https://github.com/prometheus/client_golang/blob/main/prometheus/go_collector.go#L216
So it will expose data for collection.

3vilhamster:
I'm not sure I understand what you mean by worker utilization. If we track cpu/memory, then it should be dominated by a workload that is unrelated to the worker itself.
What is exactly we want to achieve here?

internal/internal_worker_base.go
shijiesheng: If we want to do it in OSS, we cannot use the package that only works for prometheus. Cadence use tally on top of underlying metrics package.

internal/internal_worker_base.go
```diff
@@ -401,3 +412,43 @@ func (bw *baseWorker) Stop() {
 	}
 	return
 }
+
+func (bw *baseWorker) EmitHardwareUsage() {
```
shijiesheng, line 416: nit: private method

internal/internal_worker_base.go
```diff
@@ -401,3 +412,43 @@ func (bw *baseWorker) Stop() {
...
+	defer bw.shutdownWG.Done()
+	// We want the emit function run once per host instead of run once per worker
```
shijiesheng, line 418: move the comment to `emitOnce.Do()`?

internal/internal_worker_base.go
```diff
@@ -401,3 +412,43 @@ func (bw *baseWorker) Stop() {
...
+	ticker := time.NewTicker(hardwareEmitInterval)
+	defer func() {
```
shijiesheng, line 420: nit: move recover to the top

internal/internal_worker_base.go
```diff
@@ -46,6 +48,7 @@ import (
 const (
 	retryPollOperationInitialInterval = 20 * time.Millisecond
 	retryPollOperationMaxInterval     = 10 * time.Second
+	hardwareEmitInterval              = 30 * time.Second
```
shijiesheng, line 51: nit: hardwareMetricsCollectInterval

internal/internal_worker_base.go
```diff
@@ -401,3 +412,43 @@ func (bw *baseWorker) Stop() {
...
+		}
+	}()
+	for {
+		select {
+		case <-bw.shutdownCh:
+			ticker.Stop()
+			return
+		case <-ticker.C:
+			host := bw.options.host
+			scope := bw.metricsScope.Tagged(map[string]string{"host": host})
```
shijiesheng, line 437: add a `hostTag` constant like `clientVersionTag`

go.mod
```diff
@@ -12,11 +12,14 @@ require (
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/pborman/uuid v0.0.0-20160209185913-a97ce2ca70fa
 	github.com/robfig/cron v1.2.0
+	github.com/shirou/gopsutil v3.21.11+incompatible
```
shijiesheng, line 15: They said 3.21.11 https://github.com/shirou/gopsutil/releases/tag/v3.22.12  has some issues so they reverted it. Could you verify that?

internal/internal_worker_base.go
```diff
@@ -401,3 +413,42 @@ func (bw *baseWorker) Stop() {
...
+			return
+		case <-ticker.C:
+			host := bw.options.host
+			scope := bw.metricsScope.Tagged(map[string]string{clientHostTag: host})
+
+			cpuPercent, _ := cpu.Percent(0, false)
+			cpuCores, _ := cpu.Counts(false)
+
+			scope.Gauge(metrics.NumCPUCores).Update(float64(cpuCores))
+			scope.Gauge(metrics.CPUPercentage).Update(cpuPercent[0])
```
shijiesheng, line 443: we might still have a nil pointer if cpu.Percent(0, false) returns error. Either need to handle error or do a length check.

internal/internal_worker_base.go
```diff
@@ -209,6 +215,12 @@ func (bw *baseWorker) Start() {
 	bw.shutdownWG.Add(1)
 	go bw.runTaskDispatcher()
 
+	// We want the emit function run once per host instead of run once per worker
+	emitOnce.Do(func() {
```
shijiesheng, line 219: how about move emitOnce to `bw.emitHardwareUsage`

timl3136: I have tried that, but that results in go routine leak unfortunately

internal/internal_worker.go
```diff
@@ -311,6 +312,9 @@ func newWorkflowTaskWorkerInternal(
 	// 3) the result pushed to laTunnel will be send as task to workflow worker to process.
 	worker.taskQueueCh = laTunnel.resultCh
 
+	worker.options.host = params.Host
```
shijiesheng, line 315: you might just put it into `baseWorkerOptions`

shijiesheng (approved):

PR #1263, opened 2023-08-01
We now log when the automatic heartbeating fails

jakobht:
**What changed?**
We now log when the automatic heart beating fails

We just get unhelpful errors from YARPC that something went wrong in the call, but no details, this adds details 

Ran the unit tests 

I don't see any risks


3vilhamster (approved):
LGTM

PR #1269, opened 2023-08-30
Add a structured error for non-determinism failures

Groxx:
This is being added for a few reasons:

1. it's useful information, and an unstructured error has no way to safely communicate it
2. previously these were un-typed errors, so this custom type can be added without breaking backwards compatibility
3. we want to use it to expose more information about shadow-test failures, which are currently far more opaque than they need to be

Shadow tests are opaque for a lot of reasons that need to be improved, but enhancing the error is a pretty straightforward first step.  And it seems useful in general.

This intentionally does *not* change the `.Error()` text, largely just because I didn't see any reason to do so.  The error strings are reasonably well-known and there doesn't seem to be any major benefit that can be gained by improving them... currently.
That said, these strings *must not* be considered stable in general, and we should consider changing them in the future.


internal/client.go
```diff
@@ -25,7 +25,6 @@ import (
 	"context"
 	"fmt"
 	"time"
-	"go.uber.org/cadence/internal/common/isolationgroup"
```
Groxx, line 28: `make fmt` was apparently missed earlier

internal/common/isolationgroup/service_wrapper_test.go
```diff
@@ -22,15 +22,15 @@ package isolationgroup
 
 import (
 	"context"
-	"github.com/stretchr/testify/assert"
```
Groxx, line 25: same, just fmt

Shaddoll (approved):

internal/error.go
```diff
@@ -419,3 +465,56 @@ func (b ErrorDetailsValues) Get(valuePtr ...interface{}) error {
...
+		decisionText = util.DecisionToString(decision)
+	}
+	return &NonDeterministicError{
+		Reason: reason,
+
+		WorkflowType: info.WorkflowType.Name,
+		WorkflowID:   info.WorkflowExecution.ID,
+		RunID:        info.WorkflowExecution.RunID,
+		DomainName:   info.Domain,
+
```
agautam478, line 491: Should we add the `TaskList:(Name:canary-task-queue)` object too since it was a part of the original error string?

Groxx: yea, seems reasonable

davidporter-id-au (approved):

PR #1271, opened 2023-09-12
Lower "Failed to send heartbeat" log level

jakobht:
We now log when the automatic heartbeating fails introduced in #1263

**What changed?**
We introduced this log to help stakeholders debug, however now stakeholders are being flooded with this log, so lowering the log level.

**Why?**
Lowered to avoid stakeholders who cancel workflows being flooded with error logs

**How did you test it?**

**Potential risks**
No risk


jakobht:
Changed so we now check if the error is a canceledError, in that case we do not log, otherwise we keep the error log

dkrotx (approved):

PR #1272, opened 2023-09-19
Notes for adding history size, and when we can change it safely for replay

Groxx:
Unfortunately complicated :|

I think this is all both necessary and sufficient for safety though.
The client-library-version metadata is NOT necessary, as any change that would
read this and change behavior needs GetVersion to be safe, but we should still
consider adding it ASAP for troubleshooting purposes.  It'd be useful to be
able to tell if version N created an event -> if it should have data X or Y.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -367,7 +367,24 @@ OrderEvents:
 	eh.loadedEvents = eh.loadedEvents[eh.currentIndex:]
 	eh.currentIndex = 0
 
+	// calculate estimated size of all events, for workflow info
+	historySize := estimatedSize(nextEvents) + estimatedSize(markers)
+
 	return nextEvents, markers, nil
+}
+
+func estimatedSize(events []s.HistoryEvent) int {
+	sum := 0
+	for _, e := range events {
+		switch *e.EventType {
+		case s.EventTypeWorkflowExecutionStarted:
+			sum += len(e.WorkflowExecutionStartedEventAttributes.Input)
+		// ...
+		case s.EventTypeWorkflowExecutionSignaled:
+			sum += len(e.WorkflowExecutionSignaledEventAttributes.Input)
+		}
+	}
+	return sum
 }
 
 func isPreloadMarkerEvent(event *s.HistoryEvent) bool {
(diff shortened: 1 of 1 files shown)


internal/internal_task_handlers.go
```diff
@@ -891,6 +910,21 @@ ProcessEvents:
...
+				// because it may depend on the value
+				w.workflowInfo.HistorySize += historySize
+				/*
+					TODO: can't expose this unless the history is currently executing on a "new library version" that is history-size aware,
+						else you can get inconsistent replay behavior when it suddenly jumps from 0 (nonexistent data in the past) to N (calculated now during replay).
+						so this needs a new piece of metadata in decision task complete (possibly activity would be useful too for troubleshooting?) that reports the
+						library "feature version" for backwards compat reasons like this.
+
+					alternatively: could just record a new marker to say "upgraded to version N", but that means a new marker in every workflow.  seems like wasteful storage?
+				*/
```
Groxx, line 925: correction: no, not necessary.
any code that's added that changes behavior based on this new data needs to be `GetVersion`ed anyway, so that should take care of determinism based on this info existing vs not existing.

internal/internal_task_handlers.go
```diff
@@ -367,9 +367,26 @@ OrderEvents:
...
+		switch *e.EventType {
+		case s.EventTypeWorkflowExecutionStarted:
+			sum += len(e.WorkflowExecutionStartedEventAttributes.Input)
+		// ...
+		case s.EventTypeWorkflowExecutionSignaled:
+			sum += len(e.WorkflowExecutionSignaledEventAttributes.Input)
+		}
+	}
+	return sum
+}
```
Groxx, line 388: basic thought here: if we add fields in the future (to either client or server) we could change the calculation if we counted _the whole event_, so don't do that.  count only "big" user-controlled fields, and assume that's close enough (maybe with an offset to get a more accurate estimate).

so stuff like:
- input
- headers
- reset points (maybe) (because they can be large)
- search attributes
- etc

but not e.g. workflow ID fields or anything that's "small" or could be migrated to a different field / different lengths in the future.

server-side the calculation for size limits is simply the whole list of events serialized as thrift.  the server can change this arbitrarily and reject new events without causing determinism failures, but we can't do that safely here because adding any RPC field could change the calculation.

---

alternatively, we could calculate _the size we originally told the client_ and store that in the events, which would allow us to return arbitrary data and claim it's a different size.  that would get us more accuracy + strong stability at the cost of a small bit more storage.

I'm not sure that's worth it though, and it still can't be _truly_ precise because it has to be backwards compatible at all times, and making it intelligent about client versions / changing calculations / etc may be too complex anyway.  Just counting the "big" things seems fair enough.

PR #1273, opened 2023-09-20
Refactor failed heartbeats log

jakobht:
**What changed?**
Refactored the type check to use modern Go 

**What changed?**
In modern go typecasting errors is sometimes not correct as they might be wrapped, the errors.As method is aware of this and checks all wrapped errors 

**How did you test it?**
Tested locally

**Potential risks**
Worst case we log too much/too little

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1497,7 +1497,7 @@ func (i *cadenceInvoker) heartbeatAndScheduleNextRun(details []byte) error {
 			// Log the error outside the lock.
 			// If the error is a canceled error do not log, as this is expected.
 			var canceledErr *CanceledError
-			if !errors.As(err, &canceledErr) {
+			if canceledErr != nil && !errors.As(err, &canceledErr) {
 				i.logger.Error("Failed to send heartbeat", zap.Error(err), zap.String(tagWorkflowType, i.workflowType), zap.String(tagActivityType, i.activityType))
 			}
 		}()
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -1495,11 +1495,10 @@ func (i *cadenceInvoker) heartbeatAndScheduleNextRun(details []byte) error {
 			i.Unlock()
 
 			// Log the error outside the lock.
-			if err != nil {
-				// If the error is a canceled error do not log, as this is expected.
-				if _, ok := err.(*CanceledError); !ok {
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -1495,11 +1495,10 @@ func (i *cadenceInvoker) heartbeatAndScheduleNextRun(details []byte) error {
...
 
 			// Log the error outside the lock.
-			if err != nil {
-				// If the error is a canceled error do not log, as this is expected.
-				if _, ok := err.(*CanceledError); !ok {
-					i.logger.Error("Failed to send heartbeat", zap.Error(err), zap.String(tagWorkflowType, i.workflowType), zap.String(tagActivityType, i.activityType))
-				}
+			// If the error is a canceled error do not log, as this is expected.
+			var canceledErr *CanceledError
+			if errors.As(err, &canceledErr) {
```
3vilhamster, line 1500: You need to reverse the check.

jakobht: Yep, thanks!

3vilhamster (approved):

internal/internal_task_handlers.go
```diff
@@ -1495,11 +1495,10 @@ func (i *cadenceInvoker) heartbeatAndScheduleNextRun(details []byte) error {
...
 
 			// Log the error outside the lock.
-			if err != nil {
-				// If the error is a canceled error do not log, as this is expected.
-				if _, ok := err.(*CanceledError); !ok {
-					i.logger.Error("Failed to send heartbeat", zap.Error(err), zap.String(tagWorkflowType, i.workflowType), zap.String(tagActivityType, i.activityType))
-				}
+			// If the error is a canceled error do not log, as this is expected.
+			var canceledErr *CanceledError
+			if canceledErr != nil && !errors.As(err, &canceledErr) {
```
3vilhamster, line 1500: canceledErr is always nil and err is checked for nil inside errors.As.

3vilhamster (changes requested):
Please revert recent change :)

internal/internal_task_handlers.go
3vilhamster: also, maybe put this into a method and add a test so you are 100% sure about the behaviour. 
Something like 
i.processHearbeatError(err)
and then add unit test with zaptest.Logger.

3vilhamster (approved):
LGTM

PR #1275, opened 2023-10-08
Create a new poller thread pool usage metric

timl3136:
Create a new poller thread pool usage metric for the amount of buffer left in the poller channel. If there are non-remaining, then polling will halt.

**What changed?**
Added a new metric regarding the usage and remaining capacity of the poller channel

**Why?**
We are aiming to understand current poller utilizations and improve the overall utilization as a result. This metric allows us to determine whether there exists a need to tune the amount of buffer of the channel for high traffic customers.

**How did you test it?**
Tested on development environment

**Potential risks**
No risks considering that is only a new metric

### DIFF
--- a/internal/common/metrics/constants.go
+++ b/internal/common/metrics/constants.go
@@ -114,4 +114,6 @@ const (
 	MemoryUsedHeap  = CadenceMetricsPrefix + "memory-used-heap"
 	MemoryUsedStack = CadenceMetricsPrefix + "memory-used-stack"
 	NumGoRoutines   = CadenceMetricsPrefix + "num-go-routines"
+
+	PollerRemainingBuffer = CadenceMetricsPrefix + "poller-remaining-buffer"
 )
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -149,6 +149,14 @@ type (
 	polledTask struct {
 		task interface{}
 	}
+
+	// index tells us where is the poll request coming from
+	// if index == -1, means it came from previous empty poll response
+	// if index >= 0, means it came from task dispatcher with that index
+	// if index == -2, means it came from previous completed task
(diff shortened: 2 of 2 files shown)


internal/common/metrics/constants.go
```diff
@@ -114,4 +114,6 @@ const (
 	MemoryUsedHeap  = CadenceMetricsPrefix + "memory-used-heap"
 	MemoryUsedStack = CadenceMetricsPrefix + "memory-used-stack"
 	NumGoRoutines   = CadenceMetricsPrefix + "num-go-routines"
+
+	PollerRemainingBuffer = CadenceMetricsPrefix + "poller-remaining-buffer"
```
shijiesheng, line 118: metrics name is confusing. The terminology we use is "max concurrent task" in worker options. How about having two metrics:
1. concurrent-task-quota
2. concurrent-task-running (or concurrent-task-running)

internal/internal_worker_base.go
```diff
@@ -252,6 +252,7 @@ func (bw *baseWorker) runPoller() {
 		case <-bw.shutdownCh:
 			return
 		case <-bw.pollerRequestCh:
+			bw.metricsScope.Gauge(metrics.PollerRemainingBuffer).Update(float64(len(bw.pollerRequestCh)))
```
shijiesheng, line 255: as mentioned above we can emit two gauge.
```suggestion
			bw.metricsScope.Gauge(metrics.ConcurrentTaskQuota).Update(float64(cap(bw.pollerRequestCh)))
                         bw.metricsScope.Gauge(metrics.ConcurrentTaskRunning).Update(float64(cap(bw.pollerRequestCh) - len(bw.pollerRequestCh)))
```

shijiesheng (approved):

PR #1277, opened 2023-10-10
Tracking cache invalidation for different types of workflow depends on their runtime length

timl3136:
**What changed?**
Added a new metrics with tag of workflow runtime type based on their timeout value.

**Why?**
Currently the sticky cache invalidation metrics reports on only workflow level and it is difficult to identify whether that came from a short lived workflow or not. Sticky cache invalidation is important to track for the shortlived wf since high invalidation can slow the processing and increase network load significantly.

**How did you test it?**
Tested both locally and on development environment

**Potential risks**
This is only new metrics added.


internal/internal_task_handlers.go
```diff
@@ -657,7 +663,10 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 	task *s.PollForDecisionTaskResponse,
 	historyIterator HistoryIterator,
 ) (workflowContext *workflowExecutionContextImpl, err error) {
-	metricsScope := wth.metricsScope.GetTaggedScope(tagWorkflowType, task.WorkflowType.GetName())
+	// We probably need to have the tasklist name in tags as well,
```
timl3136, line 666: Or not, combining with the implementation of scalable tasklist, it might result in 10x increase in the cardinality for that metric.

internal/internal_logging_tags.go
```diff
@@ -21,25 +21,26 @@
...
+	tagSideEffectID          = "SideEffectID"
+	tagChildWorkflowID       = "ChildWorkflowID"
+	tagLocalActivityType     = "LocalActivityType"
+	tagLocalActivityID       = "LocalActivityID"
+	tagQueryType             = "QueryType"
+	tagVisibilityQuery       = "VisibilityQuery"
+	tagPanicError            = "PanicError"
+	tagPanicStack            = "PanicStack"
+	causeTag                 = "pollerrorcause"
+	tagworkflowruntimelength = "workflowruntimelength"
```
taylanisikdemir, line 45: all the tags above have PascalCase naming convention except `pollererrorcause`. better to keep these consistent. at least for the new one you are introducing

taylanisikdemir (approved):

internal/internal_logging_tags.go
taylanisikdemir: Changing an existing tag might cause problems in metric dashboards that refer to it. Might worth double checking dependencies before improving the name to fit the convention

PR #1279, opened 2023-10-19
Improve retrieval of binaryChecksum

taylanisikdemir:
**What changed?**
`binaryChecksum` gets initialized somewhere early on in worker's lifecycle and then read many times. It was protected by a `sync.Mutex` for both reads and writes. Typically `sync.Once` is used for such scenarios but there is a `SetBinaryChecksum()` function exposed so it wouldn't work here. 
Changing the implementation to avoid mutex locks for reads once `binaryChecksum` is initialized/set.

**Why?**
This change might not have visible performance improvements but still better to use atomic instead of a mutex each time `binaryChecksum` is read.

**How did you test it?**
`make test`

**Potential risks**
N/A

### DIFF
--- a/internal/internal_worker.go
+++ b/internal/internal_worker.go
@@ -36,6 +36,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.uber.org/cadence/internal/common/isolationgroup"
@@ -792,7 +793,7 @@ func (aw *aggregatedWorker) RegisterActivityWithOptions(a interface{}, options R
 }
 
 func (aw *aggregatedWorker) Start() error {
-	if err := initBinaryChecksum(); err != nil {
+	if _, err := initBinaryChecksum(); err != nil {
 		return fmt.Errorf("failed to get executable checksum: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/internal_worker.go
```diff
@@ -875,7 +875,7 @@ func (aw *aggregatedWorker) Start() error {
 }
 
 var binaryChecksum string
-var binaryChecksumLock sync.Mutex
+var binaryChecksumLock sync.RWMutex
```
Groxx, line 878: this does seem better, but eh... if we're going to rewrite this, let's just use a `sync.Once`.  the semantics are much more accurate and it's probably faster.

taylanisikdemir: There's `SetBinaryChecksum()` function so can't use `sync.Once` as I mentioned in PR description.

Groxx: ah, it is fully mutable isn't it.  I was probably thinking of the sticky cache size.

atomics are probably fine then tbh, doesn't seem like anything concurrently starting-and-updating-checksum can expect any sort of ordering between calls.

but yea, rwmutex is an improvement either way, I can 👍 the PR for that.

internal/internal_worker.go
```diff
@@ -923,14 +918,16 @@ func initBinaryChecksumLocked() error {
...
-		}
+	if len(val) != 0 { // already initialized.
+		return val
+	}
+
+	if err := initBinaryChecksum(); err != nil {
+		panic(err)
 	}
 
 	return binaryChecksum
```
Groxx, line 933: this is no longer protected, and is a racy read.
a simple/safe fix may be to just return the calculated value from init.

or switch to atomics, and don't care about races between concurrent inits - any value is equally valid (and they are likely identical).

taylanisikdemir: Good catch. Changing type to `binaryChecksum atomic.Value` and getting rid of mutexes for simplicity. Given the usage pattern it seems fine to init multiple times if it happens to be called concurrently during startup

internal/internal_worker.go
```diff
@@ -874,66 +875,51 @@ func (aw *aggregatedWorker) Start() error {
...
 }
 
-// callers MUST hold binaryChecksumLock before calling
-func initBinaryChecksumLocked() error {
-	if len(binaryChecksum) > 0 {
-		return nil
-	}
-
+func initBinaryChecksum() (string, error) {
 	exec, err := os.Executable()
```
Groxx, line 886: still needs this check - sequential startup has no need to re-compute the same thing repeatedly, and it'll also clobber whatever your current manually-set checksum might be.

internal/internal_worker.go
```diff
@@ -874,66 +875,51 @@ func (aw *aggregatedWorker) Start() error {
...
 	f, err := os.Open(exec)
 	if err != nil {
-		return err
+		return "", err
 	}
 	defer func() {
 		_ = f.Close() // error is unimportant as it is read-only
 	}()
 
 	h := md5.New()
```
Groxx, line 899: hmmm.  tbh I forgot that this was the default behavior, since internally we report git-SHAs from an environment variable rather than checksumming the binary...

this might make an rwmutex better :|  checksumming big binaries can take time.  though at least it's only at startup.

taylanisikdemir: makes sense. let's lock around this part to avoid reading/hashing binary content multiple times

Groxx (approved):
Looks good - thanks, and sorry about all the mind-changing!  I shoulda read init more closely earlier.

PR #1281, opened 2023-10-25
WIP: Address false posivite non-determinism errors during replay

taylanisikdemir:
**What changed?**
TODO

**Why?**

**How did you test it?**

**Potential risks**


internal/internal_task_handlers.go
```diff
@@ -830,6 +852,19 @@ process_Workflow_Loop:
 	return response, err
 }
 
+// ProcessWorkflowTask processes the given workflow which includes
+// - fetching, reordering and replaying historical decision events. (Decision events in this context is an umbrella term for workflow relevant events)
+// - state machine is incrementally built with every decision.
+// - state machine makes sure that when a workflow restarts for some reason same activities (or timers etc.) are not called again and previous result state is loaded into memory
+//
+// Note about Replay tests mode:
```
taylanisikdemir, line 860: @Groxx / @agautam478 This PR is not fully ready to be reviewed but requesting early feedback from you two. 
I feel like the false positives are mostly handled with this fix and `make test` is passing. Not sure how much I should trust on the test coverage yet :) We can also have a quick call tomorrow to discuss if that works.

agautam478: the tests are not reliable indicators of the changes especially in Replay tests. Yes, let's get on a call.

internal/internal_task_handlers.go
```diff
@@ -260,6 +260,28 @@ func isDecisionEvent(eventType s.EventType) bool {
 	}
 }
 
+// isDecisionEventForReplay is different from isDecisionEvent because during replays
+// we want to intentionally ignore workflow complete/fail/cancel/continueasnew events so that
+// decision tree replays matches with the workflow processing respond tasks
+func isDecisionEventForReplay(eventType s.EventType) bool {
```
agautam478, line 266: I has tried something similar a while ago (a bit crude in comparison) but then dropped it because it was messing up the normal task processing and history replay. Are the integration tests passing for you?

internal/internal_task_handlers.go
taylanisikdemir: scheduled for tomorrow

internal/internal_task_handlers.go
```diff
@@ -904,7 +946,11 @@ ProcessEvents:
...
 				return nil, err
 			}
-			if w.isWorkflowCompleted {
+
+			// Break the event processing loop if the workflow is completed except in replay mode.
+			// In replay mode we check for nondeterminism cases and
+			// breaking the loop causes missing events in respondEvents which then causes false positives or false negatives.
+			if w.isWorkflowCompleted && !isInReplay {
 				break ProcessEvents
 			}
```
Groxx, line 955: to stick my initial feelings in here, which **have not been confirmed** and could be totally off:
I _suspect_ this is what's actually passing the tests, not the other changes.  This will mean that after the workflow-func returns, the loop continues _and gets more tasks_ and that does... something.  But it seems fairly likely that it'll lead to a different set of events -> successful non-determinism check.

Which can pretty clearly catch some things, but I suspect it's not actually _correct_ and it'll misbehave in other cases that we may not have tests for.

The "do not compare end requests" change is particularly suspect here IMO, as it feels backwards: we _very much_ care if the end event sequences differ, but this change is filtering them out so we can't tell.

So: interesting!  I'm genuinely surprised this was so simple and passed all tests.  That might imply it's more correct than I think it is... but I have moderately strong doubts at the moment.

taylanisikdemir: I added comments in various places to explain what is happening because this essentially has dependency on everything happening underneath in client code. Hope the comments help a little bit to clarify things. 

In summary, this change was needed because `w.isWorkflowCompleted` becomes true while there's still stuff to do in the history such as other activities to schedule/complete. So in replay mode we shouldn't break the loop early. Without looking at remaining events it's impossible to know if there's non-determinism issues. 
On top of this change I also had to change the `isDecisionEvent` for replays because the `WorkflowExecutionCompleted` (and other workflow end state events) should be ignored. Otherwise you end up with an extra event inside `respondEvents`. Previously this wasn't an issue because loop terminated before it ever got to the final event in the history :)

internal/internal_task_handlers.go
taylanisikdemir: Tests are passing now.

test/replaytests/replay_test.go
```diff
@@ -73,6 +73,7 @@ func TestGreetingsWorkflow(t *testing.T) {
 }
 
 // Should have failed but passed. Maybe, because the result recorded in history still matches the return type of the workflow.
+// TODO(remove comment): Debug why is this still missed
```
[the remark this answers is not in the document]
taylanisikdemir: I debugged all the test cases and added explanations to what is happening. Also validating what type of nondeterminism error it is

internal/internal_task_handlers.go
Groxx: > There's never a corresponding replay decision for `WorkflowExecutionCompleted`

Ah, yeah, good point - that does make the filtering-out make a bit more sense.

the linked [ref](https://github.com/uber-go/cadence-client/blob/1fed70036ff0018915a4d35fba9104c5752d6c7d/internal/internal_event_handlers.go#L811) there though is something else entirely.  that means that when we get a pre-recorded "complete" event, it doesn't trigger any progress in the workflow func... because it shouldn't, it should have already returned when handling the same decision task.

taylanisikdemir: The linked ref is relevant in the replay scenario. It points that event handler doesn't do anything for `WorkflowExecutionCompleted` and the loop is not able to grab something [here](https://github.com/uber-go/cadence-client/blob/25c0a0996205e4b35b3fa763ea29a44b4365f3d8/internal/internal_task_handlers.go#L976) to append to `replayDecisions`. Therefore [this comparison](https://github.com/uber-go/cadence-client/blob/25c0a0996205e4b35b3fa763ea29a44b4365f3d8/internal/internal_task_handlers.go#L993) fails unless I use the new `isDecisionTaskForReplay`

Groxx: gonna have to dig into this a bit more closely 🤔 
there's a lot of blending of "history has done X" and "replay has done Y" in that, and I'm not sure if this would be a reasonable addition or not.  or if it would be more accurately modeled by e.g. having a state machine for the workflow as a whole (probably just completed -> sent)...

test/replaytests/replay_test.go
```diff
@@ -81,92 +84,110 @@ func TestGreetingsWorkflow3(t *testing.T) {
...
-
 	err := replayer.ReplayWorkflowHistoryFromJSONFile(zaptest.NewLogger(t), "branch2.json")
-	require.NoError(t, err)
+	assert.ErrorContains(t, err, "nondeterministic workflow: missing replay decision")
+}
+
+// Runs a history which ends with WorkflowExecutionContinuedAsNew. Replay fails because of the additional checks done
+// for continue as new case by replayWorkflowHistory().
+// This should not have any error because it's a valid continue as new case.
+func TestContinueAsNew(t *testing.T) {
```
taylanisikdemir, line 188: this case will be handled in a follow up PR

internal/internal_activity.go
```diff
@@ -432,6 +432,10 @@ func deSerializeFunctionResult(f interface{}, result []byte, to interface{}, dat
 	}
 
 	// For everything we return result.
+	// Code reaches here for 2 cases:
+	// 	1. activity is executed by name (not the func pointer) and it wasn't registered
+	// 	2. activity is executed by func pointer and the signature indicates it doesn't/can't return data.
+	//		for example it only has one return parameter (which can only be be error).
 	return decodeArg(dataConverter, result, to)
 }
```
Groxx, line 440: Huh.  On closer look, this whole chain of code is pretty strange:
- `deSerializeFunctionResult` is only called if `result` is non-nil
- `deSerializeFnResultFromFnType` is only called by this func, and it does nothing if `result` is nil
- none of this needs the _registered func type_ at all:
  - if everything agrees, the type does nothing to contribute to the behavior
  - if there if there is data + an out param + the registered activity _does not_ return data, the earlier code silently does nothing (the data exists but is inaccessible)
  - if it's _not_ registered, it falls back to just decoding directly, which is what `deSerializeFnResultFromFnType` does anyway

... I think `deSerializeFunctionResult` might be more-correctly replaced with just a call to `decodeArg` instead.

test/replaytests/replay_test.go
Groxx: yea, replayer shouldn't care.  _shadower_ might, but that's completely different.

separately, now that I look at this... it feels strange that it doesn't return `future, err`.  replaying to get the result (from code) seems entirely reasonable, no need to only return error...

test/replaytests/choice.json
```diff
@@ -176,7 +176,34 @@
...
+      "identity": "82203@agautam-NV709R969P@choiceGroup@41d230ae-253a-4d01-9079-322ef05c09fb",
+      "requestId": "ae2aad96-6588-4359-807b-a39a16f0896a",
+      "attempt": 0,
+      "lastFailureReason": ""
+    }
+  },
+  {
+    "eventId": 13,
+    "timestamp": 1679427717321780255,
+    "eventType": "ActivityTaskCompleted",
```
Groxx, line 194: why add the completed event?

though reading more of this file: I don't think this is a real history... "completed" implies "ended normally, i.e. by returning" but we don't create activity-task-scheduled events for activities executed during the final decision (since their result cannot be observed)...

test/replaytests/continue_as_new_wf.go
```diff
@@ -0,0 +1,26 @@
...
+			workflow.GetLogger(ctx).Info("Received age signalResult from signal!", zap.String("signal", signalName), zap.String("value", signalResult))
+		})
+		workflow.GetLogger(ctx).Info("Waiting for signal on channel.. " + signalName)
+		// Wait for signal
+		selector.Select(ctx)
+		if signalResult == "kill" {
+			return nil
+		}
+	}
+}
```
Groxx, line 26: since the only history this is running on is the `continue_as_new.json`, this could be just
```go
func ContinueAsNewWorkflow(ctx workflow.Context) error {
	return workflow.GetSignalChannel(ctx, "unused").Receive(ctx, nil)
}
```
and it would be exactly equivalent.

agautam478: afaik, continue ask new cases fail for only those scenarios where a workflow was marked continue as new and then the history has nothing after continue as new decision. 
This looks like an attempt to cover the edge case where sometimes the workflow continues as new after receiving a signal. the continue as new case has been pretty inconsistent to begin with.

I would say let's keep the original behavior here to preserve the original scenario where we first observed this bug.

Groxx: there's no signal in the history though, so it can't be a "received one, did not receive a second" kind of test.  anything that waits forever without recording an event should be equivalent (signal chan, regular chan, waitgroup, etc all would work)

internal/internal_task_handlers.go
```diff
@@ -262,6 +262,28 @@ func isDecisionEvent(eventType s.EventType) bool {
...
+		s.EventTypeMarkerRecorded,
+		s.EventTypeStartChildWorkflowExecutionInitiated,
+		s.EventTypeRequestCancelExternalWorkflowExecutionInitiated,
+		s.EventTypeSignalExternalWorkflowExecutionInitiated,
+		s.EventTypeUpsertWorkflowSearchAttributes:
+		return true
+	default:
+		return false
+	}
+}
```
Groxx, line 285: To stick a comment in here from chats a while back, for visibility:
I _generally like_ this approach and I think it stands a pretty good chance of being an improvement...

... but I'm not (yet) confident that it's _correct_, nor that it doesn't worsen other scenarios.  It's a pretty substantial change despite how small it is.  From vague memory of my last attempt to verify, I ended up with more concerns than I started with, but no real evidence in either direction :\

It may be fine even if it does introduce new flaws, it's definitely not perfect now, but we need to have an informed decision.  And we now have a LOT more internal replay-shadow-tests (compared to when this was first made) that we can run to discover what has changed - if it finds nothing but improvements against our fairly wide range of behavior in those tests, that will probably be good enough.
I just _definitely_ do not trust our test suite here to _prove correctness_, they don't even use like half of our features nor do they use them in complicated ways.  Issues they find are plenty useful, but they're very strange in a lot of ways and that makes them very very far from evidence of correctness.

---

So tl;dr: it just has to be sufficiently-validated or fully understood/described.
We've done some validation (thanks @agautam478 !) but we need to do a bit more, and none of the tests here seem like particularly strong evidence for _what changed_ so I need to dig back in and try to reason about it more carefully.

test/replaytests/choice.json
taylanisikdemir: it wasn't a realistic history because it's missing the final decision task scheduled->started->completed events. I'm not quite getting why adding the activity completion looks invalid if that's what you meant

internal/internal_task_handlers.go
taylanisikdemir: All sounds good to me. At the time I looked into these false positives I focused on a subset of the potential scenarios. I'm not super confident this is completely safe change even though it looks like it is. Finding more scenarios and adding them as test cases should be next step as discussed offline. Due to other priorities I'm handing this over to you and @agautam478.

internal/internal_activity.go
taylanisikdemir: TBH I need to re-debug this to remember how I ended up finding those 2 cases. Maybe I meant to add this somewhere else. From the code in this function it seems 2 is not possible.

test/replaytests/choice.json
Groxx: well, there are only 2 ways workflows can end:
- return normally from the code (success or error)
- cut off externally due to timeout/termination/etc

and everything that affects ^ this occurs in transactions (no interleaving between these):
- a decision and all its operations go in as one chunk
- an external event *that the workflow can observe* goes in with the event and a decision-task-schedule to process it
- an external event that *cannot* be observed just gets appended
- cutoffs write one event and then end

---

in this history we originally had:
- decision task schedule->start->complete
- activity scheduled
- workflow complete
  - a normal one, which means it returned, which means this was the result of the most-recent decision task.

^ this is possible because, unlike child workflows, we don't "prune" activity-scheduled-s that are created in the final decision task.  (we should consider doing that tbh, or keeping both for easier troubleshooting)

So in this final decision task, we executed one activity, did not wait on it, and returned immediately.  Code like this would result in that:
```go
func work(ctx workflow.Context) error {
  // ...
  workflow.ExecuteActivity(ctx, "ignored") // non-blocking
  return nil
}
```

---

As of the current PR, we have this sequence:
- decision task schedule->start->complete
- activity scheduled
- activity started
- activity complete
- workflow complete (still normal)

which is not possible.

it would mean we blocked after scheduling (this is necessary to submit it to the server, which is a necessary prerequisite to the activity starting)... which is fine because `workflow.ExecuteActivity(...).Get(..)` would end the batch at "scheduled".

_however_.

after that, asynchronously, the activity was started (this can be an isolated event because it's not observable, that's fine), and then completed (this requires both complete and decision-task-scheduled because it _is_ observable, so already it's wrong), and then _also_ the workflow function returned on its own _without a decision task_ to learn about that result.

The only possible outcomes that include an activity that was started are things like either:
- activity started -> activity completed -> decision scheduled -> started -> completed -> workflow completed
- activity started -> workflow cut off externally (timeout, terminate) without a decision task

because "activity completed -> workflow completed" would be splitting the transaction that "activity completed" is involved in.

Groxx: tbh I'm kinda surprised it didn't panic with an illegal transition. But it would be an easy check to miss in the state machine too (particularly because it's more "a collection of many sub-state-machines" rather than enforcing rules in the spaces between them)

PR #1284, opened 2023-10-27
Fixed error log

jakobht:
**What changed?**
We currently log the wrong error which is always nil, so fixed to log the correct error

**Why?**
Because we log the wrong error

**How did you test it?**
N/A

**Potential risks**
Very low, just changing a log


3vilhamster (approved):
Ok

PR #1286, opened 2023-11-01
add WithTaskListMapper and WithWorkflowTasklistMapper utils for workflow executions

shijiesheng:
**What changed?**

* Added util functions to change tasklists (activity tasklists and workflow tasklists)

**Why?**

This is needed for internal usage to support tenancy separation

**How did you test it?**

unit test

**Potential risks**

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1413,6 +1413,13 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
 	return ctx1
 }
 
+// WithWorkflowTaskListApply returns a copy of context and changes workflow tasklist with apply function.
+func WithWorkflowTaskListApply(ctx Context, apply func(*string) *string) Context {
+	ctx1 := setWorkflowEnvOptionsIfNotExist(ctx)
+	getWorkflowEnvOptions(ctx1).taskListName = apply(getWorkflowEnvOptions(ctx1).taskListName)
+	return ctx1
+}
+
 // WithWorkflowID adds a workflowID to the context.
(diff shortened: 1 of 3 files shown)


internal/workflow.go
```diff
@@ -1413,6 +1413,13 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
 	return ctx1
 }
 
+// WithWorkflowTaskListApply returns a copy of context and changes workflow tasklist with apply function.
```
taylanisikdemir, line 1416: nit: maybe a better name is `WithWorkflowTaskListMapper`

internal/workflow.go
```diff
@@ -1413,6 +1413,13 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
...
+func WithWorkflowTaskListApply(ctx Context, apply func(*string) *string) Context {
```
taylanisikdemir, line 1417: we don't need to pass the pointer to this callback func.
`func(string) string` would be more user friendly

internal/workflow.go
taylanisikdemir: same in other places

taylanisikdemir (approved):

internal/workflow.go
shijiesheng: fixed

internal/workflow.go
shijiesheng: make sense

Groxx:
er. what's the purpose here / how is it intended to be used?

PR #1287, opened 2023-11-03
Honor non-determinism fail workflow policy

taylanisikdemir:
**What changed?**
Users can specify [NonDeterministicWorkflowPolicy](https://github.com/taylanisikdemir/cadence-client/blob/master/internal/worker.go#L201) in worker options. If the FailWorkflow policy is chosen the workflow is expected to terminate as soon as it ends up with a nondeterministic state (e.g. activity order changed). 
However this wasn't honored for a category of nondeterminism cases. This PR addresses it and workflows fail once any nondeterminism scenario is encountered.

There are two categories of nondeterminism cases in terms of how they get detected by client library:
1. Issue bubbles up as illegal state panic to the task handler. Most actual prod cases.
2. Issue is caught when comparing replay decisions with history. Replay test scenarios and a subset of prod cases.

FailWorkflow policy was honored for 2 but not for 1.

**Why?**
To make `NonDeterministicWorkflowPolicy` feature correct/complete.

**How did you test it?**
Added an integration test to simulate this scenario.

**Potential risks**
Users depending on existing buggy behavior can be impacted. This would only happen if and only if all the below holds true
- user set FailWorkflow policy
- user expects to not terminate category 1 (see above) workflows

This is not very realistic expectation because users don't know about these subcategories of nondeterminism detection mechanisms. 
So the risk of this fix should be minimal.


internal/internal_task_handlers.go
```diff
@@ -944,33 +944,23 @@ ProcessEvents:
...
+			zap.String(tagRunID, task.WorkflowExecution.GetRunId()),
+			zap.Error(nonDeterministicErr),
+			zap.NamedError("ReplayError", replayErr),
+		)
 	}
 
 	if nonDeterministicErr != nil {
-
+		// TODO: add a NonDeterminismDetectionType tag based on which branch set the nonDeterministicErr above.
+		// Q: What is our approach for such tag changes of existing metrics as it is not allowed by some timeseries dbs (prometheus)
```
taylanisikdemir, line 963: reminder for myself to remove the comment once discussed with the team

internal/internal_task_handlers.go
```diff
@@ -939,39 +939,31 @@ ProcessEvents:
...
+			zap.String(tagRunID, task.WorkflowExecution.GetRunId()),
+			zap.Error(nonDeterministicErr),
+			zap.NamedError("ReplayError", replayErr),
+		)
 	}
 
 	if nonDeterministicErr != nil {
-
-		w.wth.metricsScope.GetTaggedScope(tagWorkflowType, task.WorkflowType.GetName()).Counter(metrics.NonDeterministicError).Inc(1)
+		scope := w.wth.metricsScope.GetTaggedScope(tagWorkflowType, task.WorkflowType.GetName(), tagNonDeterminismDetectionType, string(nonDeterminismType))
```
Groxx, line 965:
```
string(nonDeterminismType)
```
tbh, the more I write Go, the more I want to get rid of all type casts (outside that type's methods) :|  there are just zero safety checks around them, feels like more than half of the time they just become landmines for later.  (int conversion, non-utf bytes, etc)

I'd probably just leave these as strings unless we have a typed API that accepts only them, because that's capable of preventing misuse.

internal/internal_task_handlers.go
```diff
@@ -1835,7 +1845,11 @@ func traceLog(fn func()) {
 	}
 }
 
-func workflowCategorizedByTimeout(executionTimeout int32) string {
+func workflowCategorizedByTimeout(wfContext *workflowExecutionContextImpl) string {
+	if wfContext == nil || wfContext.workflowInfo == nil {
+		return "unknown"
+	}
```
Groxx, line 1851: nils are not possible here (as it's currently used), we can/should drop the checks.

internal/internal_task_handlers_test.go
```diff
@@ -869,18 +869,11 @@ func (t *TaskHandlersTestSuite) TestWorkflowTask_NondeterministicLogNonexistingI
...
-	// The issue is that we change the error type and that we change the error message, the customers
-	// are checking the error string - we plan to wrap all errors to avoid this issue in client v2
-	t.NoError(err)
-	t.NotNil(response)
+	t.Nil(request)
+	t.ErrorContains(err, "nondeterministic workflow")
 
 	// Check that the error was logged
-	ignoredWorkflowLogs := logs.FilterMessage("Ignored workflow panic error")
+	ignoredWorkflowLogs := logs.FilterMessage("Illegal state caused panic")
```
Groxx, line 876: probably worth changing this var name now too

test/activity_test.go
```diff
@@ -105,9 +105,7 @@ func (a *Activities) invoked() []string {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 	result := make([]string, len(a.invocations))
-	for i := range a.invocations {
-		result[i] = a.invocations[i]
-	}
+	copy(result, a.invocations)
```
Groxx, line 108: 👍 though I do look forward to generics-izing this kind of thing.  seems a shame to leave the "set wrong len / did not check return value" gap exposed.

test/integration_test.go
```diff
@@ -437,7 +443,7 @@ func (ts *IntegrationTestSuite) TestChildWFWithParentClosePolicyAbandon() {
 	ts.NoError(err)
 	resp, err := ts.libClient.DescribeWorkflowExecution(context.Background(), childWorkflowID, "")
 	ts.NoError(err)
-	ts.True(resp.WorkflowExecutionInfo.GetCloseTime() == 0)
+	ts.Truef(resp.WorkflowExecutionInfo.GetCloseTime() == 0, "Expected close time to be zero, got %d. Describe response: %+v", resp.WorkflowExecutionInfo.GetCloseTime(), resp)
```
Groxx, line 446:
```suggestion
	ts.Zerof(resp.WorkflowExecutionInfo.GetCloseTime(), "Child workflow should not be closed, describe response: %+v", resp)
```
possibly %#v too, but if + looks better here then definitely 👍 .  I just find + to be pretty rarely supported, and the %v it degrades to is pretty bad for some structures.

internal/internal_task_handlers.go
```diff
@@ -681,11 +681,11 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 		workflowContext = getWorkflowContext(runID)
 	}
 
+	executionRuntimeType := workflowCategorizedByTimeout(workflowContext)
+	metricsScope = metricsScope.Tagged(map[string]string{tagWorkflowRuntimeLength: executionRuntimeType})
```
Groxx, line 685: this will affect the deferred gauge, among other things - doesn't seem like something we want to change here.

(which looks also true for line 688, so let's change that here too / ASAP)

test/integration_test.go
```diff
@@ -148,18 +148,24 @@ func (ts *IntegrationTestSuite) SetupTest() {
...
+
+func (ts *IntegrationTestSuite) BeforeTest(suiteName, testName string) {
 	options := worker.Options{
 		DisableStickyExecution:            ts.config.IsStickyOff,
 		Logger:                            zaptest.NewLogger(ts.T()),
 		WorkflowInterceptorChainFactories: []interceptors.WorkflowInterceptorFactory{ts.tracer},
 		ContextPropagators:                []workflow.ContextPropagator{NewStringMapPropagator([]string{testContextKey})},
 	}
+
+	if testName == "TestNonDeterministicWorkflowQuery" || testName == "TestNonDeterministicWorkflowFailPolicy" {
```
Groxx, line 162: I completely believe that this is the easy option, but uh.  tbh this kind of thing is why I want to get rid of these `testify.Suite`s entirely.  spooky action at a distance and ad-hoc DSLs are a massive pain much of the time, if we didn't have this it'd have just been an argument on a `func setup(...)` and it would be obvious and controllable.

test/integration_test.go
```diff
@@ -512,6 +518,31 @@ func (ts *IntegrationTestSuite) TestDomainUpdate() {
 	ts.Equal(description, *domain.DomainInfo.Description)
 }
 
+func (ts *IntegrationTestSuite) TestNonDeterministicWorkflowFailPolicy() {
+	err := ts.executeWorkflow("test-nondeterminism-failpolicy", ts.workflows.NonDeterminismSimulatorWorkflow, nil)
+	customErr, ok := err.(*internal.CustomError)
+	ts.Truef(ok, "expected CustomError but got %T", err)
+	ts.Equal("NonDeterministicWorkflowPolicyFailWorkflow", customErr.Reason())
```
Groxx, line 525: `errors.As` and/or there might be an equivalent in the assertions.  these semantics shouldn't change if/when we add wrapping later.

Groxx (commented):
dropping some notes from a first skim.

overall very close I think, I just need another read in an IDE or something to make sure it's good to go since this area in general is a little bit risky (but totally worth it)

internal/internal_task_handlers.go
```diff
@@ -988,7 +980,7 @@ ProcessEvents:
 			// workflow timeout.
 			return nil, nonDeterministicErr
 		default:
-			panic(fmt.Sprintf("unknown mismatched workflow history policy."))
+			panic("unknown mismatched workflow history policy.")
```
Groxx, line 983: we should really just check and fix all of these, bleh.  I thought I did that a while ago, but I might be thinking of some other codebase...

taylanisikdemir: I address things as I see in relevant places. Let's all encourage this and don't worry about having ~20% refactorings in a diff. More than that should probably be its own PR

internal/internal_task_handlers.go
taylanisikdemir: good catch on the defer. Just changed the deferred func to accept tally scope as argument so it will work without this new tag as before

internal/internal_task_handlers.go
taylanisikdemir: nils are possible in this commit due to where this function is called. I checked with @timl3136 about our intent for `workflowruntimelength`. We will only have it for cache hit metric for now. So moved the scope tagging to appropriate if block back.

internal/internal_task_handlers.go
taylanisikdemir: the underlying enum type alias is a string so this is actually safe but I agree on avoiding castings. however for metric tags enums are more clean way to define fixed set of values and requires this casting due to the tally api accepting strings only.

internal/internal_task_handlers_test.go
taylanisikdemir: missed that, thanks

test/activity_test.go
taylanisikdemir: static check was highlighting this so was obvious fix

test/integration_test.go
taylanisikdemir: This workaround demonstrates one of the weaknesses of test suites where your setup and teardown can only diverge based on test name whereas a proper table driven test would have a `worker_options` field to be specified per test case.
If I find some free time I may rewrite this file by ripping out suite and asserts in a follow up PR

test/integration_test.go
taylanisikdemir: makes sense

internal/internal_task_handlers.go
Groxx: yea, I meant "fix them all" is probably a not-this-PR thing.  happy to approve obviously-safe changes like this as they appear 👍

internal/internal_task_handlers.go
Groxx: ooh, I see, because it was moved outside that != nil check.  yea, good point.

is this tag relevant for the cache miss branch further down too?  I'm kinda feeling like "no" but not totally confident.

test/integration_test.go
Groxx: yea, possibly not worth it in this PR.  more of a "let's try hard to keep this to a minimum please" request - short-term it's fine.

internal/internal_task_handlers.go
taylanisikdemir: we don't have the workflow context for misses in this path so not applicable for it.

test/integration_test.go
```diff
@@ -512,6 +518,32 @@ func (ts *IntegrationTestSuite) TestDomainUpdate() {
...
+
+	// query failed workflow should still work
+	value, err := ts.libClient.QueryWorkflow(ctx, "test-nondeterministic-query", run.GetRunID(), "__stack_trace")
+	ts.NoError(err)
+	ts.NotNil(value)
+	var trace string
+	ts.Nil(value.Get(&trace))
```
Groxx, line 544:
```suggestion
	ts.NoError(value.Get(&trace))
```
I assume that's the intent here?  might as well be a bit closer to type-correctness / it's relevant that it's not an error, not just an [anything not nil].

you can also make this not panic in the nil case with a `require.NotNil(s.T(), value)` earlier.  not particularly important since it's a fairly small error and the stack will point out the issue, just as an FYI.

test/workflow_test.go
```diff
@@ -573,6 +575,48 @@ func (w *Workflows) WorkflowWithLocalActivityCtxPropagation(ctx workflow.Context
...
 
+func (w *Workflows) NonDeterminismSimulatorWorkflow(ctx workflow.Context) ([]string, error) {
+	logger := workflow.GetLogger(ctx).Sugar()
+	ctx = workflow.WithActivityOptions(ctx, w.defaultActivityOptions())
+
+	workflow.SetQueryHandler(ctx, "custom_query", func() (int, error) {
+		return w.nonDeterminismSimulatorWorkflowCallCount, nil
+	})
+
+	res := []string{}
```
Groxx, line 586:
```suggestion
	var res []string
```
for style points / zero allocation if unused / healthier habits.

test/workflow_test.go
```diff
@@ -39,7 +39,9 @@ const (
 	consistentQuerySignalCh = "consistent-query-signal-chan"
 )
 
-type Workflows struct{}
+type Workflows struct {
+	nonDeterminismSimulatorWorkflowCallCount int
```
Groxx, line 43: since this struct is kinda intended for multi-test use... I wonder if this'll come back to bite us some day.  as compared to making a different struct each time to ensure they're isolated.

but eh.  the toggling is a pretty simple way to achieve this even in that case, seems probably fine.

test/integration_test.go
```diff
@@ -148,18 +149,24 @@ func (ts *IntegrationTestSuite) SetupTest() {
 	ts.seq++
 	ts.activities.clearInvoked()
 	ts.taskListName = fmt.Sprintf("tl-%v", ts.seq)
-	ts.worker = worker.New(ts.rpcClient.Interface, domainName, ts.taskListName, worker.Options{
-		DisableStickyExecution: ts.config.IsStickyOff,
```
Groxx, line 152: any idea when this would be false?  if ever?  it being not-hard-coded sorta implies it's controllable somehow.

Groxx: bleh, nevermind.  this is maintained in the new code, I just derped

internal/internal_task_handlers.go
Groxx: I suppose this is actually in use due to the variable 🤔 
I do kinda think we'll want a more strongly typed API for metrics eventually (in particular: to break the dependency on a specific version of Tally), it'll be more useful then too.

internal/internal_task_handlers.go
```diff
@@ -664,12 +665,12 @@ func (wth *workflowTaskHandlerImpl) getOrCreateWorkflowContext(
 	historyIterator HistoryIterator,
 ) (workflowContext *workflowExecutionContextImpl, err error) {
 	metricsScope := wth.metricsScope.GetTaggedScope(tagWorkflowType, task.WorkflowType.GetName())
-	defer func() {
+	defer func(scope tally.Scope) {
```
Groxx, line 668: probably fine but: since we _intentionally_ do not want to use the (possibly mutated) outer-closure's `metricsScope`, personally I'd vote to shadow the name here to make it totally inaccessible.

internal/internal_task_handlers.go
```diff
@@ -1195,6 +1187,24 @@ func (w *workflowExecutionContextImpl) GetDecisionTimeout() time.Duration {
 	return time.Second * time.Duration(w.workflowInfo.TaskStartToCloseTimeoutSeconds)
 }
 
+func (w *workflowExecutionContextImpl) getWorkflowPanicIfIllegaleStatePanic() (*workflowPanicError, bool) {
+	if !w.isWorkflowCompleted || w.err == nil {
```
Groxx, line 1191: `if !w.isWorkflowCompleted`
^ since this is a semantic change: are state-machine-panicked workflows _definitely_ completed?

internal/internal_task_handlers.go
```diff
@@ -939,39 +939,31 @@ ProcessEvents:
...
-						zap.String(tagWorkflowType, task.WorkflowType.GetName()),
-						zap.String(tagWorkflowID, task.WorkflowExecution.GetWorkflowId()),
-						zap.String(tagRunID, task.WorkflowExecution.GetRunId()),
-						zap.Error(nonDeterministicErr),
-						zap.NamedError("ReplayError", replayErr),
-					)
-				}
-			}
-		}
+	} else if panicErr, ok := w.getWorkflowPanicIfIllegaleStatePanic(); ok {
```
Groxx, line 949: this is now preferring a state-machine panic over history-compared-non-determinism which... I don't _think_ is what we want.

iirc: non-determinism triggered earlier could lead to a panic after several steps, and that earlier-non-determinism info is likely more useful as it can tell us "ran activity X instead of Y", which the panic won't know.

Groxx: ah, no, I think that's wrong.  previously it had the possibility of clobbering it, but now it's less likely because it's in an `else` instead of a separate `if`.

rereading more carefully!

internal/internal_task_handlers.go
```diff
@@ -939,39 +939,31 @@ ProcessEvents:
...
+		// This is a nondeterministic execution which ended up panicing
```
Groxx, line 950:
```suggestion
		// This is a nondeterministic execution which ended up panicking
```

test/integration_test.go
```diff
@@ -148,18 +149,24 @@ func (ts *IntegrationTestSuite) SetupTest() {
 	ts.seq++
 	ts.activities.clearInvoked()
 	ts.taskListName = fmt.Sprintf("tl-%v", ts.seq)
-	ts.worker = worker.New(ts.rpcClient.Interface, domainName, ts.taskListName, worker.Options{
```
Groxx, line 151: huh.  so we were previously creating a worker and throwing it away?

thanks for the cleanup :+1:, surprising that got missed when it was introduced.

test/workflow_test.go
```diff
@@ -573,6 +575,48 @@ func (w *Workflows) WorkflowWithLocalActivityCtxPropagation(ctx workflow.Context
 	return result, nil
 }
 
+func (w *Workflows) NonDeterminismSimulatorWorkflow(ctx workflow.Context) ([]string, error) {
+	logger := workflow.GetLogger(ctx).Sugar()
```
Groxx, line 579:
```suggestion
	logger := workflow.GetLogger(ctx)
```
Sugaring is unused, and a better habit to not use / not use unnecessarily since it's not free.

test/workflow_test.go
```diff
@@ -573,6 +575,48 @@ func (w *Workflows) WorkflowWithLocalActivityCtxPropagation(ctx workflow.Context
...
+		}
+		res = append(res, ans)
+	}
+
+	w.nonDeterminismSimulatorWorkflowCallCount++
+
+	selector := workflow.NewSelector(ctx)
+	timer := workflow.NewTimer(ctx, 5*time.Second)
+	selector.AddFuture(timer, func(workflow.Future) {
+		logger.Infof("Timer future is called")
```
Groxx, line 604:
```suggestion
		logger.Info("Timer future is called")
```
but obviously harmless

test/integration_test.go
taylanisikdemir: test code changes usually get overlooked in reviews so I guess that's what happened here

test/integration_test.go
taylanisikdemir: following the same pattern where `QueryWorkflow` is used but will switch to `ts.NoError`

test/workflow_test.go
taylanisikdemir: I think isolated test cases are going to be less error prone so when we refactor we would actually do everything inside `SetupSuite` per test case.

test/workflow_test.go
taylanisikdemir: good catch. I had this workflow on a separate go app where I was actually using sugared log functions. missed to clean up this part

internal/internal_task_handlers.go
taylanisikdemir: addressed

internal/internal_task_handlers.go
taylanisikdemir: `w.err` is only set during [workflow completion](https://github.com/taylanisikdemir/cadence-client/blob/380c778f353b0cc476a5ad1c9c7a0d9949a4cf3f/internal/internal_task_handlers.go#L523). So I don't see the concern here

PR #1288, opened 2023-11-03
Introduce DisableStrictNonDeterminismCheck worker option

taylanisikdemir:
**What changed?**
Adding a new option to be able to toggle non-determinism false positives fix in #1281. 

**Why?**
The strict non-determinism checks in #1281 may cause some existing workflows to fail due to the existing buggy non-determinism checks in task handling loop. So exposing this option to disable the strict non-determinism if needed.

**How did you test it?**
 #1281 will be rebased with this change and testing will be done there.

**Potential risks**
N/A

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -131,6 +131,7 @@ type (
 		contextPropagators             []ContextPropagator
 		tracer                         opentracing.Tracer
 		workflowInterceptorFactories   []WorkflowInterceptorFactory
+		enableStrictNonDeterminism     bool
 	}
 
 	activityProvider func(name string) activity
@@ -388,7 +389,7 @@ func newWorkflowTaskHandler(
 	registry *registry,
 ) WorkflowTaskHandler {
 	ensureRequiredParams(&params)
-	return &workflowTaskHandlerImpl{
+	wth := &workflowTaskHandlerImpl{
 		domain:                         domain,
 		logger:                         params.Logger,
 		ppMgr:                          ppMgr,
@@ -402,7 +403,16 @@ func newWorkflowTaskHandler(
 		contextPropagators:             params.ContextPropagators,
 		tracer:                         params.Tracer,
 		workflowInterceptorFactories:   params.WorkflowInterceptorChainFactories,
+		enableStrictNonDeterminism:     params.WorkerBugPorts.EnableStrictNonDeterminismCheck,
 	}
+
+	traceLog(func() {
+		wth.logger.Debug("Workflow task handler is created.",
+			zap.String(tagDomain, wth.domain),
+			zap.Bool("EnableStrictNonDeterminism", wth.enableStrictNonDeterminism))
+	})
+
+	return wth
 }
 
 // TODO: need a better eviction policy based on memory usage
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -131,7 +131,7 @@ type (
 		contextPropagators             []ContextPropagator
 		tracer                         opentracing.Tracer
 		workflowInterceptorFactories   []WorkflowInterceptorFactory
-		enableStrictNonDeterminism     bool
+		disableStrictNonDeterminism    bool
 	}
 
 	activityProvider func(name string) activity
@@ -403,13 +403,13 @@ func newWorkflowTaskHandler(
 		contextPropagators:             params.ContextPropagators,
 		tracer:                         params.Tracer,
 		workflowInterceptorFactories:   params.WorkflowInterceptorChainFactories,
-		enableStrictNonDeterminism:     params.WorkerBugPorts.EnableStrictNonDeterminismCheck,
+		disableStrictNonDeterminism:    params.WorkerBugPorts.DisableStrictNonDeterminismCheck,
 	}
 
 	traceLog(func() {
 		wth.logger.Debug("Workflow task handler is created.",
 			zap.String(tagDomain, wth.domain),
-			zap.Bool("EnableStrictNonDeterminism", wth.enableStrictNonDeterminism))
+			zap.Bool("disableStrictNonDeterminism", wth.disableStrictNonDeterminism))
 	})
 
 	return wth
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -228,7 +228,7 @@ func (s *internalWorkerTestSuite) TestCreateWorker_WithAutoScaler() {
 }
 
 func (s *internalWorkerTestSuite) TestCreateWorker_WithStrictNonDeterminism() {
-	worker := createWorkerWithStrictNonDeterminismOption(s.T(), s.service)
+	worker := createWorkerWithStrictNonDeterminismDisabled(s.T(), s.service)
 	err := worker.Start()
 	require.NoError(s.T(), err)
 	time.Sleep(time.Millisecond * 200)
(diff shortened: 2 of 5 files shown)


internal/worker.go
```diff
@@ -266,6 +266,27 @@ type (
...
+	}
+
+	// WorkerBugPorts allows opt-in enabling of older, possibly buggy behavior, primarily intended to allow temporarily
+	// emulating old behavior until a fix is deployed.
+	// By default, bugs (especially rarely-occurring ones) are fixed and all users are opted into the new behavior.
+	// Back-ported buggy behavior *may* be available via these flags.
+	//
+	// Bugports are always deprecated and may be removed in future versions.
+	// Generally speaking they will *likely* remain in place for one minor version, and then they may be removed to
+	// allow cleaning up the additional code complexity that they cause.
```
Groxx, line 281: If you add `Deprecated: ...`, does vscode hint you about it? https://github.com/golang/go/issues/40447 

personally I've had that succeed and fail somewhat randomly through the years, but goland has been consistent about showing it.  and in principle it should be supported.

taylanisikdemir: I got mixed signals from codebase so left the original comments from [child-workflow-options bugport](https://github.com/taylanisikdemir/cadence-client/blob/a73fe0e8ea2f180446d828ba5786a30b0668123f/internal/workflow.go#L504). Let's use with `DEPRECATED:` format going forward as the standard. Just tried and it hints me on VSCode when the deprecated field is referred on another package.

internal/internal_task_handlers.go
```diff
@@ -402,7 +403,16 @@ func newWorkflowTaskHandler(
 		contextPropagators:             params.ContextPropagators,
 		tracer:                         params.Tracer,
 		workflowInterceptorFactories:   params.WorkflowInterceptorChainFactories,
+		enableStrictNonDeterminism:     params.WorkerBugPorts.EnableStrictNonDeterminismCheck,
 	}
+
+	traceLog(func() {
```
Groxx, line 409: I always forget that this exists...

I think it's fine here either way so I'm not requesting any change, but to think out loud I do sorta think this whole concept is better built as either:
- a const, so it can be compile-time elided.  you can change these with build tags.
- just log, and rely on the logger to do the level check efficiently (it does).  should be particularly fine with a non-sugared logger like here.

taylanisikdemir: There's a global `var enableVerboseLogging = false` which determines whether this gets logged. there's no level lower than DEBUG in zap so this has been the workaround original authors went with. I used other workaround in the past which required having own wrapper around zap and expose `Trace()` functions and filtering was based on caller file name controlled by env var. 
If you have other ideas we should discuss as part of v2 redesign/refactoring. Leaving this for now as it was still useful to have this option during local testing.

internal/worker.go
```diff
@@ -266,6 +266,32 @@ type (
...
+	// Deprecated: This is deprecated and will be removed in the future.
+	//
```
Groxx, line 285:
```suggestion
	//
	// Deprecated: This is deprecated and will be removed in the future.
```

internal/worker.go
```diff
@@ -266,6 +266,32 @@ type (
 		// Optional: Host is just string on the machine running the client
 		// default: empty string
 		Host string
+
+		// Optional: See WorkerBugPorts for more details
+		//
+		// Deprecated: This field is deprecated and will be removed in the future.
```
Groxx, line 272:
```suggestion
		// Deprecated: All bugports are always deprecated and may be removed at any time
```

internal/internal_worker_test.go
```diff
@@ -227,6 +227,14 @@ func (s *internalWorkerTestSuite) TestCreateWorker_WithAutoScaler() {
 	worker.Stop()
 }
 
+func (s *internalWorkerTestSuite) TestCreateWorker_WithStrictNonDeterminism() {
+	worker := createWorkerWithStrictNonDeterminismOption(s.T(), s.service)
+	err := worker.Start()
+	require.NoError(s.T(), err)
+	time.Sleep(time.Millisecond * 200)
```
Groxx, line 234: what's the sleep doing here?

internal/internal_worker_test.go
```diff
@@ -425,6 +433,13 @@ func createWorkerWithAutoscaler(
 	return createWorkerWithThrottle(t, service, float64(0), WorkerOptions{FeatureFlags: FeatureFlags{PollerAutoScalerEnabled: true}})
 }
 
+func createWorkerWithStrictNonDeterminismOption(
+	t *testing.T,
+	service *workflowservicetest.MockClient,
+) *aggregatedWorker {
+	return createWorkerWithThrottle(t, service, float64(0), WorkerOptions{WorkerBugPorts: WorkerBugPorts{EnableStrictNonDeterminismCheck: true}})
```
Groxx, line 440:
```suggestion
	return createWorkerWithThrottle(t, service, 0, WorkerOptions{WorkerBugPorts: WorkerBugPorts{EnableStrictNonDeterminismCheck: true}})
```
Explicitly casting a const for a typed arg seems... very strange

internal/internal_worker_test.go
```diff
@@ -425,6 +433,13 @@ func createWorkerWithAutoscaler(
 	return createWorkerWithThrottle(t, service, float64(0), WorkerOptions{FeatureFlags: FeatureFlags{PollerAutoScalerEnabled: true}})
```
Groxx, line 433:
```suggestion
	return createWorkerWithThrottle(t, service, 0, WorkerOptions{FeatureFlags: FeatureFlags{PollerAutoScalerEnabled: true}})
```
maybe this too while you're here

internal/workflow.go
```diff
@@ -481,7 +481,7 @@ type (
 		// Generally speaking they will *likely* remain in place for one minor version, and then they may be removed to
 		// allow cleaning up the additional code complexity that they cause.
 		//
-		// deprecated
+		// Deprecated: This field is deprecated and will be removed in the future.
```
Groxx, line 484:
```suggestion
		// Deprecated: All bugports are always deprecated and may be removed at any time
```

internal/workflow.go
```diff
@@ -500,7 +500,7 @@ type (
 	// Generally speaking they will *likely* remain in place for one minor version, and then they may be removed to
 	// allow cleaning up the additional code complexity that they cause.
 	//
-	// deprecated
+	// DEPRECATED: This is deprecated and will be removed in the future.
```
Groxx, line 503:
```suggestion
	// Deprecated: This is deprecated and will be removed in the future.
```

Groxx:
since we'll be enabling this by default, we probably want this flag to be reversed, yea?

doing it gradually could work too (say over a couple versions), but that'd still need to change the name or type of the field, which we essentially can't do.

internal/internal_worker_test.go
taylanisikdemir: All the tests in this file have it. Removing sleep doesn't change the result when a single test case is run but when the whole suit is run without sleeps it gets `signal: interrupt`. Not going to deep dive here as it is totally irrelevant to this change. I am not fan of using test suite and assert libraries in Go. Feel free to take a look if you have time, we can discuss later.

internal/internal_worker_test.go
taylanisikdemir: agreed

taylanisikdemir:
> since we'll be enabling this by default, we probably want this flag to be reversed, yea?
> 
> doing it gradually could work too (say over a couple versions), but that'd still need to change the name or type of the field, which we essentially can't do.

We are not enabling by default. Default will be current behavior and we will explicitly enable this on low risk environments and gradually GA and then retire this option

internal/internal_worker_test.go
Groxx: ah, yea.  following others with these suites is (unfortunately) reasonable.
I'd love to rip them all out, but it is a big task :|

Groxx (approved):
yep, looks good - changelog / finer details around explanation and whatnot can come in the implementation change, when that's fully nailed down

PR #1289, opened 2023-11-05
Fix in Test Env workflow interceptor is not propagated correctly for child workflows

shijiesheng:
**What changed?**

Workflow interceptor in child env is not populated from parent env. 

**Why?**

This is needed to correctly mock production env

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -363,6 +363,7 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(param
 	childEnv.testWorkflowEnvironmentShared = env.testWorkflowEnvironmentShared
 	childEnv.workerOptions = env.workerOptions
 	childEnv.workerOptions.DataConverter = params.dataConverter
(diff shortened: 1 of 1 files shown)


internal/internal_workflow_testsuite.go
```diff
@@ -363,6 +363,7 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(param
 	childEnv.testWorkflowEnvironmentShared = env.testWorkflowEnvironmentShared
 	childEnv.workerOptions = env.workerOptions
 	childEnv.workerOptions.DataConverter = params.dataConverter
+	childEnv.workflowInterceptors = env.workflowInterceptors
```
taylanisikdemir, line 366: This seems to be an issue for tests only. Please make it clear in PR title

taylanisikdemir (approved):

internal/internal_workflow_testsuite.go
shijiesheng: sure. For some reason my title got cut in PR and now I've fixed it.

PR #1292, opened 2023-11-07
Add GetActivityTaskList and GetWorkflowTaskList utils to replace WithTaskListMapper utils

shijiesheng:
**What changed?**

Remove With*Tasklist utils introduced in https://github.com/uber-go/cadence-client/pull/1286 
Instead, users should use the new Get*TaskList and With*TaskList APIs. 

For future as discussed with @Groxx, it's better to have a functional API to replace all "With*".

For example: 
`ExecuteChildWorkflow(ctx, type, child.Args(all, args), child.WorkflowID("asdf'))`

**Why?**
With**TaskList APIs are confusing and it's better to remove it

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1413,16 +1413,13 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
 	return ctx1
 }
 
-// WithWorkflowTaskListMapper returns a copy of context and changes workflow tasklist with mapper function.
-func WithWorkflowTaskListMapper(ctx Context, mapper func(string) string) Context {
-	ctx1 := setWorkflowEnvOptionsIfNotExist(ctx)
-	var taskList string
-	if getWorkflowEnvOptions(ctx1).taskListName != nil {
-		taskList = *getWorkflowEnvOptions(ctx1).taskListName
+// GetWorkflowTaskList retrieves current workflow tasklist from context
+func GetWorkflowTaskList(ctx Context) *string {
+	wo := getWorkflowEnvOptions(ctx)
+	if wo == nil {
+		return nil
 	}
-	newTaskList := mapper(taskList)
-	getWorkflowEnvOptions(ctx1).taskListName = &newTaskList
-	return ctx1
+	return wo.taskListName
 }
 
 // WithWorkflowID adds a workflowID to the context.
@@ -1813,12 +1810,14 @@ func WithTaskList(ctx Context, name string) Context {
 	return ctx1
 }
 
-// WithTaskListMapper makes a copy of the context and apply the tasklist mapper function
-// Note this shall not confuse with WithWorkflowTaskListMapper. This is the tasklist for activities.
-func WithTaskListMapper(ctx Context, mapper func(string) string) Context {
-	ctx1 := setActivityParametersIfNotExist(ctx)
-	getActivityOptions(ctx1).TaskListName = mapper(getActivityOptions(ctx1).TaskListName)
-	return ctx1
+// GetActivityTaskList retrieves tasklist info from context
+func GetActivityTaskList(ctx Context) *string {
+	ao := getActivityOptions(ctx)
+	if ao == nil {
+		return nil
+	}
+	tl := ao.TaskListName
+	return &tl
 }
 
 // WithScheduleToCloseTimeout adds a timeout to the copy of the context.
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1416,10 +1416,11 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
 // GetWorkflowTaskList retrieves current workflow tasklist from context
 func GetWorkflowTaskList(ctx Context) *string {
 	wo := getWorkflowEnvOptions(ctx)
-	if wo == nil {
+	if wo == nil || wo.taskListName == nil {
 		return nil
 	}
-	return wo.taskListName
+	tl := *wo.taskListName // copy
+	return common.StringPtr(tl)
 }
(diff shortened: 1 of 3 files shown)


internal/workflow.go
```diff
@@ -1413,16 +1413,14 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
...
+func GetWorkflowTaskList(ctx Context) *string {
+	wo := getWorkflowEnvOptions(ctx)
+	if wo == nil || wo.taskListName == nil {
+		return nil
 	}
-	newTaskList := mapper(taskList)
-	getWorkflowEnvOptions(ctx1).taskListName = &newTaskList
-	return ctx1
+	tl := *wo.taskListName // copy
+	return common.StringPtr(tl)
```
Groxx, line 1423:
```suggestion
	return &tl
```
might as well keep it simpler when we need that variable anyway

workflow/workflow_options.go
```diff
@@ -42,9 +42,13 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
...
 }
 
-// WithWorkflowTaskListMapper returns a copy of Context with changed tasklist
-func WithWorkflowTaskListMapper(ctx Context, mapper func(name string) string) Context {
-	return internal.WithWorkflowTaskListMapper(ctx, mapper)
+// GetWorkflowTaskList returns a copy of Context with changed tasklist
+// 1. Activity context returns the current workflow tasklist
+// 2. Child workflow context returns the child workflow tasklist
+// 3. Workflow context will return current workflow tasklist
+// 4. If Context is not a workflow context or activity context, it will return nil
```
Groxx, line 49: 2 things I think:
- indent the list so it renders in docs stuff as a list instead of one giant paragraph
- activity context is just plain wrong, it's different info (though it inherits). 

it's also just kinda generally confusing I think, nil-ness and what it returns depends on if it has been _modified_, right?  e.g. an un-changed tasklist will be nil and implies child workflows and activities will use the `GetInfo(ctx).TaskListName`

shijiesheng: Should I always return a tasklist then?

Groxx: I've kinda been internally debating about that tbh.  allowing empty/nil is more precise and lets you tell the difference between "this is falling back to the workflow's tasklist" and "this tasklist has been explicitly set", so in a number of ways I think it may actually be better.

Why anyone would _care_ about that distinction, I dunno.  And it does make things more verbose for the likely-very-common case of using the fallback.  I just tend to lean most-precise first, and add other APIs for convenience if it really seems necessary.

shijiesheng: make sense. 
For `activity context is just plain wrong, it's different info (though it inherits).` Yes, but if user passes a activity context in GetWorkflowTaskList, it still returns the workflow context and thus should be intuitive.

workflow/activity_options.go
```diff
@@ -58,10 +58,10 @@ func WithTaskList(ctx Context, name string) Context {
 	return internal.WithTaskList(ctx, name)
 }
 
-// WithTaskListMapper makes a copy of current context and update the tasklist with mapper function.
-// Note this is the tasklist for activity tasklist. For workflow tasklist, use WithWorkflowTaskListMapper
-func WithTaskListMapper(ctx Context, mapper func(string) string) Context {
-	return internal.WithTaskListMapper(ctx, mapper)
+// GetActivityTaskList returns tasklist where activity should be started
+// if context is not an activity context then it will return nil
+func GetActivityTaskList(ctx Context) *string {
```
Groxx, line 63:
```suggestion
// GetActivityTaskList returns the overridden ActivityOptions.TaskList, or an empty string if not overridden.
// If empty, an activity will use the workflow.GetInfo(ctx).TaskListName by default.
func GetActivityTaskList(ctx Context) string {
```

workflow/workflow_options.go
```diff
@@ -42,9 +42,13 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
...
 
-// WithWorkflowTaskListMapper returns a copy of Context with changed tasklist
-func WithWorkflowTaskListMapper(ctx Context, mapper func(name string) string) Context {
-	return internal.WithWorkflowTaskListMapper(ctx, mapper)
+// GetWorkflowTaskList returns a copy of Context with changed tasklist.
+//  1. Activity context returns the current workflow tasklist
+//  2. Child workflow context returns the child workflow tasklist
+//  3. Workflow context will return current workflow tasklist
+//  4. If Context is not a workflow context, it will return nil
+func GetWorkflowTaskList(ctx Context) *string {
```
Groxx, line 50:
```suggestion
// GetWorkflowTaskList returns the current ChildWorkflowOptions.TaskList,
// which defaults to workflow.GetInfo(ctx).TaskListName if not set.
func GetWorkflowTaskList(ctx Context) string {
```

Groxx (approved):
After some discussion in chat and digging around in the details on the values / how they're used:

Internals look correct (good catch on the copy), but since there are common code-paths that'll lead to nil, empty-string, and different-value... let's just collapse it all and return "get me the tasklist that will be used to start an activity/workflow" from these public methods.  It's simpler to use and we're pretty unlikely to need finer-grained distinctions.  Suggestions cover that, it just needs straightforward implementation changes.

If we do discover we need to tell "set" vs "not set" apart, we can add another set of funcs, or just expose ActivtyOptions (or something similar) for richer inspection.  We're just not confident that exposing those directly is always safe, so we're avoiding that for now.

PR #1293, opened 2023-11-14
Create separate worker usage data collection and move hardware emit there

timl3136:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**


internal/internal_worker_usage_collector.go
```diff
@@ -0,0 +1,126 @@
...
+		logger:       logger,
+		ctx:          ctx,
+		cancel:       cancel,
+		wg:           &sync.WaitGroup{},
+	}
+}
+
+func (w *workerUsageCollector) Start() {
+	w.wg.Add(1)
+	go func() {
```
3vilhamster, line 63: Do we need to spawn a goroutine per worker? Why not ensure only 1 running?

timl3136: Only the hardware emitting is once per host, all other metrics will be worker-specific. (e.g activity poll response vs. decision poll response)

3vilhamster: For now I see only w.collectHardwareUsage() which will just spawn bunch of data into the same scope. I would suggest separating hardware emitter and worker specific metrics.

timl3136: That's current design, for each type of metrics based on their origin, I will create a separate gorountine for each of them. But they would be contained under a single workerusagecollector so that their result can be collected and sent in one place

internal/internal_worker_base.go
```diff
@@ -173,17 +170,29 @@ func newBaseWorker(options baseWorkerOptions, logger *zap.Logger, metricsScope t
 			logger,
 		)
 	}
+	// for now it's default to be enabled
+	var workerUC *workerUsageCollector
```
taylanisikdemir, line 174: no need this extra line to declare variable.

```
workerUC := newWorkerUsageCollector(..)
```

internal/internal_worker_usage_collector.go
```diff
@@ -0,0 +1,126 @@
...
+		defer func() {
+			if p := recover(); p != nil {
+				w.logger.Error("Unhandled panic in workerUsageCollector.")
+				w.logger.Error(p.(error).Error())
```
taylanisikdemir, line 67: `p.(error)`is not safe cast. also let's not produce multiple lines here. you can combine these in single error log. see other places where `recover()` is handled

internal/internal_worker_usage_collector.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+		}()
+		defer w.wg.Done()
+		ticker := time.NewTicker(w.cooldownTime)
```
taylanisikdemir, line 71: let's add `defer ticker.Stop()`

internal/internal_worker_usage_collector.go
```diff
@@ -0,0 +1,126 @@
...
+		for {
+			select {
+			case <-w.ctx.Done():
+				return
+			case <-ticker.C:
+				// Given that decision worker and activity worker are running in the same host, we only need to collect
+				// hardware usage from one of them.
+				if w.workerType == "DecisionWorker" {
```
taylanisikdemir, line 79: this might not be future proof and also if customer is running separate processes for decision and activity workers then we will not have the hardware usage of those hosts that only runs activity workers. we should also not create no-op `workerUsageCollector`s if only one of them will do the work.
@Groxx what would be your recommendation for host level metric reporting on the client side? I would like to avoid global static variables but this use case probably requires one.

timl3136: We tried Sync.Once before, but that would cause issues with unit testing as it will just wait indefinitely for this routine to stop while blocking all other goroutine from closing

internal/internal_worker_base.go
timl3136: Resolved

internal/internal_worker_usage_collector.go
timl3136: Resolved

internal/internal_worker_usage_collector.go
timl3136: Resolved

internal/internal_worker_usage_collector.go
taylanisikdemir: you can override it in the unit tests
```
type once interface {
   Do(func())
}

var collectHardwareUsageOnce once
```

in typical startup this would be set to `sync.Once`:
```
collectHardwareUsageOnce = sync.Once{}
```

in test code you can initialize this to a fake implementation
```
collectHardwareUsageOnce = myFakeOnce{}  // myFakeOnce implements Do(func())
```

timl3136: Thank you for your suggestion. I have implemented that in the latest commit

taylanisikdemir: I don't see `EmitOnce` being used in workerUsageCollector. We should only have one (singleton) instance of `workerUsageCollector` which would be lazily created by the first worker instance. Rest of the workers would create a noOpUsageCollector. This lazy initialization logic should be hidden from workers. Worker just calls `newWorkerUsageCollector()` and that function should determine whether it's first time or not. Let's discuss offline  if more clarification needed.

timl3136: In our usecase, only the hardware info are once per host collected. Other worker type (decision worker and activity worker) should have different workerUsageCollector as they track different task type behaviors.

internal/internal_worker_usage_collector.go
```diff
@@ -0,0 +1,151 @@
...
+			if p := recover(); p != nil {
+				w.metricsScope.Counter(metrics.WorkerUsageCollectorPanic).Inc(1)
+				topLine := fmt.Sprintf("WorkerUsageCollector panic for workertype: %v", w.workerType)
+				st := getStackTraceRaw(topLine, 7, 0)
+				w.logger.Error("WorkerUsageCollector panic.",
+					zap.String(tagPanicError, fmt.Sprintf("%v", p)),
+					zap.String(tagPanicStack, st))
+			}
+		}()
+		defer w.wg.Done()
```
taylanisikdemir, line 84: there are a few things problematic about this goroutine closure
1. this `wg.Done()` will be called once goroutine for `go w.runHardwareCollector()` is started. It shouldn't be marked as done until `runHardwareCollector()` terminates so should be moved there
2. no need for a panic recovery here
3. no need for a goroutine to invoke `runHardwareCollector`.

internal/internal_worker_usage_collector.go
taylanisikdemir: what type of information are you planning to collect per worker basis in this `workerUsageCollector`?

timl3136: Tasklist backlog/poll response since decision and activity worker have their own pollers and that need to be scaled independently

internal/worker.go
```diff
@@ -271,6 +271,12 @@ type (
 		//
 		// Deprecated: All bugports are always deprecated and may be removed at any time.
 		WorkerBugPorts WorkerBugPorts
+
+		// Optional: This implementation ensures that a specific function is executed only once per instance.
+		// The mechanism can be overridden by other interfaces that implement the 'Do()' method.
+		//
+		// default: nil, that would ensure some functions are executed only once
+		Sync oncePerHost
```
taylanisikdemir, line 279: This is user visible worker options. we shouldn't expose `oncePerHost` here.

timl3136: If we do not exposed that here, we might need to insert that as part of the `workerExecutionParameters` and add the Sync.Once as part of the parameter into function "NewWorker" and "newAggregatedWorker". What do you think about that idea?

taylanisikdemir: it makes sense to add to `workerExecutionParameters` as it is not exposed outside. I'd also recommend looking at `WithSomeOption` pattern in go https://www.sohamkamani.com/golang/options-pattern/.

PR #1295, opened 2023-11-21
Add `__query_types` query type

arzonus:
**What changed?**

A new built-in query type `__query_type` has been added. A result of the processing of the query type must be an array of strings. The list must contain all built-in query types (included this one) and all custom query types of the workflow.

**Why?**
The changes are required to implement [the task](https://github.com/uber/cadence/issues/382), so it allows users to retrieve the list of available query types. 

**How did you test it?**
The changes were tested locally. TestProcessQuery_KnownQueryTypes was added. 

**Potential risks**
There are no expected risks. The worst potential risk is the query functionality will not work.

### DIFF
--- a/internal/internal_workflow_test.go
+++ b/internal/internal_workflow_test.go
@@ -1308,3 +1308,39 @@ func (t *tracingInterceptor) ExecuteWorkflow(ctx Context, workflowType string, a
 	t.trace = append(t.trace, "ExecuteWorkflow "+workflowType+" end")
 	return result
 }
+
+type WorkflowOptionTest struct {
+	suite.Suite
+}
+
+func TestWorkflowOption(t *testing.T) {
+	suite.Run(t, new(WorkflowOptionTest))
+}
+
+func (t *WorkflowOptionTest) TestKnowQueryType_NoHandlers() {
+	wo := workflowOptions{queryHandlers: make(map[string]func([]byte) ([]byte, error))}
+	t.ElementsMatch(
+		[]string{
+			QueryTypeStackTrace,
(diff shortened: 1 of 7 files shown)


3vilhamster (approved):
LGTM!

internal/internal_workflow_test.go
```diff
@@ -1308,3 +1308,39 @@ func (t *tracingInterceptor) ExecuteWorkflow(ctx Context, workflowType string, a
...
+type WorkflowOptionTest struct {
+	suite.Suite
+}
+
+func TestWorkflowOption(t *testing.T) {
+	suite.Run(t, new(WorkflowOptionTest))
+}
+
+func (t *WorkflowOptionTest) TestKnowQueryType_NoHandlers() {
+	wo := workflowOptions{queryHandlers: make(map[string]func([]byte) ([]byte, error))}
```
dkrotx, line 1321: nit: if this would be like
```
wo := workflowOptions{
  queryHandlers: map[string]func([]byte) ([]byte, error){},
}
```

this would be more obvious it is empty.
+ consider renaming to TestKnowQueryType_No**User**Handlers or something like this. Then it is clear there is no other handlers other than default.

dkrotx (approved):
1 nit comment

internal/internal_workflow_test.go
3vilhamster: I think both are good to read.
Though, Uber style guide is against your proposed approach. https://github.com/uber-go/guide/blob/master/style.md#initializing-maps

PR #1299, opened 2023-11-23
Release v1.2.6

arzonus:


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## [v1.2.6] - 2023-11-24
+### Added
+- Added a new query type `__query_type` #1295
```
dkrotx, line 10: `query_types`

internal/version.go
```diff
@@ -43,7 +43,7 @@ package internal
 // Due to all of this unreliability, this should be used as strictly informational
 // metadata, e.g. for caller version monitoring, never behavioral (use
 // FeatureVersion or feature flags instead).
-const LibraryVersion = "1.1.0"
+const LibraryVersion = "1.2.6"
```
dkrotx, line 46: Why is it 1.1.0 -> 1.2.6? Is it to align with something?

dkrotx (changes requested):

internal/version.go
arzonus: Due to pushing invalid versions, the latest version is still 1.2.5, but it's incorrect and we need to release a greater version to update the version in proxy.golang.org, so the next one is 1.2.6.

dkrotx: got it!

dkrotx (approved):

PR #1300, opened 2023-11-28
Remove warning message that regarding unknown event type

timl3136:
**What changed?**
Remove warning message for unknown event type

**Why?**
It's creating too much warning log that overwhelm system

**How did you test it?**
Unit test

**Potential risks**
Only remove logger so no danger

### DIFF
--- a/internal/internal_utils.go
+++ b/internal/internal_utils.go
@@ -468,8 +468,6 @@ func estimateHistorySize(logger *zap.Logger, event *s.HistoryEvent) int {
 			sum += len(event.SignalExternalWorkflowExecutionInitiatedEventAttributes.Control)
 			sum += len(event.SignalExternalWorkflowExecutionInitiatedEventAttributes.Input)
 		}
-
-	default: // Do not fail to be forward compatible with new events
 	}
 
 	return sum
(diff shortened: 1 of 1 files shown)


internal/internal_utils.go
```diff
@@ -469,9 +469,7 @@ func estimateHistorySize(logger *zap.Logger, event *s.HistoryEvent) int {
 			sum += len(event.SignalExternalWorkflowExecutionInitiatedEventAttributes.Input)
 		}
 
-	default:
-		logger.Warn("unknown event type", zap.String("Event Type", event.GetEventType().String()))
-		// Do not fail to be forward compatible with new events
+	default: // Do not fail to be forward compatible with new events
```
taylanisikdemir, line 472: you can remove default case since it's no-op

taylanisikdemir (approved):

internal/internal_utils.go
timl3136: Done

internal/internal_utils.go
```diff
@@ -468,10 +468,6 @@ func estimateHistorySize(logger *zap.Logger, event *s.HistoryEvent) int {
 			sum += len(event.SignalExternalWorkflowExecutionInitiatedEventAttributes.Control)
 			sum += len(event.SignalExternalWorkflowExecutionInitiatedEventAttributes.Input)
 		}
-
-	default:
-		logger.Warn("unknown event type", zap.String("Event Type", event.GetEventType().String()))
```
3vilhamster, line 473: Maybe it could be useful to have this log with Debug level?

timl3136: Good idea, let me change that to debug level logs

PR #1303, opened 2023-12-07
Write coverage output to a static path

taylanisikdemir:
**What changed?**
Unit test coverage output is currently written to a file with potentially non-static path. e.g. `.build/go1.21.2_darwin_arm64/coverage/unit_test_cover.out`
Some 3rd party coverage reporting tools expects a static path to integrate. 
The change is to also write the coverage results to a predictable path `.build/cover.out`

**Why?**
To easily integrate with 3rd party coverage reporting tools.

**How did you test it?**
`make unit_test`


Groxx (approved):
Yea, coverage seems worth breaking out of that per-version sandbox.  Particularly since it's not an intermediate / nothing depends on it, just outside readers that would benefit from stability (sonarqube, IDEs, etc).

SGTM 👍

PR #1304, opened 2023-12-13
Support two-legged OAuth flow

mantas-sidlauskas:
**What changed?**
Adding AuthorizationProvider which retrieves token from external provider

**Why?**
This allows to provide access token which is retrieved from configured, external auth provider (Okta, Auth0, AWS Cognito, etc.)
This is two legged authentication flow when client sends ID and secret to auth provider.
Token will be sent to cadence server where it has to be validated to allow or deny access.

**How did you test it?**

**Potential risks**
None


internal/oauth_authorization.go
```diff
@@ -0,0 +1,67 @@
+package internal
```
taylanisikdemir, line 1: license thing is missing here

internal/oauth_authorization.go
```diff
@@ -0,0 +1,67 @@
...
+	// This needs to be provided for some OAuth providers
+	EndpointParams map[string]string `yaml:"endpointParams"`
+}
+
+type AuthorizationProvider struct {
+	tokenSource oauth2.TokenSource
+	config      OAuthAuthorizerConfig
+}
+
+func NewOAuthAuthorizationProvider(config OAuthAuthorizerConfig) auth.AuthorizationProvider {
```
taylanisikdemir, line 38: please add unit tests

internal/oauth_authorization.go
```diff
@@ -0,0 +1,67 @@
...
+
```
taylanisikdemir, line 39: nit: remove empty line here

internal/oauth_authorization.go
mantas-sidlauskas: test are added

internal/oauth_authorization.go
mantas-sidlauskas: done

go.mod
```diff
@@ -24,7 +24,8 @@ require (
 	go.uber.org/thriftrw v1.25.0
 	go.uber.org/yarpc v1.55.0
 	go.uber.org/zap v1.13.0
-	golang.org/x/net v0.7.0
+	golang.org/x/net v0.19.0
+	golang.org/x/oauth2 v0.15.0
```
taylanisikdemir, line 28: @Groxx do these version upgrades look fine to you? would cause issues with monorepo?

worker/worker.go
```diff
@@ -356,3 +359,8 @@ func SetBinaryChecksum(checksum string) {
 func NewAdminJwtAuthorizationProvider(privateKey []byte) AuthorizationProvider {
 	return internal.NewAdminJwtAuthorizationProvider(privateKey)
 }
+
+// NewOAuthAuthorizationProvider asks for a token from external OAuth provider
+func NewOAuthAuthorizationProvider(config OAuthConfig) AuthorizationProvider {
```
taylanisikdemir, line 364: changes LGTM but will leave final approval to @Groxx as I have limited context.

go.mod
davidporter-id-au: I think these are breaking

davidporter-id-au: @mantas-sidlauskas I think due to the google cloud archiver, we're going to have a hard time getting past changes here:
```
// not sure we can currently upgrade past this:
cloud.google.com/go/iam v0.3.0
```

davidporter-id-au: @mantas-sidlauskas  I think upgrading this bumps the IAM library which breaks some other thing in our dep tree, at the moment I don't know we can upgrade to this min version: 
```
go get golang.org/x/oauth2@v0.15.0
Setting GOPROXY to https://goproxy.uberinternal.com
go: downloading golang.org/x/oauth2 v0.15.0
go: upgraded cloud.google.com/go v0.104.0 => v0.110.2
go: upgraded cloud.google.com/go/compute v1.7.0 => v1.20.1
go: upgraded cloud.google.com/go/iam v0.3.0 => v0.13.0 // breaks, we can't currently get past 0.3.0
...
```

Groxx:
tbh I'm not seeing why this is internal at all, this seems pretty straightforward to build externally (where it'd be able to add whatever dependencies it needs, and wouldn't imply ongoing support and maintenance here).  the same is true of the jwt authorizer - it's just a client-RPC-interface wrapper, that's trivial to build externally.

is there something about this that _requires_ internal code?  that's not part of this PR, so I'm not sure if anything would prevent it from just being another external-friendly wrapper.

go.mod
```diff
@@ -55,13 +56,14 @@ require (
...
-	golang.org/x/tools v0.1.12 // indirect
+	golang.org/x/mod v0.8.0 // indirect
+	golang.org/x/sys v0.15.0 // indirect
+	golang.org/x/text v0.14.0 // indirect
+	golang.org/x/tools v0.6.0 // indirect
+	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20200212174721-66ed5ce911ce // indirect
 	google.golang.org/grpc v1.28.0 // indirect
-	google.golang.org/protobuf v1.26.0-rc.1 // indirect
+	google.golang.org/protobuf v1.31.0 // indirect
```
Groxx, line 66: (gomod in general, not just this line)
current set looks pretty reasonable 👍

I'm not certain it won't have conflicts internally, but that's a separate issue / we should upgrade libraries there too if at all possible.

mantas-sidlauskas:
> tbh I'm not seeing why this is internal at all, assuming this is going to end up similarly to the jwt authorizer - it's just a client-RPC-interface wrapper, that's trivial to build externally. (the same is true of the jwt authorizer, but that ship has sailed. at least until we remove it.)
> 
> is there something about this that _requires_ internal code? that's not part of this PR, so I'm not sure if anything would prevent it from just being another external-friendly wrapper.
> 
> edit: ah, we have... some recently changed oauth stuff in the server too 🤔. hmm. that too would be MUCH better broken out into a "real" plugin, rather than the hardcoded half-plugin it currently is. then this'd all be pretty easy to do separately.

100% agree on the plugin way. Anyway, we can move this later. 
Regarding the "internal" part: if OAuth validation is turned on, internal RPC communication needs to be validated also. Right now, self-signed JWT with admin privileges [are generated](https://github.com/uber/cadence/blob/7ac5a84fb5da253b8a0b8d0a7d62a28af182271b/common/rpc/outbounds.go#L97). This PR is adding external token support for internal Cadence RPC calls as well.

3vilhamster (approved):
LGTM

PR #1306, opened 2023-12-19
downgrade thrift to backend compatible version

3vilhamster:
**What changed?**
I downgraded apache/thrfit version to the version compatible on the Cadence backend.

**Why?**
Cadence backend uses client to run internal operations, so client should be compatible with the backend.

**How did you test it?**
unit tests/integration tests

**Potential risks**
Potential failures for customers using up to date thrift version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -111,3 +111,6 @@ retract (
 	v0.3.0 // Published accidentally
 	v0.2.0 // Published accidentally
 )
+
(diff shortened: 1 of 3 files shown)


3vilhamster:
hmm in the monorepo the version is set to 0.17.0 and it client works.
Let me try 0.17.0 then, might be it will be fine.

PR #1308, opened 2024-01-18
Replace JWT library for signing tokens

mantas-sidlauskas:
**What changed?**
* Extending `Claims` with "Registered claims" from https://www.rfc-editor.org/rfc/rfc7519#section-4.1
* Defining issuer for easier key detection on server-side
* Adding `ExpiresAt` but also keeping TTL for backwards compatibility

**Why?**
To align with upcoming server changes

**How did you test it?**
Unit tests

**Potential risks**


internal/common/auth/service_wrapper.go
```diff
@@ -30,7 +31,8 @@ import (
 )
 
 const (
-	jwtHeaderName = "cadence-authorization"
+	jwtHeaderName     = "cadence-authorization"
+	JWTInternalIssuer = "internal-jwt"
```
3vilhamster, line 35: I propose separating internal constants from external.

https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations

internal/common/auth/service_wrapper.go
```diff
@@ -22,6 +22,7 @@ package auth
 
 import (
 	"context"
+	"github.com/golang-jwt/jwt/v5"
```
3vilhamster, line 25: separate std imports from others.

Unrelated @Groxx Should we onboard client to gci too?

internal/jwt_authorization.go
```diff
@@ -40,24 +40,28 @@ func NewAdminJwtAuthorizationProvider(privateKey []byte) auth.AuthorizationProvi
...
-	}
 	key, err := util.LoadRSAPrivateKey(j.PrivateKey)
 	if err != nil {
 		return nil, err
 	}
-	signer, err := jwt.NewSignerRS(jwt.RS256, key)
-	if err != nil {
-		return nil, err
+
+	ttl := int64(60 * 10)
```
3vilhamster, line 48: why it is hardcoded? Should it be configurable?

internal/jwt_authorization_test.go
```diff
@@ -21,60 +21,41 @@
 package internal
 
 import (
-	"encoding/json"
-	"io/ioutil"
+	"github.com/stretchr/testify/assert"
```
3vilhamster, line 24: goimports

3vilhamster (commented):
Have some questions, but overall looks ok.

internal/jwt_authorization_test.go
mantas-sidlauskas: done

internal/common/auth/service_wrapper.go
mantas-sidlauskas: thanks, moving this const to another package, this doesn't belong here.

internal/jwt_authorization.go
mantas-sidlauskas: No sure why this was hard-coded during the [initial implementation](https://github.com/uber-go/cadence-client/pull/1116) but my take is this: this ttl is for internal/self-signed tokens. There are no external calls to acquire a token, there's no caching involved, they are issued for every request, in other words, they are cheap and should be short lived.

3vilhamster (approved):
LGTM

PR #1309, opened 2024-01-31
Added CloseTime filter to shadower

agautam478:
**What changed?**
Added an extra filter in the shadow options to filter the workflows on the basis of Closed Time.

**Why?**
These changes have been made to make some QOL changes to the shadower. Currently shadower only filters on the basis of start time of the workflows. The expectation is that including a close time filter will enable better throttling options to the users.

**How did you test it?**
Unit test, tested with a shadow test in cadence-samples.

**Potential risks**
NA

### DIFF
--- a/internal/query_builder.go
+++ b/internal/query_builder.go
@@ -70,11 +70,14 @@ var (
 )
 
 type (
-	// QueryBuilder builds visibility query
+	// QueryBuilder builds visibility query. It's shadower's own Query builders that processes the shadow filter
+	// options into a query to pull t he required workflows.
+
 	QueryBuilder interface {
 		WorkflowTypes([]string) QueryBuilder
 		WorkflowStatus([]WorkflowStatus) QueryBuilder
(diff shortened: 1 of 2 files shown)


internal/query_builder.go
```diff
@@ -131,6 +134,19 @@ func (q *queryBuilderImpl) StartTime(minStartTime, maxStartTime time.Time) Query
 	return q
 }
 
+func (q *queryBuilderImpl) CloseTime(minCloseTime, maxCloseTime time.Time) QueryBuilder {
+	CloseTimeQueries := make([]string, 0, 2)
+	if !minCloseTime.IsZero() {
```
timl3136, line 139: nit: maybe add some comments behind the reason to append the new string into queries.

timl3136 (approved):

internal/query_builder.go
agautam478: This is to signify min and max time stamp. Will add the log in the next PR. Thanks

PR #1312, opened 2024-02-20
Add documentation for propagators and how they are executed

alshopov:
Clarify behavior on return of error
Document the whole execution sequence

Docs added to both internal package and exposed type alias due to: https://github.com/golang/go/issues/44905

### DIFF
--- a/internal/headers.go
+++ b/internal/headers.go
@@ -36,8 +36,24 @@ type HeaderReader interface {
 	ForEachKey(handler func(string, []byte) error) error
 }
 
-// ContextPropagator is an interface that determines what information from
-// context to pass along
+// ContextPropagator determines what information from context to pass along.
+//
+// The information passed is called Headers - a sequence of string to []byte
+// tuples of serialized data that should follow workflow and activity execution
+// around.
+//
+// Inject* methods are used on the way from the process to persistence in
+// Cadence - thus they use HeaderWriter-s to write the metadata. Extract*
+// methods are used on the way from persisted state in Cadence to execution
+// - thus they use HeaderReader-s to read the metadata and fill it in the
+// returned context. Returning error from Extract* methods prevents the
+// successful workflow run.
+//
+// The whole sequence of execution is:
+//
+// Process initiating the workflow -> Inject -> Cadence -> Go Workflow Worker
+// -> ExtractToWorkflow -> Start executing a workflow -> InjectFromWorkflow ->
+// Cadence -> Go Activity Worker -> Extract -> Execute Activity
 type ContextPropagator interface {
 	// Inject injects information from a Go Context into headers
 	Inject(context.Context, HeaderWriter) error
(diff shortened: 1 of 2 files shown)


internal/headers.go
```diff
@@ -36,8 +36,25 @@ type HeaderReader interface {
 	ForEachKey(handler func(string, []byte) error) error
 }
 
-// ContextPropagator is an interface that determines what information from
-// context to pass along
+// ContextPropagator determines what information from context to pass along.
```
taylanisikdemir, line 39: thanks for filling in this documentation gap. @shijiesheng can you review please?

internal/headers.go
```diff
@@ -36,8 +36,25 @@ type HeaderReader interface {
...
+// Returning an error from any of its methods will prevent the normal execution
```
shijiesheng, line 40: This is actually not true. At least this Inject method will fail open.

3vilhamster: @alshopov Could you update the doc here? Or should we handle it on our side?

internal/headers.go
```diff
@@ -36,8 +36,24 @@ type HeaderReader interface {
...
+// methods are used on the way from persisted state in Cadence to execution
+// - thus they use HeaderReader-s to read the metadata and fill it in the
+// returned context. Returning error from Extract* methods prevents the
+// successful workflow run.
+//
+// The whole sequence of execution is:
+//
+// Process initiating the workflow -> Inject -> Cadence -> Go Workflow Worker
+// -> ExtractToWorkflow -> Start executing a workflow -> InjectFromWorkflow ->
+// Cadence -> Go Activity Worker -> Extract -> Execute Activity
```
shijiesheng, line 56: nit: this might make it easier for users to understand
```suggestion
// -> ExtractToWorkflow -> Start executing a workflow -> Schedule an activity -> InjectFromWorkflow ->
// Cadence -> Go Activity Worker -> Extract -> Execute Activity
```

workflow/context_propagator.go
```diff
@@ -29,7 +29,23 @@ type (
...
+	// Cadence - thus they use HeaderWriter-s to write the metadata. Extract*
+	// methods are used on the way from persisted state in Cadence to execution
+	// - thus they use HeaderReader-s to read the metadata and fill it in the
+	// returned context. Returning error from Extract* methods prevents the
+	// successful workflow run.
+	//
+	// The whole sequence of execution is:
+	//
+	// Process initiating the workflow -> Inject -> Cadence -> Go Workflow Worker
+	// -> ExtractToWorkflow -> Start executing a workflow -> InjectFromWorkflow ->
```
shijiesheng, line 48: ditto

3vilhamster (approved):
LGTM

internal/headers.go
3vilhamster: I think that it is out of the scope of the context propagator, so we might add it if it causes confusion.

PR #1315, opened 2024-02-29
Revert breaking changes in v1.2.8

ketsiambaku:
**What changed?**
Reverting the 2 commits below in this order:
https://github.com/uber-go/cadence-client/pull/1256
https://github.com/uber-go/cadence-client/pull/1257

**Why?**
At least one of these commits caused unit test failures for Uber internal cadence users

**How did you test it?**
unit test

**Potential risks**

### DIFF
--- a/internal/internal_workflow_testsuite.go
+++ b/internal/internal_workflow_testsuite.go
@@ -362,7 +362,6 @@ func (env *testWorkflowEnvironmentImpl) newTestWorkflowEnvironmentForChild(param
 	childEnv.startedHandler = startedHandler
 	childEnv.testWorkflowEnvironmentShared = env.testWorkflowEnvironmentShared
 	childEnv.workerOptions = env.workerOptions
-	childEnv.workflowInterceptors = env.workflowInterceptors
 	childEnv.workerOptions.DataConverter = params.dataConverter
 	childEnv.workflowInterceptors = env.workflowInterceptors
 	childEnv.registry = env.registry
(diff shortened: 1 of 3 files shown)


agautam478:
Can you confirm which one is causing failures? These were minor changes introduced to enable effective mocking. One is a complete test file and the other is a mini fix that makes registering Workflow and Activity registration optional when they are mocked.

ketsiambaku:
> Can you confirm which one is causing failures? These were minor changes introduced to enable effective mocking. One is a complete test file and the other is a mini fix that makes registering Workflow and Activity registration optional when they are mocked.

@agautam478 
We observed unit test failures due to Duplicate Registration errors. #1256 seems to set DisableAlreadyRegisteredCheck flag to true so I thought it was a good candidate. reverting this did in fact fix the test failures
I reverted #1257 because I felt it was built on top of the first PR but reverting one might just be enough in this case.

jakobht (approved):

ketsiambaku:
@edmondop I thought you should be aware :( We will be happy to re-review your PRs

> > Can you confirm which one is causing failures? These were minor changes introduced to enable effective mocking. One is a complete test file and the other is a mini fix that makes registering Workflow and Activity registration optional when they are mocked.
> 
> @agautam478 We observed unit test failures due to Duplicate Registration errors. #1256 seems to set DisableAlreadyRegisteredCheck flag to true so I thought it was a good candidate. reverting this did in fact fix the test failures I reverted #1257 because I felt it was built on top of the first PR but reverting one might just be enough in this case.

edmondop:
Thanks, the change with the interceptor should not be able to break anything.

@agautam478 can you explain the issue you are seeing? Were you by chance registering an activity twice? I can look into it

PR #1325, opened 2024-03-25
Enforce 85% new line coverage

taylanisikdemir:
**What changed?**
Updating codecov's `patch` setting to enforce new line coverage is above 85%.
Similar to what we have on Cadence repo: https://github.com/uber/cadence/pull/5805

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -19,7 +19,10 @@ coverage:
         if_ci_failed: ignore   # require the CI to pass before setting the status
     patch:
       default:
-        informational: true
+        target: 85%            # specify the target coverage for each commit status
+                               #   option: "auto" (compare against parent commit or pull request base)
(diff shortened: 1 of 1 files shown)


Groxx (approved):
sgtm 👍 

how do we work around this if it's too much red tape?  there's always the admin-merge of course, but is there another option?

taylanisikdemir:
> sgtm 👍
> 
> how do we work around this if it's too much red tape? there's always the admin-merge of course, but is there another option?

I didn't see a bypass option via tag etc. exposed by codecov. Current options are:
- Admin merge
- Change the threshold in codecov.yml in the PR and then revert it back as follow up

Groxx:
reasonably discoverable workarounds anyway.  makes sense 👍

PR #1321, opened 2024-03-18
Update idl and add wrapper implementaton for async start/signalwithstart APIs

taylanisikdemir:
**What changed?**
Updating client to expose newly added async APIs:
- `SignalWithStartWorkflowExecutionAsync`
- `StartWorkflowExecutionAsync`

Steps followed to update idls
- Update submodule
```
cd idls
git checkout 34b4519b270945fbb38f413137dfb50b5fc20af7
```
- Update idl go dependency
```
go get github.com/uber/cadence-idl@34b4519b270945fbb38f413137dfb50b5fc20af7
```
- Build and fix failures
```
make build
make unit_test
make integ_test_sticky_off
```

**How did you test it?**
Added unit tests for new methods

PR #1327, opened 2024-03-29
Update client wrappers with new async APIs

taylanisikdemir:
**What changed?**
Updating client wrappers to expose new async methods. 
- SignalWithStartWorkflowExecutionAsync
- StartWorkflowExecutionAsync

These changes were missed in #1321.

**How did you test it?**
Unit test


client/client.go
```diff
@@ -98,6 +98,15 @@ type (
 		//	- InternalServiceError
 		StartWorkflow(ctx context.Context, options StartWorkflowOptions, workflowFunc interface{}, args ...interface{}) (*workflow.Execution, error)
 
+		// StartWorkflowAsync behaves like StartWorkflow except that the request is first queued and then processed asynchronously.
+		// See StartWorkflow for parameter details.
+		// The returned AsyncWorkflowExecution doesn't contain run ID, because the workflow hasn't started yet.
+		// The errors it can return:
+		//	- EntityNotExistsError, if domain does not exists
+		//	- BadRequestError
+		//	- InternalServiceError
```
davidporter-id-au, line 107: Are we returning InternalServiceError in thrift? This should not be included in this set since it's an unexpected error (it can happen, but there's no valid case where this should be sent to a user)

client/client.go
```diff
@@ -163,6 +172,15 @@ type (
 		SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
 			options StartWorkflowOptions, workflowFunc interface{}, workflowArgs ...interface{}) (*workflow.Execution, error)
 
+		// SignalWithStartWorkflowAsync behaves like SignalWithStartWorkflow except that the request is first queued and then processed asynchronously.
+		// See SignalWithStartWorkflow for parameter details.
+		// The errors it can return:
+		//  - EntityNotExistsError, if domain does not exist
+		//  - BadRequestError
+		//	- InternalServiceError
```
davidporter-id-au, line 180: same, we shouldn't ever return this intentionally

client/client.go
taylanisikdemir: Not sure what these errors in the comment mean to you but it says "The errors it can return" to hint the caller's err handling. All of [these](https://github.com/uber-go/cadence-client/blob/a73fe0e8ea2f180446d828ba5786a30b0668123f/internal/compatibility/thrift/error.go#L34) are possible technically This comment is doomed to be out-dated though. I just kept it similar to corresponding sync API method so they are at least consistently incomplete :)

internal/workflow.go
```diff
@@ -400,6 +400,11 @@ type (
 		RunID string
 	}
 
+	// WorkflowExecutuonAsync Details.
+	WorkflowExecutionAsync struct {
+		ID string
```
Groxx, line 405: probably `WorkflowID` instead?  it's meaningful that it's not some other ID, and particularly that it's not a RunID.

internal/workflow.go
```diff
@@ -400,6 +400,11 @@ type (
 		RunID string
 	}
 
+	// WorkflowExecutuonAsync Details.
```
Groxx, line 403:
```suggestion
	// WorkflowExecutionAsync Details.
```

internal/internal_workflow_client.go
```diff
@@ -155,98 +155,65 @@ func (wc *workflowClient) StartWorkflow(
...
-	memo, err := getWorkflowMemo(options.Memo, wc.dataConverter)
 	if err != nil {
 		return nil, err
 	}
 
-	searchAttr, err := serializeSearchAttributes(options.SearchAttributes)
-	if err != nil {
-		return nil, err
+	if wc.metricsScope != nil {
+		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, startRequest.WorkflowType.GetName())
```
Groxx, line 181:
```suggestion
		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, *startRequest.WorkflowType.Name)
```
maybe?  both because it's guaranteed to exist (so we might as well enforce it), and because there's a `*startRequest.WorkflowId` below (also guaranteed to exist).

internal/internal_workflow_client.go
```diff
@@ -255,13 +222,13 @@ func (wc *workflowClient) StartWorkflow(
 	}
 
 	if wc.metricsScope != nil {
-		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, workflowType.Name)
+		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, startRequest.WorkflowType.GetName())
```
Groxx, line 225:
```suggestion
		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, *startRequest.WorkflowType.Name)
```
same reasoning

internal/internal_workflow_client.go
```diff
@@ -155,98 +155,65 @@
...
+		ID:    *startRequest.WorkflowId,
+		RunID: response.GetRunId(),
 	}
+	return executionInfo, nil
+}
 
-	jitterStartSeconds := common.Int32Ceil(options.JitterStart.Seconds())
-	if jitterStartSeconds < 0 {
-		return nil, errors.New("Invalid JitterStart option")
+// StartWorkflowAsync queues a workflow execution which is going to be picked up and started by Cadence backend asynchronously.
```
Groxx, line 192: Reword to match SignalWithStartWorkflowAsync?  I think that one's a bit clearer.

internal/internal_workflow_client_test.go
```diff
@@ -1443,3 +1703,150 @@ func (m *PartialCancelRequestMatcher) Matches(a interface{}) bool {
...
+			options: StartWorkflowOptions{
+				ID:                              workflowID,
+				TaskList:                        tasklist,
+				ExecutionStartToCloseTimeout:    10 * time.Second,
+				DecisionTaskStartToCloseTimeout: 5 * time.Second,
+				DelayStart:                      0 * time.Second,
+				JitterStart:                     0 * time.Second,
+			},
+			workflowFunc: func(ctx Context, a, b int) {}, // this causes error because args not provided
+			wantErr:      "expected 2 args for function",
```
Groxx, line 1821: tbh I think I'd pass an explicitly-empty args, or just one arg or something.

relying on the zero value when we haven't seen it for over 100 lines (it's in fact completely unused in ALL of these tests) seems a bit strange.

internal/internal_workflow_client_test.go
```diff
@@ -1443,3 +1703,150 @@ func (m *PartialCancelRequestMatcher) Matches(a interface{}) bool {
...
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			service := workflowservicetest.NewMockClient(mockCtrl)
+			wc, ok := NewClient(service, domain, &ClientOptions{
+				Identity: "test-identity",
+			}).(*workflowClient)
+
+			if !ok {
+				t.Fatalf("expected NewClient to return a *workflowClient, but got %T", wc)
+			}
```
Groxx, line 1835:
```suggestion
			wc := NewClient(service, domain, &ClientOptions{
				Identity: "test-identity",
			}).(*workflowClient)
```
the panic already tells you the same info, and even slightly more.  simpler to just let it bubble.

internal/internal_workflow_client_test.go
```diff
@@ -1443,3 +1703,150 @@ func (m *PartialCancelRequestMatcher) Matches(a interface{}) bool {
...
+
+			gotReq, err := wc.getWorkflowStartRequest(context.Background(), "", tc.options, tc.workflowFunc, tc.args...)
+			if tc.wantErr != "" {
+				assert.ErrorContains(t, err, tc.wantErr)
+				return
+			}
+
+			assert.NoError(t, err)
+
+			// set the fields that are not set in the expected request
```
Groxx, line 1845:
```suggestion
			// set randomized fields for easier Equal comparing
```
to be a bit clearer about what's going on here / why "not set" means "copy from X" instead of just being left as the zero value.

Groxx (changes requested):
gonna request some relatively minor things, as there's some odd inconsistencies / unrelated checks / tests that don't really test anything.

overall looks good though, and I'm curious if you have any reasons for/against adding another metrics tag for async APIs.

client/client.go
davidporter-id-au: Yeah, I was referring to the comment. 

This is a long-running thrift thing where thift's somewhat surprising way of requiring upfront declaration of expected errors meant that people declared all errors as expected (equivalent to putting all http error codes into the 4XX range). 

Why should we care? The reason is that - a while back anyway, this messed up Muttley's ability to differentiate caller errors from callee errors (at least in thrift). So, among other things, SLAs got incorrectly higher uptime due to returning 5XX errors like 4XX errors. 

Anyway, I am probably just droning on at this point. The takeaway is that if we have any thrift APIs, we shouldn't include 5XX class errors there in the expected section (of which InternalError is one)

internal/workflow.go
taylanisikdemir: I agree on more explicit name such as `WorkflowID` but want to keep this similar to `WorkflowExecution` struct above. Inconsistency across similar domain objects would cause confusion for users

internal/workflow.go
taylanisikdemir: fixed

internal/internal_workflow_client.go
taylanisikdemir: makes sense.

internal/internal_workflow_client.go
taylanisikdemir: I wrote this first and then a while after wrote the docs for `SignalWithStartWorkflowAsync`. Missed to update here looks like. addressing

internal/internal_workflow_client.go
```diff
@@ -445,13 +388,13 @@ func (wc *workflowClient) SignalWithStartWorkflow(ctx context.Context, workflowI
 	}
 
 	if wc.metricsScope != nil {
-		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, workflowType.Name)
+		scope := wc.metricsScope.GetTaggedScope(tagTaskList, options.TaskList, tagWorkflowType, signalWithStartRequest.WorkflowType.GetName())
 		scope.Counter(metrics.WorkflowSignalWithStartCounter).Inc(1)
```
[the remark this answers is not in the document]
taylanisikdemir: good catch. I actually want separate metric

internal/internal_workflow_client_test.go
```diff
@@ -1204,10 +1209,51 @@ func (s *workflowClientTestSuite) TestStartWorkflow_WithMemoAndSearchAttr() {
...
+		ExecutionStartToCloseTimeout:    timeoutInSeconds,
+		DecisionTaskStartToCloseTimeout: timeoutInSeconds,
+	}
+	f1 := func(ctx Context, r []byte) string {
+		return "result"
+	}
+
+	_, err := client.StartWorkflow(context.Background(), options, f1, []byte("test"))
+	s.Equal(getDefaultDataConverter(), client.dataConverter)
+	s.Error(err)
```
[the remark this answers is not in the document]
taylanisikdemir: not intentional. it was leftover from copy/paste. down side of suite based approach

internal/internal_workflow_client_test.go
```diff
@@ -1204,10 +1209,51 @@ func (s *workflowClientTestSuite) TestStartWorkflow_WithMemoAndSearchAttr() {
...
+		TaskList:                        tasklist,
+		ExecutionStartToCloseTimeout:    timeoutInSeconds,
+		DecisionTaskStartToCloseTimeout: timeoutInSeconds,
+	}
+	wf := func(ctx Context) string {
+		return "result"
+	}
+	startResp := &shared.StartWorkflowExecutionResponse{}
+
+	s.service.EXPECT().StartWorkflowExecution(gomock.Any(), gomock.Any(), gomock.Any()).Return(startResp, errors.New("failed")).AnyTimes()
```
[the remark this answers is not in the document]
taylanisikdemir: yes I need 1+ times assertion.

internal/internal_workflow_client_test.go
taylanisikdemir: yep agreed. I missed putting explicitly empty args

internal/internal_workflow_client_test.go
taylanisikdemir: I am guessing panic would not give me the actual type which I'm explicitly adding in log message with `%T`. Either way it's a minor thing, leaving as is.

internal/internal_workflow_client_test.go
taylanisikdemir: added more clarity

internal/internal_workflow_client_test.go
```diff
@@ -1239,7 +1285,221 @@ func (s *workflowClientTestSuite) SignalWithStartWorkflowWithMemoAndSearchAttr()
...
+		Do(func(_ interface{}, asyncReq *shared.SignalWithStartWorkflowExecutionAsyncRequest, _ ...interface{}) {
+			req := asyncReq.Request
+			var resultMemo, resultAttr string
+			err := json.Unmarshal(req.Memo.Fields["testMemo"], &resultMemo)
+			s.NoError(err)
+			s.Equal("memo value", resultMemo)
+
+			err = json.Unmarshal(req.SearchAttributes.IndexedFields["testAttr"], &resultAttr)
+			s.NoError(err)
+			s.Equal("attr value", resultAttr)
```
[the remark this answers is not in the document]
taylanisikdemir: agree but leaving as a future improvement. same code block is repeated in multiple tests

internal/internal_workflow_client_test.go
```diff
@@ -1239,7 +1285,221 @@ func (s *workflowClientTestSuite) SignalWithStartWorkflowWithMemoAndSearchAttr()
...
+	// Pass a context with a deadline so error retry doesn't take forever
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	_, err := s.client.SignalWithStartWorkflowAsync(ctx, "wid", "signal", "value", options, wf)
+	s.Error(err)
+}
+
+func (s *workflowClientTestSuite) TestStartWorkflowAsync() {
+	client, ok := s.client.(*workflowClient)
+	s.True(ok)
```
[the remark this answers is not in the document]
taylanisikdemir: yeah don't see the value here. this pattern exist in a lot of existing test methods in this file. fixing

internal/internal_workflow_client_test.go
```diff
@@ -1239,7 +1285,221 @@ func (s *workflowClientTestSuite) SignalWithStartWorkflowWithMemoAndSearchAttr()
...
+	_, err := client.StartWorkflowAsync(context.Background(), options, f1, []byte("test"))
+	s.Equal(getDefaultDataConverter(), client.dataConverter)
+	s.NoError(err)
+}
+
+func (s *workflowClientTestSuite) TestStartWorkflowAsync_WithDataConverter() {
+	dc := newTestDataConverter()
+	s.client = NewClient(s.service, domain, &ClientOptions{DataConverter: dc})
+	client, ok := s.client.(*workflowClient)
+	s.True(ok)
```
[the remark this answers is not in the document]
taylanisikdemir: yep. followed the pattern in other tests without too much thinking on best way to rewrite these tests. addressing this unnecessary assertion since it's a small/local change

internal/internal_workflow_client_test.go
```diff
@@ -1239,7 +1285,221 @@ func (s *workflowClientTestSuite) SignalWithStartWorkflowWithMemoAndSearchAttr()
...
+	}
+	f1 := func(ctx Context, r []byte) string {
+		return "result"
+	}
+	input := []byte("test")
+
+	s.service.EXPECT().StartWorkflowExecutionAsync(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil).
+		Do(func(_ interface{}, asyncReq *shared.StartWorkflowExecutionAsyncRequest, _ ...interface{}) {
+			req := asyncReq.Request
+			dc := client.dataConverter
```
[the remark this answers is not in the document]
taylanisikdemir: great find. I'm not focused on improving tests for data converter in this PR. will leave it out of scope. feel free to send a follow up PR with the fix before we forget this test gap.

internal/internal_workflow_client_test.go
```diff
@@ -1063,15 +1064,19 @@ func (s *workflowClientTestSuite) TestSignalWithStartWorkflow_Error() {
 	signalInput := []byte("my signal input")
 	options := StartWorkflowOptions{}
 
-	resp, err := s.client.SignalWithStartWorkflow(context.Background(), workflowID, signalName, signalInput,
-		options, workflowType)
+	// Pass a context with a deadline so error retry doesn't take forever
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	resp, err := s.client.SignalWithStartWorkflow(ctx, workflowID, signalName, signalInput, options, workflowType)
```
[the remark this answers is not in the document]
taylanisikdemir: > because it retries up to the timeout, and it's not test-injectable.
yeah unfortunately. 
I was about to make the retry policy overridable but went this ctx deadline route. 
Overall test duration reduced as well because it was spending 1m retrying.

internal/internal_workflow_client_test.go
```diff
@@ -1204,10 +1209,51 @@ func (s *workflowClientTestSuite) TestStartWorkflow_WithMemoAndSearchAttr() {
...
+	f1 := func(ctx Context, r []byte) string {
+		return "result"
+	}
+
+	_, err := client.StartWorkflow(context.Background(), options, f1, []byte("test"))
+	s.Equal(getDefaultDataConverter(), client.dataConverter)
+	s.Error(err)
+}
+
+func (s *workflowClientTestSuite) TestStartWorkflow_Error() {
```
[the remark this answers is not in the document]
taylanisikdemir: I went with existing `_Error()` suffix but I'm also not happy with this. _RPCError() makes sense

client/client.go
taylanisikdemir: IIUC you are saying all "expected" errors should fall into 4xx range which is how it should be. But that doesn't change the fact that there will be 500s (unexpected cases) and it's possible these functions may return InternalServiceError. 
I'd support removing these partial list of error types from the client code comment to avoid confusions until we revisit error propagation/handling and document it better for users.

internal/internal_workflow_client_test.go
Groxx: It does. You get the interface type, the concrete type in the interface, and the target type-asserted type.

internal/workflow.go
Groxx: ah, yeah, just barely out of diff-context view.
matching that makes sense 👍

Groxx (approved):
I'll make a task or something for the followups 👍  I might just tackle that today.

PR #1328, opened 2024-03-29
Pin mockery and regenerate everything

Groxx:
Seems like this has been left un-pinned and un-documented for a long time.
Might as well fix that, so we can address the new async APIs in e.g. #1327 in a standardized way.


Makefile
```diff
@@ -185,7 +189,7 @@ $(if $(wildcard THRIFT_FILES),,$(error idls/ submodule must exist, or build will
 endef
 
 # codegen is done when thrift is done (it's just a naming-convenience, $(BUILD)/thrift would be fine too)
-$(BUILD)/codegen: $(BUILD)/thrift | $(BUILD)
+$(BUILD)/codegen: | $(BUILD)
```
Groxx, line 192: just unnecessary since it's defined above, and re-definitions merge dependencies

taylanisikdemir (approved):

davidporter-id-au (approved):

PR #1329, opened 2024-03-30
tidy

Groxx:
- **Pin mockery and regenerate everything**
- **makefile**
- **might be finished**
- **minor polish to regen less**
- **no need for manual copyright run**
- **Server-like `make build` and ensuring builds are clean in CI**


Makefile
```diff
@@ -268,7 +273,10 @@ $Q +$(MAKE) --no-print-directory $(addprefix $(BUILD)/,$(1))
 endef
 
 .PHONY: build
-build: $(BUILD)/dummy ## ensure all packages build
+build: $(BUILD)/fmt ## ensure all packages build
+	go build ./...
+	$Q # caution: some errors are reported on stdout for some reason
+	go test -exec true ./... >/dev/null
```
davidporter-id-au, line 279: any reason for >/dev/null here?

scripts/golint.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set -ex
+
+make tidy
+make build
+
+# intentionally capture stderr, so status-errors are also PR-failing.
+# in particular this catches "dubious ownership" failures, which otherwise
+# do not fail this check and the $() hides the exit code.
+if [ -n "$(git status --porcelain  2>&1)" ]; then
```
davidporter-id-au, line 11: nit: generally prefer `[[` and it's not needed, but also set -o pipefail

scripts/golint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
```
davidporter-id-au, line 1: shouldn't this be `/bin/bash`?

davidporter-id-au (approved):

Makefile
Groxx: Just that it's quite noisy, printing every package at it builds it, and somewhat misleading as it looks like test runs but it didn't run any actual tests

taylanisikdemir (approved):

scripts/golint.sh
Groxx: I'm _pretty_ sure I do still need to capture output with `$()`, and bash / `[[` don't change this semantic in that sense... but this is more than niche enough that I haven't actually tried it so now I'm curious.  let's see what happens.

Groxx: ... and I'm struggling to reproduce exactly what I thought I remembered 🤔 

gonna just stick with what the server does here since it's reasonably well proven over there, and also seems to work correctly here.

bash + pipefail does seem promising enough to check at another time though.

Makefile
```diff
@@ -165,10 +164,6 @@ $(BIN)/mockery: internal/tools/go.mod
 $(BIN)/copyright: internal/cmd/tools/copyright/licensegen.go
 	go build -mod=readonly -o $@ ./internal/cmd/tools/copyright/licensegen.go
 
-# dummy binary that ensures most/all packages build, without needing to wait for tests.
-$(BUILD)/dummy: $(ALL_SRC) $(BUILD)/go_mod_check
-	go build -mod=readonly -o $@ internal/cmd/dummy/dummy.go
```
Groxx, line 170: `go build ./...` does this same check substantially better

PR #1337, opened 2024-05-20
Update compatibility adapter to support new enum value

Shaddoll:
**What changed?**
- Update idl version
- Update compatibility adapter to support new enum value

**Why?**

**How did you test it?**
unit tests

**Potential risks**


internal/compatibility/proto/enum.go
```diff
@@ -364,6 +364,8 @@ func CancelExternalWorkflowExecutionFailedCause(t *shared.CancelExternalWorkflow
 	switch *t {
 	case shared.CancelExternalWorkflowExecutionFailedCauseUnknownExternalWorkflowExecution:
 		return apiv1.CancelExternalWorkflowExecutionFailedCause_CANCEL_EXTERNAL_WORKFLOW_EXECUTION_FAILED_CAUSE_UNKNOWN_EXTERNAL_WORKFLOW_EXECUTION
+	case shared.CancelExternalWorkflowExecutionFailedCauseWorkflowAlreadyCompleted:
+		return apiv1.CancelExternalWorkflowExecutionFailedCause_CANCEL_EXTERNAL_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_COMPLETED
 	}
 	panic("unexpected enum value")
```
taylanisikdemir, line 370: Not as part of this PR but we should revisit the idea of panic'ing here. This makes the client not forward-compatible. Especially for the fields that are only used for informational purposes we don't need to be strict like this. 
@Groxx WDYT?

taylanisikdemir (approved):

internal/compatibility/proto/enum.go
Shaddoll: personally, I think falling back to a default value is a safe option and it makes sense

Groxx: main thing that makes me actually _prefer_ panics is that they tend to reliably fail tests.  `assert.Error(t, err)` will happily ignore `return errors.New("impossible: unrecognized enum")` despite that being clearly intended to be a "must fail tests" value.  they also tend to be auto-categorized as "severe problem, alert oncall" because panics simply shouldn't be happening.

so mistakes tend to be noticed and fixed faster.

they are rather clearly more dangerous... but bad data is often worse, as it's so hard to fix and tends to go unnoticed for a long time.

---

personally I think things like this should be immediately alerted, so we probably need some kind of universally-recognized "this is a flaw that we could not prevent at compile time" marker.  historically I've used `ProgrammerError(..)` types, and hooked them up to always emit a metric when they're returned anywhere.  it worked pretty well.

we would need to add those checks and switch to error wrapping (so we don't lose the type) and also exclude them from _all_ `assert.Error` checks to keep them as identifiable as panics tho.

PR #1340, opened 2024-05-30
Documentation fix in MinConcurrentDecisionTaskPollers

MrBoss276:
**What changed?**
Have changed the description of MinConcurrentDecisionTaskPollers.

**Why?**
It was giving the opposite meaning of what it was supposed to give. It was depicted that changing this value would not take effect if PollerAutoScaler is set to true, but it was supposed to be the opposite.

**How did you test it?**
Just a documentation change

**Potential risks**
None

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -104,7 +104,7 @@ type (
 		MaxConcurrentDecisionTaskPollers int
 
 		// optional: Sets the minimum number of goroutines that will concurrently poll the
-		// cadence-server to retrieve decision tasks. If FeatureFlags.PollerAutoScalerEnabled is set to true,
+		// cadence-server to retrieve decision tasks. Unless FeatureFlags.PollerAutoScalerEnabled is set to true,
 		// changing this value will NOT affect the rate at which the worker is able to consume tasks from a task list.
 		// Default value is 1
 		MinConcurrentDecisionTaskPollers int


taylanisikdemir (approved):

zawadzkidiana:
Thanks for this, @subhash-veluru — and apologies for the year-plus wait.

You're right about the inversion: the comment on `MinConcurrentDecisionTaskPollers` says "If `FeatureFlags.PollerAutoScalerEnabled` is set to true, changing this value will NOT affect..." when it should read "**Unless** ... is set to true, changing this value will NOT affect..." — as written, it claims the setting stops mattering exactly when autoscaling is *off*, which is backwards. Checking current `master`, the same backwards sentence is still there verbatim on `MinConcurrentDecisionTaskPollers` (`internal/worker.go`), so the fix is still needed and still correct.

Worth noting for context: the sibling field, `MinConcurrentActivityTaskPollers`, already carries the corrected wording ("Changing this value will NOT affect the rate ... unless FeatureFlags.PollerAutoScalerEnabled is set to true") from an unrelated, later change — so this PR would bring the decision-task field's doc back in line with its activity-task counterpart, which is a good consistency argument on its own.

One more thing that's changed since this was opened: both fields have since picked up `// Deprecated: NO effect and use AutoScalerOptions instead.` — so the option this comment describes no longer does anything either way. That doesn't make the fix wrong, just slightly less consequential than when it was filed; it's still worth landing so the doc isn't actively misleading for anyone reading it before they reach the deprecation notice.

This PR already has reviewers requested (@Groxx, @jakobht, @dkrotx, @3vilhamster, @shijiesheng, @agautam478, @demirkayaender) but nobody's weighed in yet — it's a one-line, low-risk doc fix, so hopefully a quick rebase (it's showing a few commits behind `master` now) and a maintainer glance is all this needs. Thanks for taking the time to report and fix this.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1342, opened 2024-06-08
expose APIs to get registered workflows and activities

ketsiambaku:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**


internal/internal_worker.go
```diff
@@ -776,6 +776,30 @@ type aggregatedWorker struct {
...
+	return aw.registry.GetWorkflowFn(registerName)
+}
+
+func (aw *aggregatedWorker) GetRegisteredActivities() []string {
+	return aw.registry.GetRegisteredActivities()
+}
+
+func (aw *aggregatedWorker) GetActivityAlias(fnName string) (string, bool) {
+	return aw.registry.GetActivityAlias(fnName)
+}
```
Groxx, line 797: digging around in stuff:
do you think we have a need for this one?  the alias-vs-registered-name details are kinda subtle and possibly fragile.

e.g. in particular this would only let a caller find out if there is an alias for their _full_ inferred function name... which they don't have a way to get, because GetRegisteredActivities returns the registered name, which is the alias if there is an alias.

I kinda wonder if what we might want here is a `GetActivityName(any)` that lets you pass in a func or string, and get the name-or-alias that should match something in `GetRegisteredActivities()`.  since that's essentially how we use it internally.  though that's potentially even more fragile...

ketsiambaku: Yeah I get your point.
My thought process was that it could be useful to show an alias to function (or alias to function name) mapping for debugging as it can get confusing with EnableShortName option for example. 
I think we might be better off without it now that you point it out. 

We can get the registered names from GetRegisteredActivities() and we can use it to GetActivityFn() and some reflect magic on the empty interface can give us the original function name. Thus, we do have a alias to function name mapping

internal/registry.go
```diff
@@ -124,6 +124,39 @@ func (r *registry) RegisterActivityWithOptions(af interface{}, options RegisterA
...
+}
+
+func (r *registry) GetActivityAlias(fnName string) (string, bool) {
+	return r.getActivityAlias(fnName)
+}
+
+func (r *registry) GetActivityFn(registerName string) (interface{}, bool) {
+	a, ok := r.GetActivity(registerName)
+	if !ok {
+		return nil, ok
```
dkrotx, line 155: style nit: better to explicitely return false instead of ok.

dkrotx (approved):

internal/registry.go
ketsiambaku: fixed

PR #1347, opened 2024-06-18
Partial fix for Continue as new case

agautam478:
**What changed?**
Removed the additional checks for the cases where the workflows continued as new. The check sometimes clashes with the completed status and there is no point in matching the inputs to this granularity if a wf continues as new and a workflows completed are almost similar. If there was a  timeout failure in the workflow it would be before it continued as new and it makes no sense to match it.

In another scenario, the identification of the last event is not reliable. `last := events[len(events)-1]`
So in case of an early return or when the events have something else going on you don't actually get the real last event. if this is not reliable then we would be doing the checks unnecessarily.

**Why?**
In an attempt to fix the false negtaives thrown by the replayer in continue as new cases.

**How did you test it?**
unit tests

**Potential risks**
more false negatives.

### DIFF
--- a/internal/workflow_replayer.go
+++ b/internal/workflow_replayer.go
@@ -339,10 +339,13 @@ func (r *WorkflowReplayer) replayWorkflowHistory(
 		return err
 	}
 
-	if last.GetEventType() != shared.EventTypeWorkflowExecutionCompleted && last.GetEventType() != shared.EventTypeWorkflowExecutionContinuedAsNew {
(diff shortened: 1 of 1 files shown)


Groxx (approved):
yea, I think this implementation is just too flawed to worry about too much.  previous code is causing problems because it's _even more_ flawed, so let's just start getting rid of things to simplify and get a bit closer to what we want.

when we have more reliable replay checks, we can revisit some of this.  and likely make it configurable, since this behavior absolutely does not make sense in all reasonable scenarios.

PR #1349, opened 2024-06-22
Adding tests for internal_workflow_client

dkrotx:
Added tests for internal_workflow_client

To increase code coverage

unit-tests

no risk


internal/internal_workflow_client_test.go
```diff
@@ -1756,6 +1768,704 @@ func (s *workflowClientTestSuite) TestTerminateWorkflow() {
...
+				s.service.EXPECT().
+					QueryWorkflow(gomock.Any(), gomock.Any(), gomock.Any()).
+					Do(func(_ context.Context, req *shared.QueryWorkflowRequest, _ ...yarpc.CallOption) {
+						// do common validation for some fields
+						s.Equal(domain, req.GetDomain())
+						s.Equal(workflowID, req.GetExecution().GetWorkflowId())
+						s.Equal(runID, req.GetExecution().GetRunId())
+						s.Equal(queryType, req.GetQuery().GetQueryType())
+						// ... and custom validation for the testcase
+						tt.requestValidator(req)
```
3vilhamster, line 2452: nit: you can check for nil and avoid this call if it is not provided.
nit: you can also move common checks to a function and explicitly provide it as a default value.

side nit: each test should focus on the use case, so common validation might not even be required.

3vilhamster (approved):
LGTM.

internal/internal_workflow_client_test.go
dkrotx: Good points!
Addressed all 3. Speaking of the last (focus) - I also prefer when fail-tests are not duplicating the validation.

PR #1352, opened 2024-07-01
Handle panics while polling for tasks

natemort:
The goroutines started in doPoll execute the logic for actually making the poll RPCs, as well as any wrapping layers around them. Notably the proto conversion logic contains panics if there are unexpected values, and if we fail to handle these panics the application crashes.

**What changed?**
- Add handling for panics while polling

**Why?**
- Prevent services from crashing due to unexpected behavior

**How did you test it?**
- Tested via unit tests

**Potential risks**

These panics indicate a pretty core disconnect between the client and the server state, and if encountered it's unlikely that the situation will resolve without a change to the server or the client. It seems safer to indefinitely retry these panicking requests and hope that a server-side change will resolve the issue than crashing the worker, but workers in this state will likely have a higher RPS than they would otherwise. Cadence's rate limiting support should be able to mitigate risk from that.

### DIFF
--- a/internal/internal_task_pollers.go
+++ b/internal/internal_task_pollers.go
@@ -63,7 +63,9 @@ type (
 
 	// basePoller is the base class for all poller implementations
 	basePoller struct {
-		shutdownC <-chan struct{}
+		shutdownC    <-chan struct{}
+		metricsScope *metrics.TaggedScope
+		logger       *zap.Logger
 	}
 
 	// workflowTaskPoller implements polling/processing a workflow task
@@ -75,8 +77,6 @@ type (
 		service      workflowserviceclient.Interface
 		taskHandler  WorkflowTaskHandler
 		ldaTunnel    *locallyDispatchedActivityTunnel
-		metricsScope *metrics.TaggedScope
-		logger       *zap.Logger
 
 		stickyUUID                   string
 		disableStickyExecution       bool
@@ -97,8 +97,6 @@ type (
 		identity            string
 		service             workflowserviceclient.Interface
 		taskHandler         ActivityTaskHandler
-		metricsScope        *metrics.TaggedScope
-		logger              *zap.Logger
 		activitiesPerSecond float64
 		featureFlags        FeatureFlags
 	}
@@ -123,10 +121,8 @@ type (
 
 	localActivityTaskPoller struct {
 		basePoller
-		handler      *localActivityTaskHandler
-		metricsScope tally.Scope
-		logger       *zap.Logger
-		laTunnel     *localActivityTunnel
+		handler  *localActivityTaskHandler
+		laTunnel *localActivityTunnel
 	}
 
 	localActivityTaskHandler struct {
@@ -256,9 +252,19 @@ func (bp *basePoller) doPoll(
 	ctx, cancel, _ := newChannelContext(context.Background(), featureFlags, chanTimeout(pollTaskServiceTimeOut))
 
 	go func() {
+		defer close(doneC)
+		defer func() {
+			if p := recover(); p != nil {
+				bp.metricsScope.Counter(metrics.InternalPanicCounter).Inc(1)
+				st := getStackTraceRaw("base poller [panic]:", 7, 0)
+				bp.logger.Error("Unhandled panic.",
+					zap.String(tagPanicError, fmt.Sprintf("%v", p)),
+					zap.String(tagPanicStack, st))
+				err = newPanicError(p, st)
+			}
+		}()
+		defer cancel()
 		result, err = pollFunc(ctx)
-		cancel()
-		close(doneC)
 	}()
 
 	select {
@@ -279,15 +285,17 @@ func newWorkflowTaskPoller(
 	params workerExecutionParameters,
 ) *workflowTaskPoller {
 	return &workflowTaskPoller{
-		basePoller:                   basePoller{shutdownC: params.WorkerStopChannel},
+		basePoller: basePoller{
+			shutdownC:    params.WorkerStopChannel,
+			metricsScope: metrics.NewTaggedScope(params.MetricsScope),
+			logger:       params.Logger,
+		},
 		service:                      service,
 		domain:                       domain,
 		taskListName:                 params.TaskList,
 		identity:                     params.Identity,
 		taskHandler:                  taskHandler,
 		ldaTunnel:                    ldaTunnel,
-		metricsScope:                 metrics.NewTaggedScope(params.MetricsScope),
-		logger:                       params.Logger,
 		stickyUUID:                   uuid.New(),
 		disableStickyExecution:       params.DisableStickyExecution,
 		StickyScheduleToStartTimeout: params.StickyScheduleToStartTimeout,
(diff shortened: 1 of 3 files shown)


internal/internal_task_pollers.go
```diff
@@ -279,15 +285,17 @@ func newWorkflowTaskPoller(
 	params workerExecutionParameters,
 ) *workflowTaskPoller {
 	return &workflowTaskPoller{
-		basePoller:                   basePoller{shutdownC: params.WorkerStopChannel},
+		basePoller: basePoller{
+			shutdownC:    params.WorkerStopChannel,
+			metricsScope: metrics.NewTaggedScope(params.MetricsScope),
```
3vilhamster, line 290: I think this should be tagged by domain. So we can create an alert on our side capturing both the service failing and the domain that is affected.

internal/internal_task_pollers.go
```diff
@@ -256,9 +252,19 @@ func (bp *basePoller) doPoll(
 	ctx, cancel, _ := newChannelContext(context.Background(), featureFlags, chanTimeout(pollTaskServiceTimeOut))
 
 	go func() {
+		defer close(doneC)
+		defer func() {
+			if p := recover(); p != nil {
+				bp.metricsScope.Counter(metrics.WorkerPanicCounter).Inc(1)
```
3vilhamster, line 258: We should have a dedicated metric here since WorkerPanicCounter is caused by panics in user-defined code, but this panic is caused by Cadence client internals.

internal/internal_task_pollers.go
```diff
@@ -256,9 +252,19 @@ func (bp *basePoller) doPoll(
 	ctx, cancel, _ := newChannelContext(context.Background(), featureFlags, chanTimeout(pollTaskServiceTimeOut))
 
 	go func() {
+		defer close(doneC)
+		defer func() {
```
taylanisikdemir, line 256: I'd move panic handling to the top (so it's the first defer entry in the stack) to be future proof. e.g. if the `close(doneC)` may cause panic if it's closed before already.

internal/internal_task_pollers.go
taylanisikdemir: The metrics emitted by this `metricsScope` already has domain, tasklist and service tags.

internal/internal_task_pollers.go
```diff
@@ -256,9 +252,19 @@ func (bp *basePoller) doPoll(
...
+			if p := recover(); p != nil {
+				bp.metricsScope.Counter(metrics.WorkerPanicCounter).Inc(1)
+				st := getStackTraceRaw("base poller [panic]:", 7, 0)
```
taylanisikdemir, line 259: can you do a soft validation on the stackTrace to make sure that omitting 7 lines is actually trimming right amount of lines from the stack and leaving the correct caller as the top of the stack?

internal/internal_task_pollers.go
natemort: My thought was that we should always have the close be the final statement to ensure that there isn't a race condition. Otherwise we might be changing the result to add the err after the other goroutine is done waiting.

internal/internal_task_pollers.go
natemort: Done

internal/internal_task_pollers.go
natemort: I just verified it, apparently 7 is the magic number regardless of the call site. Every use of this function passes the same value.

internal/internal_task_pollers.go
taylanisikdemir: I don't see a race condition here. just to make sure we are clear, this panic recovery will not catch panics from the initiator routine.

taylanisikdemir: synced offline. we don't need `doneC`, we can just use `ctx.Done()` instead. and the defer order should be
- defer cancel()
- defer func() { recover() ..) }

internal/internal_task_pollers.go
```diff
@@ -279,15 +283,17 @@ func newWorkflowTaskPoller(
 	params workerExecutionParameters,
 ) *workflowTaskPoller {
 	return &workflowTaskPoller{
-		basePoller:                   basePoller{shutdownC: params.WorkerStopChannel},
+		basePoller: basePoller{
+			shutdownC:    params.WorkerStopChannel,
+			metricsScope: metrics.NewTaggedScope(params.MetricsScope),
+			logger:       params.Logger,
```
Groxx, line 289: might be worth adding a tag (to both) to make it clear which thing is panicking... though the stack trace will probably make that clear too.

internal/internal_task_pollers_test.go
```diff
@@ -54,3 +59,79 @@ func TestLocalActivityPanic(t *testing.T) {
...
+		WorkerStopChannel: workerStopChannel,
+		WorkerOptions: WorkerOptions{
+			Identity: "identity",
+			Logger:   zaptest.NewLogger(t),
+		},
+	})
+	service.EXPECT().PollForActivityTask(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, _ *shared.PollForActivityTaskRequest, opts ...yarpc.CallOption) (*shared.PollForActivityTaskResponse, error) {
+		workerStopChannel <- struct{}{}
+		blockForever := make(chan struct{}, 1)
+		<-blockForever
```
Groxx, line 104: should still be something that's cleared out after the test is done, otherwise it leaks test goroutines (we _should_ be using [goleak](https://pkg.go.dev/go.uber.org/goleak), but there are a number of other test leaks at the moment and we haven't spent time on that yet).

if it's only called once, it's very easy to just move this outside the `DoAndReturn` and defer a `close(block)`.

internal/internal_task_pollers.go
```diff
@@ -252,17 +248,25 @@ func (bp *basePoller) doPoll(
...
-	doneC := make(chan struct{})
 	ctx, cancel, _ := newChannelContext(context.Background(), featureFlags, chanTimeout(pollTaskServiceTimeOut))
 
 	go func() {
+		defer cancel()
+		defer func() {
+			if p := recover(); p != nil {
+				bp.metricsScope.Counter(metrics.InternalPanicCounter).Inc(1)
+				st := getStackTraceRaw("base poller [panic]:", 7, 0)
+				bp.logger.Error("Unhandled panic.",
```
Groxx, line 259:
```suggestion
				bp.logger.Error("Unhandled panic",
```
just because there's generally no punctuation like that in logs

internal/internal_task_pollers.go
```diff
@@ -252,17 +248,25 @@ func (bp *basePoller) doPoll(
 	var err error
 	var result interface{}
 
-	doneC := make(chan struct{})
 	ctx, cancel, _ := newChannelContext(context.Background(), featureFlags, chanTimeout(pollTaskServiceTimeOut))
```
Groxx, line 252: while you're here:
```suggestion
	ctx, cancel, _ := newChannelContext(context.Background(), featureFlags, chanTimeout(pollTaskServiceTimeOut))
	defer cancel()

```
and then I'd just drop the explicit `cancel()` code in the select case further down.
cancel is safe to call any number of times, and it's a bit more future-proof to do it immediately.

internal/internal_task_pollers_test.go
```diff
@@ -54,3 +59,79 @@ func TestLocalActivityPanic(t *testing.T) {
...
+
+		WorkerStopChannel: workerStopChannel,
+		WorkerOptions: WorkerOptions{
+			Identity: "identity",
+			Logger:   zaptest.NewLogger(t),
+		},
+	})
+	service.EXPECT().PollForActivityTask(gomock.Any(), gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, _ *shared.PollForActivityTaskRequest, opts ...yarpc.CallOption) (*shared.PollForActivityTaskResponse, error) {
+		workerStopChannel <- struct{}{}
+		blockForever := make(chan struct{}, 1)
```
taylanisikdemir, line 103: let's define this in outer block and notify it after PollTask results are asserted to avoid hanging this call forever.

taylanisikdemir (approved):

internal/internal_task_pollers.go
natemort: Added, good call

davidporter-id-au (approved):

davidporter-id-au (approved):

PR #1356, opened 2024-07-09
Add workerStats as worker options to collect stats on poller start/stop events

ketsiambaku:
---
**Detailed Description**

This PR introduces a new worker option _EventMonitor_ which wraps a LifeCycle interface to collect stats on poller start and stop (in this PR) and running activities and workflows (in a follow-up PR) for debugging purposes.

The change is backward compatible with older versions of the client because a default implementation is created when a worker is instantiated without this option being provided. A constructor for the LifeCycle interface is also added to the public API to serve as basic guidelines for usage by the OS community.

**Impact Analysis**

**Testing Plan**
- **Unit Tests**: [Do we have unit test covering the change?] The change is covered by unit tests
- Local tests with worker service

**Rollout Plan**
The new exposed types are marked as deprecated at the moment. 
The plan is to pre-release and experiment in Uber internal for sometime before removing the deprecated annotation
Revert plan: The change can be safely reverted. Need coordination with consumers code using the introduced worker option

---


internal/internal_worker.go
```diff
@@ -304,7 +309,8 @@ func newWorkflowTaskWorkerInternal(
 		taskWorker:        localActivityTaskPoller,
 		identity:          params.Identity,
 		workerType:        "LocalActivityWorker",
-		shutdownTimeout:   params.WorkerStopTimeout},
+		shutdownTimeout:   params.WorkerStopTimeout,
+		statsCollector:    params.StatsCollector},
```
taylanisikdemir, line 313: nit: let's move the closing curly bracket to new line to better indent the contents of `baseWorkerOptions`

internal/internal_worker.go
```diff
@@ -281,7 +285,8 @@ func newWorkflowTaskWorkerInternal(
 		taskWorker:        poller,
 		identity:          params.Identity,
 		workerType:        "DecisionWorker",
-		shutdownTimeout:   params.WorkerStopTimeout},
+		shutdownTimeout:   params.WorkerStopTimeout,
+		statsCollector:    params.StatsCollector},
```
taylanisikdemir, line 289: same here for curly bracket moving to new line

internal/internal_worker.go
```diff
@@ -482,7 +488,9 @@ func newActivityTaskWorker(
 			identity:          workerParams.Identity,
 			workerType:        workerType,
 			shutdownTimeout:   workerParams.WorkerStopTimeout,
-			userContextCancel: workerParams.UserContextCancel},
+			userContextCancel: workerParams.UserContextCancel,
+			statsCollector:    workerParams.StatsCollector},
```
taylanisikdemir, line 492: same here

internal/internal_worker_base.go
```diff
@@ -244,7 +245,10 @@ func (bw *baseWorker) isShutdown() bool {
 
 func (bw *baseWorker) runPoller() {
 	defer bw.shutdownWG.Done()
+	defer bw.options.statsCollector.ShutdownPoller()
+
 	bw.metricsScope.Counter(metrics.PollerStartCounter).Inc(1)
+	bw.options.statsCollector.StartPoller()
```
taylanisikdemir, line 251: do we need one statsCollector per poller? It doesn't know anything about poller (it seems a worker level component) so not sure what we expect from it per poller

ketsiambaku: It's per worker actually. I am calling these methods here to track the running pollers per worker but I plan to add more methods for collection on running activities and workflows on a worker. So it is on a worker level indeed

Groxx:
A whole bunch of stuff from IRL chat, but to put it in writing and try to make it a bit less meandering and aimless:
1. Let's name this _highly_ specific to what it's collecting, not "stats collector".  We'll want to make changes to it (adding more things at the least) and that'll require some boilerplate to deal with every change being a breaking change, so it's best to make it sound both unstable and as specific as it truly is.

---

I think we'll likely want an API like:
```go
WorkerOptions struct {
  UnstableEventMonitoringOrSomething struct {
    PollerLifecycle PollerLifecycle
  }
PollerLifecycle interface {
  Start(identity string)
  Stop(identity string)
}
```
so we can make additions and breaking changes like this:
```diff
 UnstableEventMonitoringOrSomething struct {
   PollerLifecycle   PollerLifecycle
+  PollerLifecycleV2 PollerLifecycleV2
 }
 PollerLifecycle interface {
   Start(identity string)
   Stop(identity string)
 }
+PollerLifecycleV2 interface {
+  PollerLifecycle
+  Poll(identity)
+}
```
and support old versions + new versions + more kinds of collectors in the future.

Another pattern that's fairly common in open source is to upcast a general interface:
```go

type StatsCollector interface { CollectsStats() }

```

internal/internal_worker.go
```diff
@@ -187,6 +187,10 @@ func ensureRequiredParams(params *workerExecutionParameters) {
 	if params.UserContext == nil {
 		params.UserContext = context.Background()
 	}
+	if params.EventMonitoring.PollerLifeCycle == nil {
+		params.EventMonitoring.PollerLifeCycle = newNoopPollerLifeCycle()
+		params.Logger.Info("No PollerLifeCycle configured for cadence worker. Use noop PollerLifeCycle as default.")
```
3vilhamster, line 192: nit: this should be a debug; nobody is forced to implement it.
However, we should have guidance on how we use it and how it could be helpful for OS customers.

internal/worker.go
```diff
@@ -271,6 +271,12 @@ type (
 		//
 		// Deprecated: All bugports are always deprecated and may be removed at any time.
 		WorkerBugPorts WorkerBugPorts
+
+		// Optional: EventMonitoring provides a set of methods that can be used to collect
+		// stats on the Worker for debugging purposes.
+		// default: noop implementation
+		// Deprecated: in development and very likely to change
```
3vilhamster, line 278: nit: Deprecated is not the right word, but go does not have Unstable/Experimental annotation. 
I guess this should be enough to show IDEs to not rely on this.

internal/worker.go
```diff
@@ -291,6 +297,25 @@ type (
 		// Deprecated: All bugports are always deprecated and may be removed at any time
 		DisableStrictNonDeterminismCheck bool
 	}
+
+	// PollerRun is a helper for simpler tracking of the go routine run
+	PollerRun interface {
```
3vilhamster, line 302: nit: move this into a separate file/package instead of one huge internal/worker. This file/package already have too many responsibilities

internal/internal_worker.go
```diff
@@ -1317,3 +1328,15 @@ func StartVersionMetrics(metricsScope tally.Scope) {
 		}()
 	})
 }
+
+// pollerLifeCycleImpl implements the PollerLifeCycle interface
+type pollerLifeCycleImpl struct{}
```
3vilhamster, line 1333: nit: I suggest moving this into a separate package and providing an example of how customers can use it or how you expect to use it.

internal/worker.go
ketsiambaku: moved

internal/worker.go
ketsiambaku: Yeah deprecated was the closest I found to a "do not use" annotation :)

internal/internal_worker.go
ketsiambaku: changed to debug.
I'm proposing a default implementation which just count the running pollers (which is basically what I was planning for the Uber internal implementation) and exposed a NewLifeCycle() constructor so as a start the OS users can at least have a wrapper around it

internal/internal_worker.go
```diff
@@ -39,6 +39,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"go.uber.org/cadence/internal/debug"
```
3vilhamster, line 42: nit: I think this should be in the same imports group as isolationgroup. But I guess we are missing gci integration. I think it might be fixed in a separate PR with CI enforcement.

3vilhamster (approved):

internal/debug/lifecycle_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPollerLifeCycle(t *testing.T) {
+	lifeCycle := NewLifeCycle()
+
+	// Initially, poller count should be 0
+	require.Equal(t, int32(0), lifeCycle.ReadPollerCount(), "Expected initial poller count to be 0")
```
jakobht, line 34:
```suggestion
	require.Equal(t, int32(0), lifeCycle.ReadPollerCount())
```

I do not think the message adds much, the default message will point to the line where it is clear what the issue is. Same for all the other asserts.

jakobht, line 34:
```suggestion
	assert.Equal(t, int32(0), lifeCycle.ReadPollerCount(), "Expected initial poller count to be 0")
```
Nit: I think since this is doing tests of the function i would just do all `assert`. Same for the other `require` in this function

internal/internal_worker.go
```diff
@@ -187,6 +189,10 @@ func ensureRequiredParams(params *workerExecutionParameters) {
 	if params.UserContext == nil {
 		params.UserContext = context.Background()
 	}
+	if params.EventMonitoring.LifeCycle == nil {
+		params.EventMonitoring.LifeCycle = debug.NewLifeCycle()
+		params.Logger.Debug("No LifeCycle configured for EventMonitoring option. Use default one.")
```
jakobht, line 194:
```suggestion
		params.Logger.Debug("No LifeCycle configured for EventMonitor. Will use the default.")
```
Nit

internal/debug/lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+	// lifeCycleImpl implements the LifeCycle interface
+	lifeCycleImpl struct {
+		pollerCount atomic.Int32
+	}
+
+	// runImpl implements the Run interface
+	runImpl struct {
+		lifeCycle *lifeCycleImpl
+		stopped   atomic.Bool
+	}
```
jakobht, line 35: I don't understand why we need this struct. The lifecycle holds the whole state - and where it is call there is no risk of calling stop twice. 

So we could call .Stop() on the LifeCycle itself.

internal/debug/lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+// NewLifeCycle creates a new LifeCycle instance
+func NewLifeCycle() LifeCycle { return &lifeCycleImpl{} }
+
+var _ LifeCycle = &lifeCycleImpl{}
+
+func (r *runImpl) Stop() {
+	// Check if Stop() has already been called
+	if r.stopped.CompareAndSwap(false, true) {
+		r.lifeCycle.pollerCount.Dec()
+	} else {
```
jakobht, line 83: I don't understand why we are tracking if it's stopped we call this function exactly once, when the poller stops, so I do not think this is nessesary.

internal/debug/lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// EventMonitoring provides a set of methods that can be used to collect
+	// stats on the Worker for debugging purposes.
+	// Deprecated: in development and very likely to change
+	EventMonitoring struct {
+		LifeCycle LifeCycle
+	}
+)
+
+func (lc *lifeCycleImpl) PollerStart(workerID string) Run {
```
jakobht, line 62:
```suggestion
func (lc *lifeCycleImpl) PollerStart() Run {
```
We are not using the argument here? Maybe we should have a map or something? But currently it's not used.

jakobht (commented):
Added some comments, I'm not sure I completely understand what we want to achieve, so some of the comments might reflect that.

internal/debug/lifecycle.go
ketsiambaku: The initial plan was to define the interface only and the implementation would need to be provided by users. Internally, we’d have only a noop implementation for backward compatibility. Having only a Start() and Stop() with no args would limit the users on their implementation. That’s why I added at least workerID as an argument. 
Eventually I added a default implementation and though it is not used in the default workerID just seemed like a good starting point for consumers implementation

internal/common/debug/lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+	// Run is a helper for simpler tracking of the go routine run
+	// Deprecated: in development and very likely to change
+	Run interface {
+		// Stop is the method to report stats once a poller thread is done
+		Stop()
+	}
+
+	// LifeCycle contains a set of methods to collect information on a running worker
+	// Deprecated: in development and very likely to change
+	LifeCycle interface {
```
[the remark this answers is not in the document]
ketsiambaku: made the changes per yesterday’s discussion. Per @jakobht 's point Now that we removed the ID argument from Start() we don’t have to return a Stopper and can instead have Stop() directly in the PollerTracker. I don’t have a strong opinion or preference on this.

PR #1358, opened 2024-07-16
Override idls module to submodule to simplify idl update flow

3vilhamster:
**What changed?**
Switched go mod for cadence-idls to submodule so we need to only update submodule when we want to updates idls.

**Why?**
Eliminate a confusion of separate version of idls used for thrift and proto and separate version used.

**How did you test it?**
make build/make test

**Potential risks**

### DIFF
--- a/docker/buildkite/Dockerfile
+++ b/docker/buildkite/Dockerfile
@@ -4,10 +4,10 @@ RUN mkdir -p /go/src/go.uber.org/cadence
 WORKDIR /go/src/go.uber.org/cadence
 
 ADD go.mod go.sum /go/src/go.uber.org/cadence/
+ADD idls /go/src/go.uber.org/cadence/idls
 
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -114,3 +114,5 @@ retract (
 	v0.3.0 // Published accidentally
 	v0.2.0 // Published accidentally
 )
+
+replace github.com/uber/cadence-idl => ./idls
```
Groxx, line 118: replaces do not work for anyone importing the client, so this only works for local developers

Groxx (changes requested):
while I like the idea and there might be something we can do to streamline this, I'm pretty confident this can't work the way we want it to.  it'll just cause even-stranger issues when we make changes and fail to update one or the other, and they're less likely to be caught in CI since they'll only affect external users (due to replace).

PR #1359, opened 2024-07-25
Added minor adjustments in the shadow status options

agautam478:
**What changed?**
Added minor adjustments in the shadow status options so that Terminated workflows are never replayed.
Removed the terminated workflow status from the All option as well. 
Added default option for replaying only closed and open workflows and no other workflows if workflow status filter is not passed.

**Why?**
This is to ensure that the terminated workflows are not replayed. We also have bugs associated with continue as new workflows that need to be fixed before those wfs can be replayed.

**How did you test it?**
test cases.

**Potential risks**
shadower accuracy may get affected.

### DIFF
--- a/internal/query_builder.go
+++ b/internal/query_builder.go
@@ -50,8 +50,6 @@ var (
 	WorkflowStatusFailed = WorkflowStatus(shared.WorkflowExecutionCloseStatusFailed.String())
 	// WorkflowStatusCanceled is the WorkflowStatus for canceled workflows
 	WorkflowStatusCanceled = WorkflowStatus(shared.WorkflowExecutionCloseStatusCanceled.String())
-	// WorkflowStatusTerminated is the WorkflowStatus for terminated workflows
-	WorkflowStatusTerminated = WorkflowStatus(shared.WorkflowExecutionCloseStatusTerminated.String())
 	// WorkflowStatusContinuedAsNew is the WorkflowStatus for continuedAsNew workflows
 	WorkflowStatusContinuedAsNew = WorkflowStatus(shared.WorkflowExecutionCloseStatusContinuedAsNew.String())
 	// WorkflowStatusTimedOut is the WorkflowStatus for timedout workflows
@@ -183,7 +181,7 @@ func ToWorkflowStatus(statusString string) (WorkflowStatus, error) {
 	status := WorkflowStatus(strings.ToUpper(statusString))
 	switch status {
 	case WorkflowStatusOpen, WorkflowStatusClosed, WorkflowStatusCompleted,
-		WorkflowStatusFailed, WorkflowStatusCanceled, WorkflowStatusTerminated,
+		WorkflowStatusFailed, WorkflowStatusCanceled,
 		WorkflowStatusContinuedAsNew, WorkflowStatusTimedOut, WorkflowStatusALL:
 		return status, nil
 	default:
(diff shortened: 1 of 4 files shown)


internal/query_builder.go
```diff
@@ -183,7 +181,7 @@ func ToWorkflowStatus(statusString string) (WorkflowStatus, error) {
 	status := WorkflowStatus(strings.ToUpper(statusString))
 	switch status {
 	case WorkflowStatusOpen, WorkflowStatusClosed, WorkflowStatusCompleted,
-		WorkflowStatusFailed, WorkflowStatusCanceled, WorkflowStatusTerminated,
+		WorkflowStatusFailed, WorkflowStatusCanceled,
```
taylanisikdemir, line 184: you can still leave it here in the mapping function.

internal/workflow_shadower.go
```diff
@@ -338,10 +338,9 @@ func (o *ShadowOptions) validateAndPopulateFields() error {
 			}
 			statuses = append(statuses, status)
 		}
-		//All the open statuses are taken by default. This list seems to not work as expected.
-		//TODO: verify that the status list works as expected. currently all wfs of all types get picked up.
+		//This filter doesn't seem to be working as expected.
```
taylanisikdemir, line 341: please elaborate in the comment what you mean

internal/workflow_shadower_test.go
```diff
@@ -203,7 +203,7 @@ func (s *workflowShadowerSuite) TestShadowOptionsValidation() {
 			options:   ShadowOptions{},
 			expectErr: false,
 			validationFn: func(options *ShadowOptions) {
-				s.Equal("(CloseTime = missing)", options.WorkflowQuery)
+				s.Equal("(CloseTime = missing or CloseTime != missing)", options.WorkflowQuery)
```
taylanisikdemir, line 206: is not this equivalent to empty query?

agautam478: it is but the query builder forms it this way because of open and closed statuses getting added together in the test. This portion of default status check needs a rewrite that I will be doing as a part of the shadower fix initiative.

internal/workflow_shadower.go
agautam478: This filter should have only taken open and closed wfs but it returns continue as new wfs as well. not sure if this is because of some other check somewhere or the query gets formed incorrectly.

internal/query_builder.go
agautam478: I will readd it later. removed it to be sure that the status can not be used in any situation at all.

internal/workflow_shadower_test.go
agautam478: we do have an ALL status filter somewhere but it barely gets used. Now the problem with that one is that it picks up everything present in that defaut list of statuses.

internal/workflow_shadower.go
```diff
@@ -338,10 +338,9 @@ func (o *ShadowOptions) validateAndPopulateFields() error {
...
 		if len(statuses) == 0 {
-			statuses = []WorkflowStatus{WorkflowStatusOpen}
+			statuses = []WorkflowStatus{WorkflowStatusOpen, WorkflowStatusClosed}
 		}
 		queryBuilder.WorkflowStatus(statuses)
```
3vilhamster, line 345: I guess modern DBs with good query analyzers will just drop this expression, but evaluation of this does not make any sense.
I suggest making this optional:
```
if len(statuses) != 0 {
  queryBuilder.WorkflowStatus(statuses)
}
```

agautam478: will incorporate this and update PR.

abhishekj720 (approved):
LGTM. 
For future me/others, during the shadower fix initiative, default status check would be rewritten.

PR #1360, opened 2024-07-31
Add FirstRunAt to StartWorkflowOptions

timl3136:
**What changed?**
Add FirstRunAt in StartWorkflowOptions. Server logic PR: https://github.com/uber/cadence/pull/6178

**Why?**
Add the feature allowing client to specify a exact time for the first run of their workflow.

**How did you test it?**
Unit tests, also tested in cadence-samples repo as well as dev environment

**Potential risks**
It should not break anything.

**Detailed Description**
Add a new start workflow option called First_Run_At, which would affect the scheduled time for first run of a workflow.

**Impact Analysis**
- It’s an additional optional field so it should be compatible with all clients as the processing happens in the server
- In terms of server backward compatibility, old server should be compatible with new requests as the field is added to the end of the request and will be dropped. The logic when this field is ignored will be the same as the older version.

- It’s forward compatible because new server will treat the value as unset when the old request that's missing that field is received.

**Testing Plan**
- Yes, it would be implemented in both server and client repo
- The data will not be persisted

**Rollout Plan**
- What is the rollout plan?
1. Add and merge the fields in cadence-idl repo
2. Update SHA for cadence-idl repo and handled data fields in shared.go in cadence-server repo
3. Rollout to all clusters before updating SHA on cadence-client
4. Updating SHA for cadence-idl and add the fields in cadence-client
6. Update client version in cadence-canary for monitoring
7. Add canary workflows to ensure the new field work as expected
8. Release new version in cadence-client with the added field

Yes, we need to get the server build updated before deploying such change in client.
Yes, rollback server version is safe given that the change is backward compatible.
No


internal/internal_workflow_client.go
```diff
@@ -1048,6 +1048,14 @@ func (wc *workflowClient) getWorkflowStartRequest(
 		return nil, errors.New("Invalid JitterStart option")
 	}
 
+	firstRunAtTimestamp := options.FirstRunAt.UnixNano()
+	if options.FirstRunAt.IsZero() {
+		firstRunAtTimestamp = 0
+	}
+	if firstRunAtTimestamp < 0 {
+		return nil, errors.New("Invalid FirstRunAt option")
```
taylanisikdemir, line 1056: please cover these negative cases in unit tests

taylanisikdemir (approved):

internal/internal_workflow_client.go
timl3136: Done

PR #1361, opened 2024-08-16
Remove deprecated fossa

dkrotx:
Fossa's being removed / decomm'd. Removing for now until we move to a replacement

Fossa's being removed / decomm'd. Removing for now until we move to a replacement

We're blocked because of this in builds.
It seems to emit wrong warnings.


dkrotx:
Same as for Cadence repo here: https://github.com/uber/cadence/pull/6212

ketsiambaku (approved):

PR #1362, opened 2024-08-26
Add ActivityTracker to worker stats option

ketsiambaku:
**What changed?**
This changes adds the ActivityTracker in the worker options. In this PR, to count and get information on running activities in a worker. 

The activityTracker field is added to activityTaskHandler and localActivityTaskHandler and emitted upon calling Execute() ExecuteWithActualArgs() for local activity.

**Why?**
Want to retrieve running activities on a worker instance

**How did you test it?**
unit tests
local test

**Potential risks**
When deploying the noop implementation is used by default
potential risks include nil pointer dereference panics in the event that we missed initializing the activityTracker in a particular flow

### DIFF
--- a/internal/common/debug/example_test.go
+++ b/internal/common/debug/example_test.go
@@ -21,10 +21,10 @@
 package debug
 
 import (
+	"encoding/json"
 	"fmt"
-	"sync"
-
 	"go.uber.org/atomic"
+	"sync"
 )
 
 type (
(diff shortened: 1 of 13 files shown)


internal/common/debug/example_test.go
```diff
@@ -36,8 +37,54 @@ type (
...
+var _ Stopper = &activityStopperImpl{}
+
+func (ati *activityTrackerImpl) Start(info ActivityInfo) Stopper {
+	ati.Lock()
+	ati.m[info]++
+	ati.Unlock()
+	return &activityStopperImpl{info: info, tracker: ati}
+}
+
+func (ati *activityTrackerImpl) Stats() Activities {
```
ketsiambaku, line 65: We can group Activities by activityType or other fields of ActivityInfo if needed

internal/common/debug/types.go
```diff
@@ -43,6 +43,32 @@ type (
...
 	WorkerStats struct {
-		PollerTracker PollerTracker
+		PollerTracker   PollerTracker
+		ActivityTracker ActivityTracker
+	}
+
+	// ActivityInfo contains details on the executing activity
+	// Deprecated: in development and very likely to change
+	ActivityInfo struct {
+		WorkflowID   string
```
3vilhamster, line 53: We cannot have workflowIDs/RunIDs as a granularity because this will explode in memory.
Maps only grow and will never be downscaled. I remember that was my proposal, but I was wrong here :(

internal/common/debug/example_test.go
```diff
@@ -36,8 +38,54 @@ type (
...
+	}
 )
 
+var _ ActivityTracker = &activityTrackerImpl{}
+var _ Stopper = &activityStopperImpl{}
+
+func (ati *activityTrackerImpl) Start(info ActivityInfo) Stopper {
+	ati.Lock()
+	ati.m[info]++
+	ati.Unlock()
```
jakobht, line 62: Nit: can we defer this instead? I dont think line 61 will fail, but just in case

internal/common/debug/example_test.go
```diff
@@ -36,8 +38,54 @@ type (
...
+		}
+	}
+	return activities
+}
+
+func (asi *activityStopperImpl) Stop() {
+	asi.Do(func() {
+		asi.tracker.Lock()
+		defer asi.tracker.Unlock()
+		asi.tracker.m[asi.info]--
```
3vilhamster, line 85: This should not only do a decrement, we need to clean the map.

internal/common/debug/example_test.go
```diff
@@ -36,8 +38,54 @@ type (
 	stopperImpl struct {
 		pollerTracker *pollerTrackerImpl
 	}
+
+	// activityTrackerImpl implements the ActivityTracker interface
+	activityTrackerImpl struct {
+		sync.RWMutex
+		m map[ActivityInfo]int64
```
jakobht, line 45: Nit: could we call this map `activityCounts` or something like that?

jakobht (commented):
A few nits, but looks good

internal/common/debug/types.go
ketsiambaku: 🥲

jakobht (approved):

3vilhamster (approved):

PR #1363, opened 2024-09-02
Add debugger interface to allow extracting the workerStats from Worker

ketsiambaku:
**What changed?**
This PR adds a debugger interface to the debug package

**Why?**
Currently the WorkerStats struct is injected in the client by passing it as a worker options however there is no way of retrieving the WorkerStats or its fields from a Worker instance unless we keep track of Trackers instances that were constructed and passed as options seperately. In this PR, we make improvement to the design by adding a debugger interface. The internal worker implementation implements Debugger and allow consumers to retrieve the WorkerStats

**How did you test it?**
Unit tests
Local Test

**Potential risks**

### DIFF
--- a/debug/types.go
+++ b/debug/types.go
@@ -51,4 +51,8 @@ type (
 	// Activities is a list of executing activities on the worker
 	// Deprecated: in development and very likely to change
 	Activities = internal.Activities
+
+	// Debugger exposes stats collected on a running Worker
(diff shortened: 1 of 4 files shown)


dkrotx (commented):
Good approach!

dkrotx (approved):

PR #1364, opened 2024-09-09
Improve example test in debug package

ketsiambaku:
**What changed?**

Improving the example test by sorting the activities returned by Stats()
The example implementation range over a map and return the result in a slice. Because of this, the assertion against expected result would not always pass since the order is random. This made the test flaky 

**Why?**

**How did you test it?**

**Potential risks**


ketsiambaku:
fixed in #1366

PR #1366, opened 2024-09-10
improve example test by sorting output

ketsiambaku:
**What changed?**
Improving the example test by sorting the activities returned by Stats()
The example implementation range over a map and return the result in a slice. Because of this, the assertion against expected result would not always pass since the order is random. This made the test flaky

**Why?**

**How did you test it?**
unit tests

**Potential risks**

### DIFF
--- a/internal/common/debug/example_test.go
+++ b/internal/common/debug/example_test.go
@@ -78,7 +78,7 @@ func (ati *activityTrackerImpl) Stats() Activities {
 		}
 	}
 	sort.Slice(activities, func(i, j int) bool {
-		return strings.Compare(strings.ToLower(activities[i].Info.ActivityType), strings.ToLower(activities[j].Info.ActivityType)) < 0
+		return strings.Compare(activities[i].Info.ActivityType, activities[j].Info.ActivityType) < 0
 	})
 	return activities
 }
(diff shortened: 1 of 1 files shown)


internal/common/debug/example_test.go
```diff
@@ -75,6 +77,9 @@ func (ati *activityTrackerImpl) Stats() Activities {
 			}{Info: a, Count: count})
 		}
 	}
+	sort.Slice(activities, func(i, j int) bool {
+		return strings.Compare(strings.ToLower(activities[i].Info.ActivityType), strings.ToLower(activities[j].Info.ActivityType)) < 0
```
Groxx, line 81: Sorting makes perfect sense, but there's no reason to `ToLower` here afaict.  It'll just make things more ambiguous than they are IRL, where strings need to be identical in both workers and example-test output checks.

ketsiambaku: changed

Groxx (approved):
From the docs on `strings.Compare` it looks like we can/probably should just use `<` here... but they appear to be identical so I'll stamp now and land whichever you prefer :)  thankfully it's not a locale-sensitive comparison at least.

ketsiambaku:
> From the docs on `strings.Compare` it looks like we can/probably should just use `<` here... but they appear to be semantically identical so I'll stamp now and land whichever you prefer :) thankfully it's not a locale-sensitive comparison at least.

I saw the last comment after merging. Followed up in #1367

Issue #1369, opened 2024-10-08
Workers with single decision poller with sticky tasklist feature enabled will stop processing


shijiesheng:
**Describe the bug**
Decision worker will stop processing if decision poller count is 1 and sticky tasklist is enabled

**To Reproduce**
MaxConcurrentDecisionTaskPollers is set to 1 (default 2)
Or MinConcurrentDecisionTaskPollers is set to 1 and Poller Auto Scaler is enabled

**Expected behavior**
Workflow workers should process decision tasks

shijiesheng:
**Root Cause**
Current implementation is:
X number of decision goroutines are started
Each goroutine will either call sticky tasklist or normal tasklist based on metrics
When X = 1, the [condition](https://github.com/uber-go/cadence-client/blob/3d320e77009f08330bdc3146bb2346ab57f9464b/internal/internal_task_pollers.go#L748) will always be true. This causes decision polling to stop for normal tasklist

Temporal keeps the same logic https://github.com/temporalio/sdk-go/blob/cdd30702f1ba479b57111763095e24a25a04ea53/internal/internal_task_pollers.go#L781

shijiesheng:
We decided to add enforcement to have at least 2 decision pollers to avoid breaking existing behaviors

PR #1370, opened 2024-10-08
added worker.NewV2 with validation on decision poller count

shijiesheng:
**What changed?**

- added a `worker.NewV2` interface that will return error and a deprecation message on `New` api.
- added validation on WorkerOptions
-- decision poller should be larger than 1 if value is set (not zero) and sticky execution is enabled. 

**Why?**
https://github.com/uber-go/cadence-client/issues/1369

**How did you test it?**
Unit Test

**Potential risks**
Low

===== below is required for breaking change ======

**Detailed Description**
added a new `worker.NewV2` interface that will return error and a deprecation message on `New` api.

**Impact Analysis**
- **Backward Compatibility**:  yes
- **Forward Compatibility**:  yes

**Testing Plan**
- **Unit Tests**: yes
- **Persistence Tests**: not related
- **Integration Tests**: not related
- **Compatibility Tests**: tested in Uber internal canary test

**Rollout Plan**
- What is the rollout plan? version upgrade and change usage of `New` to `NewV2`
- Does the order of deployment matter? no
- Is it safe to rollback? Does the order of rollback matter? yes

### DIFF
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -335,10 +335,7 @@ func NewWorker(
 	taskList string,
 	options WorkerOptions,
 ) (*aggregatedWorker, error) {
-	if err := options.Validate(); err != nil {
-		return nil, fmt.Errorf("worker options validation error: %w", err)
-	}
-	return newAggregatedWorker(service, domain, taskList, options), nil
+	return newAggregatedWorker(service, domain, taskList, options)
 }
 
 // ReplayWorkflowExecution loads a workflow execution history from the Cadence service and executes a single decision task for it.
--- a/internal/worker.go
+++ b/internal/worker.go
@@ -23,6 +23,7 @@ package internal
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"go.uber.org/cadence/internal/common/debug"
@@ -108,7 +109,7 @@ type (
 		// optional: Sets the minimum number of goroutines that will concurrently poll the
 		// cadence-server to retrieve decision tasks. If FeatureFlags.PollerAutoScalerEnabled is set to true,
 		// changing this value will NOT affect the rate at which the worker is able to consume tasks from a task list.
-		// Default value is 1
+		// Default value is 2
 		MinConcurrentDecisionTaskPollers int
 
 		// optional: Sets the interval of poller autoscaling, between which poller autoscaler changes the poller count
@@ -333,8 +334,11 @@ func NewWorker(
 	domain string,
 	taskList string,
 	options WorkerOptions,
-) *aggregatedWorker {
-	return newAggregatedWorker(service, domain, taskList, options)
+) (*aggregatedWorker, error) {
+	if err := options.Validate(); err != nil {
+		return nil, fmt.Errorf("worker options validation error: %w", err)
+	}
+	return newAggregatedWorker(service, domain, taskList, options), nil
 }
 
 // ReplayWorkflowExecution loads a workflow execution history from the Cadence service and executes a single decision task for it.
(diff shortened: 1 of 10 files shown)


abhishekj720 (approved):

internal/worker.go
```diff
@@ -333,8 +334,11 @@ func NewWorker(
 	domain string,
 	taskList string,
 	options WorkerOptions,
-) *aggregatedWorker {
-	return newAggregatedWorker(service, domain, taskList, options)
+) (*aggregatedWorker, error) {
+	if err := options.Validate(); err != nil {
+		return nil, fmt.Errorf("worker options validation error: %w", err)
+	}
+	return newAggregatedWorker(service, domain, taskList, options), nil
```
Groxx, line 341: might be better to put this inside newAggregatedWorker tbh, since it modifies the options after this.

validating before that makes plenty of sense, but it does feel a bit odd to *not* validate after it, in case we messed up the `AugmentWorkerOptions` somehow.

worker/worker.go
```diff
@@ -283,7 +283,11 @@ func New(
 	taskList string,
 	options Options,
 ) Worker {
-	return internal.NewWorker(service, domain, taskList, options)
+	w, err := internal.NewWorker(service, domain, taskList, options)
+	if err != nil {
+		panic(err)
```
Groxx, line 288: hm :\

might be worth deferring error handling until it's started?  or does that not return an error.

I think this won't make much of a difference internally, since everything's in fx and that'll just translate a panic to an error, but this is definitely one of those "it's a breaking change even though it doesn't break code" things...

shijiesheng: make sense. I'll go with the start approach first.

Groxx: honestly I'm pretty surprised it didn't return an error earlier.  constructors like this should generally be reserving that for future changes like this...

Groxx: ah, to double check: does all-caps `// DEPRECATED: ...` mark it visibly-deprecated in your IDE?  I think they're gradually getting more flexible about that, but Go uses `// Deprecated: ...` in the standard library and I'm not sure how picky godocs / etc are: https://pkg.go.dev/path/filepath#HasPrefix

Groxx: yea... I think it's case-sensitive: 
Screenshot 2024-10-09 at 18 50 48

worth changing IMO.  and also probably the indentation on the v2 docs, since I don't think "Returns an error..." is intended to be part of the code block.

worker/worker.go
```diff
@@ -277,12 +277,36 @@ const (
...
+
+// NewV2 returns an instance of worker for managing workflow and activity executions and an error.
+//
+//		service  - thrift connection to the cadence server
+//		domain   - the name of the cadence domain
+//		taskList - is the task list name you use to identify your client worker, also
+//		           identifies group of workflow and activity implementations that are
+//		           hosted by a single worker process
+//		options  - configure any worker specific options like logger, metrics, identity
+//	 Returns an error if the worker cannot be created.
```
taylanisikdemir, line 303: let's fix comment indentation

worker/worker.go
```diff
@@ -277,12 +277,36 @@ const (
 //	           identifies group of workflow and activity implementations that are
 //	           hosted by a single worker process
 //	options  - configure any worker specific options like logger, metrics, identity
+//
+// DEPRCATED: use NewV2 instead since this implementation will panic on error
```
taylanisikdemir, line 281: typo and also change casing to `// Deprecated` as @Groxx mentioned in other comment

worker/worker.go
```diff
@@ -277,12 +277,36 @@ const (
...
 func New(
```
taylanisikdemir, line 282: please update CHANGELOG.md about this new validation and the new `worker.NewV2` constructor

worker/worker.go
shijiesheng: fixed

worker/worker.go
shijiesheng: sure

worker/worker.go
shijiesheng: Changed but local godoc still doesn't show the deprecation warning message. I don't know how to test it.

worker/worker.go
shijiesheng: fixed

worker/worker.go
shijiesheng: fixed

worker/worker.go
shijiesheng: Screenshot 2024-10-10 at 9 58 34 AM
working now after using the suggested tool `https://github.com/golang/pkgsite`

Groxx: yea I have no idea why `godoc` / `go doc` has been seemingly abandoned in the docs changes they've been making :\

Groxx (approved):
thanks!

PR #1373, opened 2024-10-30
[internal] Improve code coverage of internal_task_pollers.go

3vilhamster:
**What changed?**
Refactored task pollers and improve code coverage.

**Why?**
Improvign code coverage.

**How did you test it?**
unit tests

**Potential risks**


Makefile
```diff
@@ -219,7 +219,6 @@ $(THRIFT_GEN): $(THRIFT_FILES) $(BIN)/thriftrw $(BIN)/thriftrw-plugin-yarpc
 # this needs to be both the files defining the generate command, AND the files that define the interfaces.
 $(BUILD)/generate: client/client.go encoded/encoded.go internal/internal_workflow_client.go $(BIN)/mockery
 	$Q $(BIN_PATH) go generate ./...
-	$Q touch $@
```
Groxx, line 222: if you get rid of this, it'll re-generate every single time.
just add the new-`//go:generate` file to the list of dependencies.

codecov.yml
```diff
@@ -31,6 +31,7 @@ codecov:
 ignore:
   - "**/*_generated.go"
   - "**/*_mock.go"
```
Groxx, line 33:
```
❯ find . -type file -name '*_mock.go'

```
huh.  yea, prefix seems fine - helps correlate them all.

internal/internal_public.go
```diff
@@ -33,6 +33,8 @@ import (
 	s "go.uber.org/cadence/.gen/go/shared"
 )
 
+//go:generate mockery --srcpkg . --name WorkflowTaskHandler --output . --outpkg internal --inpackage --with-expecter --case snake --boilerplate-file  ../LICENSE
```
Groxx, line 36:
```suggestion
//go:generate mockery --name WorkflowTaskHandler --inpackage --with-expecter --case snake --boilerplate-file  ../LICENSE
```
quite a bit simpler and produces identical output (since those three flags are exactly what `--inpackage` does)

internal/internal_task_pollers.go
```diff
@@ -45,6 +45,8 @@ import (
 	"go.uber.org/cadence/internal/common/serializer"
 )
 
+//go:generate mockery --srcpkg . --name LocalDispatcher --output . --outpkg internal --inpackage --with-expecter --case snake --boilerplate-file  ../LICENSE
```
Groxx, line 48:
```suggestion
//go:generate mockery --name LocalDispatcher --inpackage --with-expecter --case snake --boilerplate-file ../LICENSE
```

internal/internal_task_pollers.go
```diff
@@ -159,6 +161,11 @@ type (
 		stopCh       <-chan struct{}
 		metricsScope *metrics.TaggedScope
 	}
+
+	// LocalDispatcher is an interface to dispatch locally dispatched activities.
+	LocalDispatcher interface {
```
Groxx, line 166: no reason for this to be public (AFAICT)

Makefile
3vilhamster: It is very annoying that is materializes a file if you need to update/add a new generated file.
What is the benefit of it?

codecov.yml
3vilhamster: mockery does not allow the setting of a file name and adds mock as a prefix when it is inpackage (for some unknown reason)

internal/internal_task_pollers.go
3vilhamster: Good point. Updated

Makefile
3vilhamster: make generate is noisy (because mockery is noisy). But it runs super fast, so I'd rather remove this make magic.

Groxx: it's normal makefile stuff, and we make changes to these mocks so _incredibly_ infrequently that it has probably already paid for itself several times over in CI and dev time.

but yea - it can be made automatic, and probably fast enough to not notice.

codecov.yml
Groxx: `--filename` sets the filename.

it is kinda odd that it isn't used in the `mocks/` folder though.

Makefile
3vilhamster: On CI, it is always running from scratch, and for local, if you need to run generate, you will have to remove the file.

codecov.yml
3vilhamster: I missed that. I'll switch to _mock suffix for a unified approach.

Makefile
Groxx: because the list of dependencies hasn't been updated:
> # this needs to be both the files defining the generate command, AND the files that define the interfaces.

do that and it'll generate exactly when necessary.

I can make a PR to infer these files too.

internal/internal_task_pollers.go
```diff
@@ -365,10 +372,10 @@ func (wtp *workflowTaskPoller) processWorkflowTask(task *workflowTask) error {
 		if completedRequest == nil && err == nil {
 			return nil
 		}
-		if _, ok := err.(decisionHeartbeatError); ok {
+		if errors.As(err, new(*decisionHeartbeatError)) {
```
Groxx, line 375: yea, looks good - it's only returned from one place right now, and no wrapping, so this won't change behavior 👍

internal/internal_task_pollers.go
```diff
@@ -397,8 +404,7 @@ func (wtp *workflowTaskPoller) processResetStickinessTask(rst *resetStickinessTa
 	return nil
 }
 
-func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(completedRequest interface{}, taskErr error, task *s.PollForDecisionTaskResponse, startTime time.Time) (response *s.RespondDecisionTaskCompletedResponse, err error) {
-
+func (wtp *workflowTaskPoller) RespondTaskCompleted(completedRequest interface{}, taskErr error, task *s.PollForDecisionTaskResponse, startTime time.Time) (response *s.RespondDecisionTaskCompletedResponse, err error) {
```
Groxx, line 407: not entirely sure how I feel about kinda-shadowing an internal helper method with a "public" method like this 🤔 
- it's rather clear they're related, and public is "preferred", which is nice
- there are a lot of case-insensitive searches all over, and people seem not all that likely to notice `r` vs `R`, so this seems like it might lead to a bit more false-matches / easy to miss during code review / etc

dunno.  do we have any other examples in this repo?  I can't think of any at the moment at least.  it might be a bit nicer than using `respondTaskCompletedInternal` or something if we like it in general.

internal/internal_task_pollers.go
```diff
@@ -425,103 +431,114 @@ func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(completedRequest
...
+		err = wtp.service.RespondQueryTaskCompleted(ctx, request, getYarpcCallOptions(wtp.featureFlags)...)
+		operation = "RespondQueryTaskCompleted"
+	default:
+		// should not happen
+		panic("unknown request type from ProcessWorkflowTask()")
+	}
+
+	traceLog(func() {
+		wtp.logger.Debug("Call failed.", zap.Error(err), zap.String("Operation", operation))
+	})
```
Groxx, line 471: probably want an `if err != nil` case around this, as it's not only being called when failing.

also it think I'll vote for a more descriptive text than "call failed", as that message is the main key for filtering logs / searching logs.  "respond task completed failed" maybe?  "operation" as a key seems probably fine (lower case is generally easier for tooling tho), it's just that "call failed" is about as vague as it gets.

internal/internal_task_pollers.go
```diff
@@ -425,103 +431,114 @@ func (wtp *workflowTaskPoller) RespondTaskCompletedWithMetrics(completedRequest
...
-					}
+			response, err = wtp.respondTaskCompletedAttempt(completedRequest, task)
+			return err
+		}, createDynamicServiceRetryPolicy(ctx), isServiceTransientError)
+
+	return response, err
+}
+
+func (wtp *workflowTaskPoller) respondTaskCompletedAttempt(completedRequest interface{}, task *s.PollForDecisionTaskResponse) (*s.RespondDecisionTaskCompletedResponse, error) {
+	ctx, cancel, _ := newChannelContext(context.Background(), wtp.featureFlags)
```
Groxx, line 447: tbh I think I'd just save `opts` here and pass it through below, since it's just being thrown away and then duplicated 3x now.  it'll probably be more flexible for the future too, if we ever pass more args through, because it'll be 3 fewer places where they might get dropped.

internal/internal_task_pollers_test.go
```diff
@@ -54,3 +68,132 @@ func TestLocalActivityPanic(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Nil(t, res)
+	})
+}
+
+func TestRespondTaskCompleted_Unsupported(t *testing.T) {
+	poller, _, _, _ := buildWorkflowTaskPoller(t)
+
+	assert.Panics(t, func() {
+		_, _ = poller.RespondTaskCompleted(assert.AnError, nil, &s.PollForDecisionTaskResponse{}, time.Now())
```
Groxx, line 126: given how many things this could be, probably worth adding a
```
PanicsWithValue(t, "unknown request type from ProcessWorkflowTask()", func() {...})
```

internal/internal_task_pollers_test.go
```diff
@@ -54,3 +68,132 @@ func TestLocalActivityPanic(t *testing.T) {
...
+		ch := make(chan struct{})
+		poller.shutdownC = ch
+		close(ch)
+
+		err := poller.ProcessTask(&workflowTask{})
+		assert.ErrorIs(t, err, errShutdown)
+	})
+	t.Run("unsupported task type", func(t *testing.T) {
+		poller, _, _, _ := buildWorkflowTaskPoller(t)
+		assert.Panics(t, func() {
```
Groxx, line 142: same `PanicsWithValue` thing.

if it's too hard to get equality / feels too over-precise:
```go
assert.panics(t, func() {
  defer func() {
    assert.Contains(t, recover(), "something") // sadly args are reversed
  }()
  thing()
})
```
takes care of it too

Groxx (approved):
Only minor stuff to change I think:
- just add the dependencies so it keeps working
- `opts` maybe?  I kinda see why you did it, just not sure it's worth it
- tracelog needs an err check at least, probably minor changes
- minor test precision since "panics" is extremely broad about what it's checking for

otherwise LGTM, breaking that up seems like a solid improvement 👍

internal/internal_task_pollers.go
3vilhamster: This was not working before at all.
The error was returned as &decicisionHeartbeatError so type assertion err.(decisionHeartbeatError) was always false.

internal/internal_task_pollers.go
3vilhamster: Opts are entirely independent from the context, and passing them around will recreate a slice 3+ times.

I'd vote for removing feature flags from this builder completely.

internal/internal_task_pollers.go
3vilhamster: Good catch! I've missed the if err != nil.
I'm confused about these logs since I have never heard anyone use them. I think a tag operation will indicate which operation failed, but I can switch to sprint to preserve behavior.

internal/internal_task_pollers.go
3vilhamster: I returned them. I kind of understand that it is possible for us to have feature flags that might go into the context. However, we don't, and this functionality looks completely orthogonal to the context right now.

internal/internal_task_pollers.go
3vilhamster: Renamed it back to RespondTaskCompletedWithMetrics.

Makefile
3vilhamster: I mean, I don't understand `already paid for itself several times over in CI and dev time.`. 
CI runs generate every single time, and devs are not running make generate if they do not need to generate anything.

The default behavior in the mono repo or backend is to always regenerate everything, and it has many more files to generate. I don't see a big issue with regenerating here since we have a few files, and if we switch from mockery to mockgen - it will work silently.

3vilhamster: Anyway, I'll leave this to a separate discussion: why do we even need to optimize this?
Reverted back to touch

internal/internal_task_pollers.go
Groxx: ah, good point.  and good fix 👍

Makefile
Groxx: CI skips this step for everything except the validation job, because it doesn't need to be done.  That job pretty much just exists to make sure stuff like this change doesn't happen.

And in dev time, correct, devs do not run `make generate`.  Because they _don't need to_ - the dependencies are wired up correctly so it generates when it might need to, and does nothing when it doesn't.

---

I think half of this is that you're looking at `make generate` in isolation instead of everything else that depends on it.  
By getting rid of the `touch $@`, this means _nothing_ that depends on code generation can be pruned by `make` - currently that includes lint/fmt/build, though not `make test` (mostly because I haven't fixed that half of the makefile yet, it's still quite janky from before and it's a lot harder because our CI pipeline depends on it in a lot of weakly-connected ways).

The server repo doesn't do this because it's so much slower and the generation dependencies are so much more complex.  There it's easier to force people to run `make go-generate` by hand as needed.  There isn't any need to do that in this repo.

internal/internal_task_pollers.go
Groxx: yea, at the moment I think it's separate enough that it's kinda strange that we pass them in like this at all.  so I think it's fine either way.

maybe yarpc has an "add these opts to every call" option somewhere nowadays.  we could do it with a middleware, but that's a pain too.

PR #1377, opened 2024-11-01
Fix go-generate calling, do more before running tests

Groxx:
Does a couple things:
- `make $(BUILD)/generate` now infers files that might affect generation.  This should work as long as we keep simple codegen (the server has moved beyond this, sadly).
- `make generate` now forces re-generation regardless of need, because it's a nice workaround for dependency issues like happened in #1373
- `make test` and related targets now go-generate and fmt before running tests.  Slower, but you can't forget to re-generate, and they're slow already so it's probably fine.

### DIFF
--- a/internal/encoded.go
+++ b/internal/encoded.go
@@ -27,6 +27,8 @@ import (
 	"go.uber.org/cadence/internal/common/util"
 )
 
+//go:generate mockery --name Value --output ../mocks --boilerplate-file ../LICENSE
+
 type (
 	// Value is used to encapsulate/extract encoded value from workflow/activity.
 	Value interface {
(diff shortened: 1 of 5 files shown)


internal/encoded.go
```diff
@@ -27,6 +27,8 @@ import (
 	"go.uber.org/cadence/internal/common/util"
 )
 
+//go:generate mockery --name Value --output ../mocks --boilerplate-file ../LICENSE
```
Groxx, line 30: I moved these so they're in the file with the thing they define (so the dependency setup is trivial), and because it's simpler: no need to specify the internal-package bit.

code generation after this is identical.

3vilhamster (approved):

PR #1378, opened 2024-11-01
Unit tests for activity task handler

3vilhamster:
**Detailed Description**
To improve test coverage, I've moved existing components to separate files and added unit tests for the activity handler.

**Impact Analysis**
- **Backward Compatibility**: Full
- **Forward Compatibility**: Full

**Testing Plan**
- **Unit Tests**: Unit tests and existing integration tests test
- **Persistence Tests**: unrelated
- **Integration Tests**: This could be verified with existing unit/integration tests and on canary deployment.
- **Compatibility Tests**: Unrelated

**Rollout Plan**
- What is the rollout plan? Simple rollout starting with canary
- Is there a kill switch to mitigate the impact immediately? Rollback


internal/auto_heartbeater_test.go
```diff
@@ -0,0 +1,117 @@
...
+			hearbeater.Run(ctx, time.Second)
+			wg.Done()
+		}()
+		clock.BlockUntil(1)
+		clock.Advance(time.Second / 2)
+		cancel()
+
+		require.True(t, util.AwaitWaitGroup(wg, time.Second))
+	})
+	t.Run("heartbeat fail", func(t *testing.T) {
```
taylanisikdemir, line 94: why not use table test to have less redundancy? this approach is hard to maintain especially while the function under test changes. requires changing N test bodies instead of 1

internal/activity_task_handler_test.go
```diff
@@ -0,0 +1,353 @@
...
+		HeartbeatTimeoutSeconds:         common.Int32Ptr(1),
+		WorkflowType: &s.WorkflowType{
+			Name: common.StringPtr("wType"),
+		},
+		WorkflowDomain: common.StringPtr("domain"),
+	}
+	res, err := activityHandler.Execute(tasklist, pats)
+	require.NoError(t, err)
+	_, ok := res.(*s.RespondActivityTaskCompletedRequest)
+	require.True(t, ok, "response is not of type *s.RespondActivityTaskCompletedRequest")
```
taylanisikdemir, line 311: nit: include the actual type of `res` in the message

taylanisikdemir (approved):

internal/auto_heartbeater_test.go
3vilhamster: I'm not sure if I can save much with table tests.
I think a setup will have many customizations and be hard to read.

internal/activity_task_handler_test.go
3vilhamster: Good point. Added.

PR #1385, opened 2024-11-05
Wrote tests for the utils.go file

jakobht:
**What changed?**
Wrote tests for the utils.go file

**Why?**
Improve code coverage

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/common/util/stringer.go
+++ b/internal/common/util/stringer.go
@@ -78,12 +78,12 @@ func valueToString(v reflect.Value) string {
 
 // HistoryEventToString convert HistoryEvent to string
 func HistoryEventToString(e *s.HistoryEvent) string {
-	data := getData(e)
+	data := getHistoryEventData(e)
 
 	return e.GetEventType().String() + ": " + anyToString(data)
 }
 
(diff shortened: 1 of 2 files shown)


internal/common/util/util_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_MergeDicts(t *testing.T) {
+	dic1 := map[string]string{"a": "1", "b": "2"}
+	dic2 := map[string]string{"b": "3", "c": "4"}
+	expected := map[string]string{"a": "1", "b": "3", "c": "4"}
+
+	result := MergeDicts(dic1, dic2)
```
dkrotx, line 37: nit: I feel like case-wise I would add two more cases: dicts have intersection  / same keys, one or both dicts are empty.

dkrotx (approved):

internal/common/util/util_test.go
jakobht: Sure, added some more

PR #1395, opened 2024-11-07
[common] Add unit test for convertions

3vilhamster:
**What changed?**
I've introduced a new helper function PtrOf that could be used instead of helpers. I've left helpers for now since they are used all over the code.
Also, I added tests to conversions and thrift encoders.

**Why?**
Improving code coverage and logic

**How did you test it?**
unit tests and code stability

### DIFF
--- a/internal/common/thrift_util.go
+++ b/internal/common/thrift_util.go
@@ -27,19 +27,13 @@ import (
 	"github.com/apache/thrift/lib/go/thrift"
 )
 
-// TSerialize is used to serialize thrift TStruct to []byte
-func TSerialize(ctx context.Context, t thrift.TStruct) (b []byte, err error) {
-	return thrift.NewTSerializer().Write(ctx, t)
-}
-
 // TListSerialize is used to serialize list of thrift TStruct to []byte
-func TListSerialize(ts []thrift.TStruct) (b []byte, err error) {
+func TListSerialize(ts []thrift.TStruct) ([]byte, error) {
 	if ts == nil {
-		return
+		return nil, nil
 	}
 
 	t := thrift.NewTSerializer()
-	t.Transport.Reset()
 
 	// NOTE: we don't write any markers as thrift by design being a streaming protocol doesn't
 	// recommend writing length.
@@ -48,26 +42,11 @@ func TListSerialize(ts []thrift.TStruct) (b []byte, err error) {
 	ctx := context.Background()
 	for _, v := range ts {
 		if e := v.Write(ctx, t.Protocol); e != nil {
-			err = thrift.PrependError("error writing TStruct: ", e)
-			return
+			return nil, thrift.PrependError("error writing TStruct: ", e)
 		}
 	}
 
-	if err = t.Protocol.Flush(ctx); err != nil {
-		return
-	}
-
-	if err = t.Transport.Flush(ctx); err != nil {
-		return
-	}
-
-	b = t.Transport.Bytes()
-	return
-}
-
-// TDeserialize is used to deserialize []byte to thrift TStruct
-func TDeserialize(ctx context.Context, t thrift.TStruct, b []byte) (err error) {
-	return thrift.NewTDeserializer().Read(ctx, t, b)
+	return t.Transport.Bytes(), t.Protocol.Flush(ctx)
 }
 
 // TListDeserialize is used to deserialize []byte to list of thrift TStruct
@@ -94,13 +73,8 @@ func TListDeserialize(ts []thrift.TStruct, b []byte) (err error) {
 func IsUseThriftEncoding(objs []interface{}) bool {
 	// NOTE: our criteria to use which encoder is simple if all the types are serializable using thrift then we use
 	// thrift encoder. For everything else we default to gob.
-
-	if len(objs) == 0 {
-		return false
-	}
-
-	for i := 0; i < len(objs); i++ {
-		if !IsThriftType(objs[i]) {
+	for _, obj := range objs {
+		if !IsThriftType(obj) {
 			return false
 		}
 	}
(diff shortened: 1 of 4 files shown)


Groxx:
`**message` does work with this.  Using your example and changing just:
```diff
  func main() {
- 	rVal := reflect.ValueOf(&myStruct{})
+ 	var ptr *myStruct
+ 	rVal := reflect.ValueOf(&ptr)
```
gives you `a TStruct`: https://go.dev/play/p/ZCz2fZKuazn


Re changing the behavior: IMO not in this PR, we should hunt for some concrete uses first and be paranoid about breaking any.  But perhaps.  It would bring things more in line with `encoding/json` and that may be worth it.

---

_Should_ we expand this to allow `*message` in addition to `**message`: .... maybe.

`**message` is the only way to show that you got a nil, and it matches the general pattern for encoding:
```go
// encode
value := &someType{a:b} // *someType
json.Marshal(val)

// decode
var value *someType     // same type as above
json.Unmarshal(data, &val)
```

Personally I've seen *a lot* of `*message`-style decoding lead to missed error handling because you have zero values rather than nils that explode immediately, and that goes on to poison later data in dramatically-harder-to-troubleshoot ways.  So I do in fact _always_ use `**message`-style when decoding, and I will _always_ recommend it.  If you want a valid zero value even when nil, then make a `message := &empty{}` explicitly.

It does break from `json.Unmarshal`'s behavior though, and in a lot of ways that sets the tone for all Go (de)serialization code.

---

_Can_ we change this to allow `*message`: after one check, I think yes.  
The only place this would change behavior currently is when the wrong value is being passed in, and it never worked, but now it does.  
Breaking only forever-dead code seems safe enough to me.

The thing to check first though is: currently they're getting fallback-to-JSON behavior.  How does that behave?  I'm kinda assuming it errors, but if apache thrift generates an `UnmarshalJSON` then it may still be doing _something_ currently, and we probably shouldn't change that.

internal/common/thrift_util.go
```diff
@@ -111,13 +85,8 @@ func IsUseThriftEncoding(objs []interface{}) bool {
 func IsUseThriftDecoding(objs []interface{}) bool {
 	// NOTE: our criteria to use which encoder is simple if all the types are de-serializable using thrift then we use
 	// thrift decoder. For everything else we default to gob.
-
-	if len(objs) == 0 {
-		return false
-	}
```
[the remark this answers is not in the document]
3vilhamster: Fixed

internal/common/thrift_util.go
```diff
@@ -94,13 +73,8 @@ func TListDeserialize(ts []thrift.TStruct, b []byte) (err error) {
 func IsUseThriftEncoding(objs []interface{}) bool {
 	// NOTE: our criteria to use which encoder is simple if all the types are serializable using thrift then we use
 	// thrift encoder. For everything else we default to gob.
-
-	if len(objs) == 0 {
-		return false
-	}
```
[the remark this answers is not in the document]
3vilhamster: Fixed

internal/common/thrift_util_test.go
```diff
@@ -0,0 +1,122 @@
...
+	result := IsUseThriftEncoding(ts)
+	assert.True(t, result)
+
+	ts = []interface{}{
+		&mockThriftStruct{},
+		"string",
+	}
+
+	result = IsUseThriftEncoding(ts)
+	assert.False(t, result)
```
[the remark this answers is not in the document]
3vilhamster: Good point; for 2 cases, I was lazy to write a for loop, but with an additional nil case, that makes sense. Updated

jakobht (approved):

internal/common/thrift_util.go
```diff
@@ -48,26 +42,11 @@ func TListSerialize(ts []thrift.TStruct) (b []byte, err error) {
...
-
-	b = t.Transport.Bytes()
-	return
-}
-
-// TDeserialize is used to deserialize []byte to thrift TStruct
-func TDeserialize(ctx context.Context, t thrift.TStruct, b []byte) (err error) {
-	return thrift.NewTDeserializer().Read(ctx, t, b)
+	return t.Transport.Bytes(), t.Protocol.Flush(ctx)
```
Groxx, line 49: this (and not resetting) is a rather significant set of changes - is there something that implies it's correct?  e.g. newer thrift/tchannel docs?

internal/common/thrift_util.go
```diff
@@ -92,15 +71,13 @@ func TListDeserialize(ts []thrift.TStruct, b []byte) (err error) {
...
-	// thrift encoder. For everything else we default to gob.
-
 	if len(objs) == 0 {
 		return false
 	}
-
-	for i := 0; i < len(objs); i++ {
-		if !IsThriftType(objs[i]) {
+	// NOTE: our criteria to use which encoder is simple if all the types are serializable using thrift then we use
+	// thrift encoder. For everything else we default to gob.
```
Groxx, line 78: ~~gob~~ json (we don't use gob anywhere), but this is just a move so it's probably fine.

internal/common/thrift_util.go
3vilhamster: I've looked into thrift.NewTSerializer() and it uses InMemoryTransport that has a comment of Flush is Noop for it, since it uses plain buffer.
Also, Protocol.Flush is just a wrapper around Transport.Flush().

3vilhamster:
```
// Flushing a memory buffer is a no-op
func (p *TMemoryBuffer) Flush(ctx context.Context) error {
	return nil
}
```

3vilhamster: We always initialize one buffer with
```
t := thrift.NewTSerializer()
```
Why it should be reset?

Groxx: Why would it be reset: I agree it doesn't _look_ correct, but the only thing that matters is if it _is_ correct.  If that hasn't been verified, it needs to be, but hasn't been mentioned anywhere so I'm not sure if it has been.

PR #1398, opened 2024-11-08
[common] Introduce testlogger as a workaround of poor lifecycle

3vilhamster:
**What changed?**
I've copied over the workaround racy logs in tests from the backend.
We have the same issues with the client who cannot establish a clear stop for internal processes at the first stop. This is not an issue for real workflow usage but can lead to test failures due to racy log messages after the test ends.

Sidenote:
Changed the approach from using atomic to RWMutex, since this situation requires semantic:
1. We can have concurrent log.Writes - so RLock is not stopping them.
2. We must wait for all writes to go on Cleanup and then acquire the lock
3. All writes after the update should go to fallback.

**Why?**
Improving test stability

**How did you test it?**
Run unit tests a few hundred times.
This is a bit tricky, since the TestLoggerShouldNotFailIfLoggedLate requires rerun, but I've done `for i in {1..100}; do go test . --count 1; done` with race and without.

**Potential risks**
Some tests could be flaky, though they were failing before the fix.

### DIFF
--- a/internal/common/testlogger/testlogger.go
+++ b/internal/common/testlogger/testlogger.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"slices"
 	"strings"
+	"sync"
 
 	"github.com/stretchr/testify/require"
 	"go.uber.org/atomic"
@@ -80,6 +81,7 @@ func NewObserved(t TestingT) (*zap.Logger, *observer.ObservedLogs) {
 }
 
 type fallbackTestCore struct {
+	sync.Mutex
 	t         TestingT
 	fallback  zapcore.Core
 	testing   zapcore.Core
@@ -132,6 +134,9 @@ func (f *fallbackTestCore) Write(entry zapcore.Entry, fields []zapcore.Field) er
 		}
 		return f.fallback.Write(entry, fields)
 	}
+	// Ensure no concurrent writes to the test logger.
+	f.Lock()
+	defer f.Unlock()
 	return f.testing.Write(entry, fields)
 }
 
--- a/internal/common/testlogger/testlogger.go
+++ b/internal/common/testlogger/testlogger.go
@@ -22,10 +22,11 @@ package testlogger
 
 import (
 	"fmt"
-	"go.uber.org/cadence/internal/common"
 	"slices"
 	"strings"
 	"sync"
+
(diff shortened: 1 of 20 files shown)


taylanisikdemir (approved):

internal/common/testlogger/testlogger.go
```diff
@@ -0,0 +1,148 @@
...
+		})
+		if !hasStack {
+			fields = append(fields, zap.Stack("log_stack"))
+		}
+		return f.fallback.Write(entry, fields)
+	}
+	// Ensure no concurrent writes to the test logger.
+	f.Lock()
+	defer f.Unlock()
+	return f.testing.Write(entry, fields)
```
Groxx, line 140: why the added mutex?  zap cores are race-safe, as is `t.Log` - they flush whole lines without interleaving.

Groxx (approved):
LGTM though I don't see the reason for the added mutex.
If something seemed to be flaky without it, I can pretty much _guarantee_ that it didn't fix it, so there's something else that needs fixing instead.

internal/common/testlogger/testlogger.go
Groxx: From some more reading and thinking and IRL chat: I'm pretty sure this doesn't change anything, but it's not risky or anything.

Now that there's a counter-example showing this logger isn't reliable, I think it's pretty clear it's just that the atomics aren't ensuring "fallback logger is used after test completes".  it currently allows: pausing after checking the atomic is not-completed -> test completes -> resuming and using the `T` logger.

But I can make a PR for that if you want.  The mutex here shouldn't make anything worse, and a mutex will be needed to fix the "atomics are not enough" issue anyway.

Groxx:
From some more reading and thinking and IRL chat: I'm pretty sure this doesn't change anything, but it's not risky or anything.  So yeah, still approved, merge any time.

Now that there's a counter-example showing this logger isn't reliable, I think it's pretty clear it's just that the atomics aren't ensuring "fallback logger is used after test completes". it currently allows: pausing after checking the atomic is not-completed -> test completes -> resuming and using the T logger.

But I can make a PR for that if you want. The mutex here shouldn't make anything worse, and a mutex will be needed to fix the "atomics are not enough" issue anyway.

PR #1399, opened 2024-11-08
Document a significant caveat to SideEffect funcs

Groxx:
An unfortunate edge-case-use was discovered recently, and SideEffect does not have adequate protections to prevent breaking workflows at the moment.

For now, just document it.
A true fix will require failing calls to blocking or history-recording funcs while the callback runs, as they cannot be used safely.
Future versions of this API should remove the context arg, to stop implying it can be used.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -265,65 +265,87 @@ func GetSignalChannel(ctx Context, signalName string) Channel {
 	return internal.GetSignalChannel(ctx, signalName)
 }
 
-// SideEffect executes the provided function once, records its result into the workflow history. The recorded result on
-// history will be returned without executing the provided function during replay. This guarantees the deterministic
-// requirement for workflow as the exact same result will be returned in replay.
-// Common use case is to run some short non-deterministic code in workflow, like getting random number or new UUID.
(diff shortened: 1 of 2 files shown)


taylanisikdemir (approved):

workflow/workflow.go
```diff
@@ -265,65 +265,87 @@ func GetSignalChannel(ctx Context, signalName string) Channel {
...
+// (it becomes available later).
+//
+// In other words, in general: use this *only* for code that does not need the workflow.Context.
+// Incorrect context-using calls are not currently prevented in tests or during execution,
+// but they should be eventually.
+//
+//	// Bad example: this will work until a replay occurs,
+//	// but then the workflow will fail to replay with a non-deterministic error.
+//	var out string
+//	err := workflow.SideEffect(func(ctx workflow.Context) interface{} {
```
3vilhamster, line 290: Is there a ticket/initiative to fix this?
We cannot change the API since it is used everywhere, but we could introduce workflow.Context check will fail in the case of history-related operations inside these calls.
Something like:
workflow.Context{
  insideSidefffect: true
}

and inside forbidden things bail out if this flag is set.

PR #1400, opened 2024-11-11
Remove worker hardware utilization code

3vilhamster:
**What changed?**
Reverting #1260 since this code is unused and this logic is not required in the worker code.

**Why?**
To remove unused code and improve code coverage.

**How did you test it?**
Unit tests.

**Potential risks**
In theory, somebody could start using these metrics, but they should not be part of Cadence client responsibilities, and there is better tooling (prom-client) that does it better.


ketsiambaku:
@timl3136 Did we have a change of plans ? wondering why this ended up not being used

ketsiambaku (approved):

PR #1405, opened 2024-11-21
add Concurrency entity for worker

shijiesheng:
**What changed?**

* added worker package for modularity
* added resizable Permit entity wrapping the underlying implementation; these permits are synchronization primitive for concurrency control over poller and tasks processing goroutines
* wrapped buffered channel with Permit interface for task concurrency.
* removed poller autoscaler methods

**Why?**

add dynamic params component towards worker auto configuration
Screenshot 2024-11-21 at 11 24 19 AM

**How did you test it?**

unit test
integration test

**Potential risks**


internal/worker/dynamic_params.go
```diff
@@ -0,0 +1,98 @@
...
+	go func() {
+		defer wg.Done()
+		if err := p.sem.Acquire(ctx, 1); err != nil {
+			close(ch)
+			return
+		}
+		select { // try to send to channel, but don't block if listener is gone
+		case ch <- struct{}{}:
+		default:
+		}
```
shijiesheng, line 79: add release

internal/worker/dynamic_params.go
```diff
@@ -0,0 +1,98 @@
...
+
+var _ Permit = (*permit)(nil)
+
+// Synchronization contains synchronization primitives for dynamic configuration.
+type DynamicParams struct {
+	PollerPermit Permit // controls concurrency of pollers
+	TaskPermit   Permit // controlls concurrency of task processings
+}
+
+// Permit is an adaptive
```
taylanisikdemir, line 39: incomplete comment.

internal/worker/dynamic_params.go
```diff
@@ -0,0 +1,98 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/marusama/semaphore/v2"
+)
+
+var _ Permit = (*permit)(nil)
+
+// Synchronization contains synchronization primitives for dynamic configuration.
```
taylanisikdemir, line 33: Synchronization -> DynamicParams

internal/worker/dynamic_params.go
```diff
@@ -0,0 +1,98 @@
...
+	Count() int
+	Release(count int)
+}
+
+type permit struct {
+	sem semaphore.Semaphore
+}
+
+// NewPermit creates a dynamic permit that's resizable
+func NewPermit(initCount int) Permit {
```
taylanisikdemir, line 54: let's add unit tests  for permit implementation

internal/worker/dynamic_params.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/marusama/semaphore/v2"
+)
+
+var _ Permit = (*permit)(nil)
+
+// Synchronization contains synchronization primitives for dynamic configuration.
+type DynamicParams struct {
```
taylanisikdemir, line 34: nit: DynamicParams is a little bit confusing name here. We have dynamic config feature flags but this has nothing to do with it. Params also indicate it's something set by user/operator. A more suitable name would be `ConcurrencyLimits`

internal/internal_worker_base.go
```diff
@@ -141,7 +142,7 @@ type (
 		logger               *zap.Logger
 		metricsScope         tally.Scope
 
-		pollerRequestCh    chan struct{}
+		dynamic            *worker.DynamicParams
```
taylanisikdemir, line 145: see my other comment on renaming. I'd also rename this member variable to `concurrencyLimits`. `dynamic` is too generic

internal/internal_worker_base.go
```diff
@@ -167,11 +168,18 @@ func createPollRetryPolicy() backoff.RetryPolicy {
...
 	ctx, cancel := context.WithCancel(context.Background())
 
+	dynamic := &worker.DynamicParams{
+		PollerPermit: worker.NewPermit(options.pollerCount),
+		TaskPermit:   worker.NewPermit(options.maxConcurrentTask),
+	}
+
 	var pollerAS *pollerAutoScaler
 	if pollerOptions := options.pollerAutoScaler; pollerOptions.Enabled {
+		dynamic.PollerPermit = worker.NewPermit(pollerOptions.InitCount)
```
taylanisikdemir, line 178: why not use the same `options.pollerCount` to initialize PollerPermit when autoscale is enabled?

internal/worker/dynamic_params.go
```diff
@@ -0,0 +1,98 @@
...
+// THE SOFTWARE.
+
+package worker
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/marusama/semaphore/v2"
```
taylanisikdemir, line 28: I am not sure about maturity/safety of this semaphore library. How did we end up picking this one?

taylanisikdemir (commented):
Changes look safe as long as the underlying semaphore library is not buggy. It will be the critical component determining concurrency controls of client SDK. I recommend deep diving on that to understand the implementation, check for potential deadlock cases and write comprehensive unit/concurrency tests for the wrapper `permit` implementation

internal/internal_worker_base.go
shijiesheng: pollerCount is actually the maximum number of pollers. Without poller auto scaler, this is the "initial" count. For poller auto scaler, I was trying to override the initial poller count by not throttling poller requests when the worker first starts up.

internal/worker/dynamic_params.go
shijiesheng: It's a simply library that is fast and scalable with benchmark data. Reinventing the wheel might be not worth the time. But if we do see issues we can re-assess this and figure out a better approach

internal/worker/dynamic_params.go
shijiesheng: make sense

internal/worker/dynamic_params.go
shijiesheng: added

internal/worker/concurrency.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Acquire is blocking until a permit is acquired or returns error after context is done
+func (p *permit) Acquire(ctx context.Context, count int) error {
+	if err := p.sem.Acquire(ctx, count); err != nil {
+		return fmt.Errorf("failed to acquire permit before context is done: %w", err)
+	}
+	return nil
+}
+
+// AcquireChan returns a permit ready channel. It's closed then permit is acquired
```
taylanisikdemir, line 66: Please add a comment here to mention that caller is supposed to call `Release(1)` after consuming from the returned channel

internal/worker/concurrency_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+func TestPermit_Simulation(t *testing.T) {
+	tests := []struct {
+		name                  string
+		capacity              []int // update every 50ms
+		goroutines            int   // each would block on acquiring 2-6 tokens for 100ms
+		goroutinesAcquireChan int   // each would block using AcquireChan for 100ms
+		maxTestDuration       time.Duration
+		expectFailures        int
+		expectFailuresAtLeast int
```
taylanisikdemir, line 43: if possible let's make the failure validation a range check

internal/worker/concurrency.go
shijiesheng: added

internal/worker/concurrency_test.go
shijiesheng: changed

internal/worker/concurrency.go
```diff
@@ -74,7 +77,7 @@ func (p *permit) AcquireChan(ctx context.Context, wg *sync.WaitGroup) <-chan str
 		}
 		select { // try to send to channel, but don't block if listener is gone
 		case ch <- struct{}{}:
-		default:
+		case <-time.After(10 * time.Millisecond): // wait time is needed to avoid race condition of channel sending
```
shijiesheng, line 80: found a race condition of channel sending. Adding a wait time would be more reliable

Groxx:
Reading through [marusama/semaphore/v2](https://github.com/marusama/semaphore):
- yep, looks likely correct to me.  so 👍 for that library.
- some cautions worth noting:
  - golang.org/x/sync/semaphore does FIFO acquiring, and:
    - it probably performs MUCH better under heavy load, as marusama wakes up all waiters and they all race to acquire
    - it is _strictly_ FIFO, [which prevents starvation of >1 acquires](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.9.0:semaphore/semaphore.go;l=141-154;drc=14be23e5b48bec28285f8a694875175ecacfddb3), which marusama does not prevent
    - x/sync/semaphore does not resize, so it's not really an option anyway
  - there are very clear `uint32` overflow possibilities that it does not check

So... I think we're probably fine.  For pollers we won't run into any of these in practice.  
It's probably worth leaving some comments on our use though, to make sure ^ those are checked before using it elsewhere, and before using it with any un-checked config (e.g. a user sets 2 billion concurrent as "unlimited", and behavior is undefined).

In the meantime we should probably get some integer overflow checks added, and a new release.  There's no reason to allow those to occur.

taylanisikdemir (approved):

internal/worker/concurrency.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/marusama/semaphore/v2"
+)
+
+var _ Permit = (*permit)(nil)
+
+// ConcurrencyLimit contains synchronization primitives for dynamically controlling the concurrencies in workers
+type ConcurrencyLimit struct {
+	PollerPermit Permit // controls concurrency of pollers
+	TaskPermit   Permit // controlls concurrency of task processings
```
Groxx, line 37:
```suggestion
	TaskPermit   Permit // controls concurrency of task processing
```

internal/worker/concurrency.go
Groxx: hmm :\  unfortunately this means that if processing is delayed by 10ms it will block the chan-reader forever.  that's not too unlikely with big CPU spikes, and definitely not impossible.

tbh I think that might rule this impl out entirely.  though I think it's possible to build a `AcquireChan(...) (<-chan struct{}, cancel func())` that doesn't have this issue, and that might be worth doing.

Groxx (changes requested):
I think I'm gonna have to block this for https://github.com/cadence-workflow/cadence-go-client/pull/1405/files#r1859368368 since that's potentially fatal.

Overall though:
- library looks fine (more details above)
- changes look reasonable
  - four-layer-deep `bw.concurrency.TaskPermit.AcquireChan` accesses are a bit dubious in principle, but I think they make sense here.  the added structure/layers seems useful.
  - wrapper as a whole seems good to have, though do we have any plans to do non-`1` values?  I'd get rid of `count` if not, but I suppose that depends on the "resource".  (but see library comment for risks there, e.g. we _definitely_ cannot use it for "bytes" in many cases)
- I have not checked the tests in detail but the high level approach seems useful.  might need some fine-tuning, but it's reasonably "meaningful" and that's a very solid start.

internal/worker/concurrency.go
taylanisikdemir: if we expect chan-reader to always consume from the returned `ch` unless ctx is canceled then we can replace this goroutine implementation with 

```
defer wg.Done()
if err := p.sem.Acquire(ctx, 1); err != nil { 
   return // assuming Acquire only returns err if ctx.Done
}
select {
case ch <- struct{}{}:
case <-ctx.Done():
  p.sem.Release(1)
}
```

internal/worker/concurrency.go
shijiesheng: fixed

internal/worker/concurrency.go
shijiesheng: fixed through Taylan's suggestion. It should be safe now.

shijiesheng:
> I think I'm gonna have to block this for https://github.com/cadence-workflow/cadence-go-client/pull/1405/files#r1859368368 since that's potentially fatal.
> 
> Overall though:
> 
> * library looks fine (more details above)
> * changes look reasonable
>   
>   * four-layer-deep `bw.concurrency.TaskPermit.AcquireChan` accesses are a bit dubious in principle, but I think they make sense here.  the added structure/layers seems useful.
>   * wrapper as a whole seems good to have, though do we have any plans to do non-`1` values?  I'd get rid of `count` if not, but I suppose that depends on the "resource".  (but see library comment for risks there, e.g. we _definitely_ cannot use it for "bytes" in many cases)
> * I have not checked the tests in detail but the high level approach seems useful.  might need some fine-tuning, but it's reasonably "meaningful" and that's a very solid start.
> 
> I'm... not entirely sure what to do to resolve the core "resizable semaphore but we also need chans" issue tbh. We might end up needing / badly-wanting chans, so that may be important. I suspect there's some other library, but I haven't hunted for one or thought too much on what it'd need to be correct 🤔

Double checked. We don't do non-one values in acquire at any place. I'll remove it.

internal/internal_worker_base.go
shijiesheng: changed

internal/worker/concurrency.go
```diff
@@ -0,0 +1,100 @@
...
+			return
+		}
+		select { // try to send to channel, but don't block if listener is gone
+		case ch <- struct{}{}:
+		case <-ctx.Done():
+			p.sem.Release(1)
+		}
+	}()
+	return ch
+}
```
Groxx, line 84: hmm.  this is _possible_ to use correctly, and doesn't leak goroutines beyond `ctx.Done()` which is good (and possibly good enough), but it still leaves the chan permanently blocking in some cases.

that's correct as long as the `AcquireChan(...)` caller _also_ listens to the same / a derived `ctx.Done()` and stops using the chan if that occurs.  and ideally also cancels the context when it stops reading in all other branches.

`func (bw *baseWorker) runPoller() {` does this currently, because `bw.shutdownCh` closes when `bw.limiterContext` is canceled and there effectively is no timeout, but it feels kinda error-prone 🤔

Groxx:
To stick some slack-chatting publicly:
- I'm game to approve with some detailed docs and probably a change to the main loop to prevent future risks (i.e. also select on the `limiterContext` + cancellation, or docs so furure reviewers catch it)
- sounds like we might be dropping the attempt in general, not needing a dynamically-resizing Permit

internal/common/autoscaler/autoscaler.go
```diff
@@ -24,10 +24,6 @@ package autoscaler
 type (
 	AutoScaler interface {
 		Estimator
-		// Acquire X ResourceUnit of resource
-		Acquire(ResourceUnit) error
-		// Release X ResourceUnit of resource
-		Release(ResourceUnit)
```
shijiesheng, line 30: these methods are unnecessary after `Permit` introduction

shijiesheng:
> To stick some slack-chatting publicly:
> 
> * I'm game to approve with some detailed docs and probably a change to the main loop to prevent future risks (i.e. also select on the `limiterContext` + cancellation, or docs so future reviewers catch it)
> * sounds like we might be dropping the attempt in general, not needing a dynamically-resizing Permit.  if so this is all sorta moot, but I'm happy to review if not.

correct, I dropped the resizable permit implementation for task permit and still uses channel. So the controversial error-prone async AcquireChan implementation is no longer needed. 

The plan is to add guardrail on resources (CPU/memory) for concurrencies so users can increase the max limit safely.

internal/worker/channel_permit.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (p *channelPermit) Acquire(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("failed to acquire permit before context is done")
+	case p.channel <- struct{}{}:
+		return nil
+	}
+}
```
Groxx, line 48: pretty strong sign that there need to be more tests: without a `GetChan()` reader, this will never succeed (and then it'll allow too many things / may block release permanently)

internal/worker/concurrency.go
```diff
@@ -0,0 +1,48 @@
...
+	TaskPermit   ChannelPermit // controls concurrency of task processing
+}
+
+// Permit is an adaptive permit issuer to control concurrency
+type Permit interface {
+	Acquire(context.Context) error
+	Count() int
+	Quota() int
+	Release()
+	SetQuota(int)
```
Groxx, line 42: Seems worth splitting this out so you don't have to no-op it.  It isn't even needed - this all compiles fine with `SetQuota` only defined on a non-channel permit.

internal/common/autoscaler/autoscaler.go
Groxx: with this removed, this is basically a "periodically update a cache and broadcast" thing, rather than a resource-controller.

which seems fine, just checking that that's what is needed.  without anything using the hooks or `GetCurrent()` I'm kinda struggling to figure out what's needed eventually vs something we should get rid of while we can.

shijiesheng:
close this PR as it has so many questions and risks

shijiesheng:
reworked in https://github.com/cadence-workflow/cadence-go-client/pull/1410

PR #1406, opened 2024-12-02
[cadence/debug] Update polletracker to support poller type breakdown

ketsiambaku:
**What changed?**
We use the PollerTracker interface to count the number of running pollers and we display it internal in a debug page.
This change is updating the interface and its usage so that the poller count is broken down by worker type (e.g: activity, workflow, shadow, localactivity, etc.)

**Why?**
The break down gives a better insight on the running instance

**How did you test it?**
unit tests

**Potential risks**

### DIFF
--- a/internal/common/debug/example_test.go
+++ b/internal/common/debug/example_test.go
@@ -25,18 +25,19 @@ import (
 	"fmt"
 	"sort"
 	"sync"
-
-	"go.uber.org/atomic"
 )
 
 type (
 	// pollerTrackerImpl implements the PollerTracker interface
 	pollerTrackerImpl struct {
-		pollerCount atomic.Int32
+		sync.RWMutex
+		count map[string]int64
 	}
 
 	// stopperImpl implements the Stopper interface
 	stopperImpl struct {
+		sync.Once
+		workerType    string
 		pollerTracker *pollerTrackerImpl
 	}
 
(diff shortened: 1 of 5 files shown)


zawadzkidiana (approved):
disregard

dianazawadzki:
> Nice, focused improvement — breaking the poller count down by worker type (activity/workflow/local-activity/shadow) is genuinely more useful on the debug page than a single aggregate number, and this doesn't grow the public surface more than it needs to.
> 
> A few things I liked:
> 
> * Switching `stopperImpl` to embed `sync.Once` fixes a latent bug in the old code: previously calling `Stop()` twice on the same stopper would double-decrement the counter. That's a real correctness improvement, not just refactoring.
> * `Stats()` deleting zero-count entries from the map keeps the output clean (no `"ActivityWorker": 0` noise) and the sort in `Stats()` makes the JSON output deterministic, which matters for a debug page and for the example test's `// Output:` comparison.
> * The example test was updated to actually exercise multiple worker types concurrently going up and down, which is a better story than the old single-counter version.
> 
> Two small things worth a look, non-blocking:
> 
> 1. `Stats()` now returns `nil` instead of an empty slice when there are no active pollers (confirmed by `workerstats_noop_test.go` asserting `assert.Nil`). That's a reasonable choice since `nil` and `[]Pollers{}` marshal to the same thing (`null` vs `[]`, technically different but both "empty"-ish) — just flagging in case any consumer of `Stats()` does a `len()` check or JSON-diffs against `[]` rather than `null`.
> 2. Since `count` is now a `map[string]int64` guarded by a mutex instead of an atomic counter, a `go test -race` run exercising concurrent `Start`/`Stop` from multiple goroutines (rather than just the sequential example test) would give good confidence there's no data race in the map access — might already be covered by CI's race detector on `go test ./...`, but worth double-checking since this is worker-hot-path code.
> 
> Housekeeping: this branch shows as behind `master` and no checks have run yet (0 statuses reported) — likely just needs a rebase to kick off CI (and possibly a maintainer approval to run Actions, since I don't see any prior CI activity on this PR at all). Given how small and self-contained the diff is, that should be quick. Thanks for picking this up — sorry it sat without a look for so long!
> 

Sorry about this Claude comment, reviewing with human eyes and will follow up shortly!
-Diana

dianazawadzki (changes requested):
Picking this up, sorry it sat since December.

Still relevant: master hasn't done this. internal/common/debug/types.go still
has PollerTracker.Stats() int32 returning the single aggregate. So this isn't
superseded and is worth landing.

One thing to call out before it does. debug/types.go re-exports these as type
aliases:

    PollerTracker = internal.PollerTracker

so changing Stats()'s return type propagates to the public
go.uber.org/cadence/debug package, and anyone with their own PollerTracker
implementation gets a compile error on upgrade. Every type there carries
"Deprecated: in development and very likely to change", so this is sanctioned
rather than forbidden, but the Potential risks section is currently empty and
this belongs in it, along with a CHANGELOG entry so it isn't a surprise at
upgrade time.

Merge state: behind master, and the only checks that have ever run on this
branch are codecov/patch and codecov/project. No build, lint, or unit tests
have ever run here. A rebase should kick those off.

When it does, confirm the race detector is actually exercising the new map
access. Swapping an atomic counter for a mutex-guarded map[string]int64 in
worker hot-path code isn't something the sequential example test will tell you
much about. The sync.Once change on stopperImpl is a genuine bug fix on its
own, a double Stop() previously double-decremented, so that stands regardless.

Next: rebase, fill in Potential risks with the public alias break, add a
CHANGELOG line, confirm CI including the race detector. Happy to approve after
that.

PR #1408, opened 2024-12-03
update IDL

shijiesheng:
**What changed?**

* new endpoint `DiagnoseWorkflowExecution`
* new field `AutoConfigHint` in Poll responses

**Why?**

new field is needed for worker auto config project

**How did you test it?**

unit test

**Potential risks**

### DIFF
--- a/internal/common/auth/service_wrapper.go
+++ b/internal/common/auth/service_wrapper.go
@@ -113,6 +113,16 @@ func (w *workflowServiceAuthWrapper) DescribeWorkflowExecution(ctx context.Conte
 	return result, err
 }
 
+func (w *workflowServiceAuthWrapper) DiagnoseWorkflowExecution(ctx context.Context, request *shared.DiagnoseWorkflowExecutionRequest, opts ...yarpc.CallOption) (*shared.DiagnoseWorkflowExecutionResponse, error) {
+	tokenHeader, err := w.getYarpcJWTHeader()
+	if err != nil {
+		return nil, err
+	}
+	opts = append(opts, *tokenHeader)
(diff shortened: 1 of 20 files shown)


internal/compatibility/proto/request.go
```diff
@@ -81,6 +81,16 @@ func DescribeWorkflowExecutionRequest(t *shared.DescribeWorkflowExecutionRequest
 	}
 }
 
+func DiagnoseWorkflowExecutionRequest(t *shared.DiagnoseWorkflowExecutionRequest) *apiv1.DiagnoseWorkflowExecutionRequest {
```
taylanisikdemir, line 84: I am not super happy with this API's response. It just returns wfid/runid corresponding to the diagnostic workflow which means user has to make subsequent API calls to get diagnosis results. I think this was designed to be used by CLI but we have to expose it here in SDKs too unfortunately.

internal/common/auth/service_wrapper.go
```diff
@@ -113,6 +113,16 @@ func (w *workflowServiceAuthWrapper) DescribeWorkflowExecution(ctx context.Conte
 	return result, err
 }
 
+func (w *workflowServiceAuthWrapper) DiagnoseWorkflowExecution(ctx context.Context, request *shared.DiagnoseWorkflowExecutionRequest, opts ...yarpc.CallOption) (*shared.DiagnoseWorkflowExecutionResponse, error) {
```
taylanisikdemir, line 116: please update unit tests

shijiesheng: fixed

taylanisikdemir (approved):

PR #1409, opened 2024-12-05
poller auto scaler rework

shijiesheng:
**What changed?**

* removed autoscaler package; it's no longer used after the rework
* Used poller wait time from autoConfigHint field for poller scaling
* Scaling now happens every 1s with a sustained cooldown time instead of periodically 1 min change to react to traffic change quicker

**Why?**

**How did you test it?**

**Potential risks**


internal/worker/concurrency.go
```diff
@@ -18,21 +18,28 @@
...
+
+// ConcurrencyLimit contains synchronization primitives for dynamically controlling the concurrencies in workers
+type ConcurrencyLimit struct {
+	PollerPermit Permit // controls concurrency of pollers
+	TaskPermit   Permit // controls concurrency of task processing
+}
+
+// Permit is an adaptive permit issuer to control concurrency
+type Permit interface {
+	Acquire(context.Context) error
```
Shaddoll, line 35: do you mind adding comments for each method?

shijiesheng:
closed and see https://github.com/cadence-workflow/cadence-go-client/pull/1411

internal/worker/concurrency.go
shijiesheng: sorry this branch was accidentally pushed. I'm still testing this branch https://github.com/cadence-workflow/cadence-go-client/pull/1411

PR #1410, opened 2024-12-05
add ConcurrencyLimit to worker to enable dynamic tuning of concurrencies

shijiesheng:
**What changed?**

[**High Risk**]
* replaced buffered channel with resizable semaphore to control task concurrency

[Low Risk]
* added worker package for modularity
* added ConcurrencyLimit entity to worker 
* removed unused methods of autoscaler interface

**Why?**

needed as first step to enable dynamic tuning of poller and task concurrencies

**How did you test it?**

Unit Test
[WIP] Canary Test + Bench Test

**Potential risks**


internal/worker/resizable_permit.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Count returns the number of permits available
+func (p *resizablePermit) Count() int {
+	return p.sem.GetCount()
+}
+
+// AcquireChan creates a PermitChannel that can be used to wait for a permit
+// After usage:
+// 1. avoid goroutine leak by calling permitChannel.Close()
+// 2. release permit by calling permit.Release()
```
taylanisikdemir, line 72: there's no .Release() exposed

internal/worker/resizable_permit.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Count returns the number of permits available
+func (p *resizablePermit) Count() int {
+	return p.sem.GetCount()
+}
+
+// AcquireChan creates a PermitChannel that can be used to wait for a permit
+// After usage:
+// 1. avoid goroutine leak by calling permitChannel.Close()
```
Groxx, line 71: I suppose this is also "or cancel the context".
A cancel-helper does ensure waiting + need fewer temp-vars though, so I kinda like it 👍.  also easier to ensure it happens because it's special.

---

It's _somewhat_ more common to return a `<-chan, cancel func()` tuple instead of an interface though, and a bit more forget-resistant because you're forced to notice that there are two values instead of one value with an unknown number of methods.

I don't feel too strongly, but if ^ that's convincing to you I'd be happy to re-stamp.  Changes should be pretty simple.

internal/worker/resizable_permit.go
```diff
@@ -0,0 +1,117 @@
...
+func (p *resizablePermit) Acquire(ctx context.Context) error {
+	if err := p.sem.Acquire(ctx, 1); err != nil {
+		return fmt.Errorf("failed to acquire permit before context is done: %w", err)
+	}
+	return nil
+}
+
+// Release release one permit
+func (p *resizablePermit) Release() {
+	p.sem.Release(1)
```
taylanisikdemir, line 51: how does semaphore behave if Release() is called multiple times? does it increase the capacity?

internal/worker/resizable_permit_test.go
```diff
@@ -0,0 +1,144 @@
...
+			capacity:            []int{200},
+			goroutines:          1000,
+			expectFailuresRange: []int{0, 0},
+		},
+		{
+			name:                "not enough permit for some to acquire, fail some",
+			maxTestDuration:     250 * time.Millisecond, // at least need 100ms * 1000 / 200 = 500ms to acquire all permit
+			capacity:            []int{200},
+			goroutines:          1000,
+			expectFailuresRange: []int{1, 999}, // should at least pass some acquires
```
Groxx, line 62: probably means we can make the bounds more precise, like `400,600`?  Loose bounds make sense to avoid noise, but it's targeting 500 and tends to be fairly close.

or maybe just a "should be ~500" comment.

internal/worker/resizable_permit_test.go
```diff
@@ -0,0 +1,144 @@
...
+			capacity:            []int{200, 400, 600},
+			goroutines:          1000,
+			expectFailuresRange: []int{0, 0},
+		},
+		{
+			name:                "enough permit at beginning but due to capacity change, some would fail",
+			maxTestDuration:     250 * time.Millisecond,
+			capacity:            []int{600, 400, 200},
+			goroutines:          1000,
+			expectFailuresRange: []int{1, 999},
```
Groxx, line 76: should expect...
- 0ms: 1k enter, 600 accepted, 400 waiting
- 50ms: resized down to 400, 400 still waiting because none have released
- 100ms: resized down to 200, *some* of the original 600 begin releasing (100-150ms)
- 150ms: all 600 released, 200 acquired (waiting at least 50ms more before any release)
- 200ms: 100+100ms start releasing (lower bound)
- 250ms: 100+100+(50ms total rand) time out

so... at least 800 success?  ignoring jitter.

internal/worker/resizable_permit_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"math/rand"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/atomic"
+)
+
```
Groxx, line 34: the simulation's great but it does hide any bugs that'd occur from using only one or the other mode, and that's how this is going to be used AFAICT.

seems worth some very simple tests for "can get" + "can fail from timeout/cancel", but I believe they'll all pass.

internal/worker/resizable_permit.go
Groxx: It'll (essentially) increase capacity until it exceeds the limit, it's not a `sync.Once` or equivalent that can guarantee at-most-once-per-acquire.

When it exceeds the limit (possibly/probably at some later time), it panics: https://github.com/marusama/semaphore/blob/master/semaphore.go#L170

Groxx (approved):
Tests are failing (extra release) and minor comments (mostly on comments), but 👍 looks good AFAICT and let me know if you tackle other tests / changing the return format.

internal/worker/resizable_permit.go
Groxx: there's a bit of a fundamental tradeoff between "ignore misuse, allow multiple calls" and "require exactly correct use, but it might be hard to find the cause".

tbh personally I prefer "require exactly correct use" in most cases.  it's easy to convert that to "ignore misuse" with a `sync.Once.Do(cancel)` wrapper, e.g. for stuff like https://github.com/cadence-workflow/cadence/blob/master/common/clock/ratelimiter.go#L354 where it's convenient to use both `defer` and early-unlock, and not have to worry about the many combinations possible.

internal/internal_poller_autoscaler_test.go
```diff
@@ -190,9 +193,9 @@ func Test_pollerAutoscaler(t *testing.T) {
 				go func() {
 					defer wg.Done()
 					for pollResult := range pollChan {
-						pollerScaler.Acquire(1)
+						pollerScaler.permit.Acquire(context.Background())
```
taylanisikdemir, line 196: `Acquire` returns error which should be handled. otherwise `Release` might be called without acquire succeeding which leads to panic

shijiesheng: Good catch. I have been searching for the issue for hours.

internal/worker/resizable_permit_test.go
shijiesheng: make sense. I've loosen the bound

internal/worker/resizable_permit_test.go
shijiesheng: I'll just say at most 500 failures. Jitter is making things more complicated

internal/worker/resizable_permit_test.go
```diff
@@ -145,18 +145,139 @@ func TestPermit_Simulation(t *testing.T) {
...
+	})
+
+	t.Run("cancel acquire", func(t *testing.T) {
+		permit := NewResizablePermit(1)
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		err := permit.Acquire(ctx)
+		assert.ErrorContains(t, err, "canceled")
+		assert.Empty(t, permit.Count())
+	})
```
Groxx, line 176: these two tests are the same fwiw - you probably want to have one that does
- acquire (use up the whole semaphore)
- acquire again (blocks until timeout)
- make sure it didn't return immediately (elapsed time > like 5ms)

internal/worker/resizable_permit_test.go
```diff
@@ -145,18 +145,139 @@ func TestPermit_Simulation(t *testing.T) {
...
+		case <-ctx.Done():
+			t.Errorf("permit not acquired")
+		}
+	})
+
+	t.Run("acquire timeout", func(t *testing.T) {
+		permit := NewResizablePermit(1)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
+		defer cancel()
+		time.Sleep(100 * time.Millisecond)
```
Groxx, line 242: same as above, this makes it identical to the cancel case below (the chan is closed before it starts)

internal/worker/resizable_permit.go
Groxx: yea, there are a few unfortunately-complicated delayed-release things in the client :\  I'd love to get rid of as many as possible, or make them much clearer if not possible.

Groxx:
minor test gap but 👍 looks good.
did you find out what the double-release earlier failure was coming from, or was that maybe just a flaky test?

PR #1411, opened 2024-12-10
rework on poller auto scaler

shijiesheng:
**Detailed Description**

Improve performance of poller auto scaler by using more accurate scaling signals and several implementation changes. 

Changes
* New WorkerOptions `AutoScalerOptions` is introduced. 
* Several WorkerOptions are deprecated and become no-op. 
* read new signal (poller wait time) to scale 
* allow kill switching poller auto scaler from server
* new implementation that makes scaling quicker to traffic change
* removed no longer used autoscaler package completely (original implementation is over complicated)

**Impact Analysis**
- **Backward Compatibility**: **NO** existing autoscaling will be stopped but this shall not have big impact since this feature was never rolled out in production. For OSS users, please follow the instructions below in rollout plan.
- **Forward Compatibility**: Yes, introduce new 

**Testing Plan**
- **Unit Tests**: Yes
- **Persistence Tests**: Not related
- **Integration Tests**: No
- **Compatibility Tests**: No, because it's autoscaler is a feature that was not rolled out in production.

**Rollout Plan**
- What is the rollout plan?
For Uber services, standard client release steps
For OSS users, turn off autoscaler feature first before the client upgrade. 

- Is there a kill switch to mitigate the impact immediately? Yes, the new autoscaler feature is an opt-in feature.


internal/internal_worker_base.go
```diff
@@ -421,3 +422,20 @@ func (bw *baseWorker) Stop() {
 	}
 	return
 }
+
+func getAutoConfigHint(task interface{}) *shared.AutoConfigHint {
+	switch t := task.(type) {
```
Groxx, line 427: when I was poking through these types, to see if this was complete, I believe I found a fourth.  I don't remember if it was locally-dispatched activity types or query tasks or something else though.

what happens if this is wrong?  does it just consider [something] as having no cost?

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,323 @@
...
+					for v := range inputChan {
+						r.Add(v)
+						doneC <- struct{}{}
+					}
+				}()
+			}
+
+			for i := range tt.input {
+				inputChan <- tt.input[i]
+				<-doneC
```
Groxx, line 316: this is serializing the whole thing, so it doesn't really test anything concurrently - you can even remove the mutexes and it'll still pass the race detector.

what's the goal with the concurrency?  it's how it's used so 👍 concurrency in tests is good, but this has artificially nerfed it.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,263 @@
...
+	// no op on zero rolling window
+	if len(r.window) == 0 {
+		return
+	}
+
+	// replace the old value with the new value
+	r.index %= len(r.window)
+	r.sum += value - r.window[r.index]
+	r.window[r.index] = value
+	r.index++
```
Groxx, line 249: general design critique: where possible, make sure all state valid when it leaves the function boundaries.

e.g. in this case, r.index is allowed to be ==len (and unsafe to use) every `len(r.window)` times it's called.  if this were instead:
```suggestion
	r.sum += value - r.window[r.index]
	r.window[r.index] = value
	r.index++
	r.index %= len(r.window)
```
it'd always be correct.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+func (c *ConcurrencyAutoScaler) logEvent(event autoScalerEvent) {
+	if c.enable.Load() {
+		c.scope.Counter("concurrency_auto_scaler.enabled").Inc(1)
+	} else {
+		c.scope.Counter("concurrency_auto_scaler.disabled").Inc(1)
+	}
+	c.scope.Gauge("poller_in_action").Update(float64(c.concurrency.PollerPermit.Count()))
+	c.scope.Gauge("poller_quota").Update(float64(c.concurrency.PollerPermit.Quota()))
```
Groxx, line 183: no complaint at all about emitting both, but: when are we expecting them to be different, aside from for extremely small amounts of time (between "poll complete" and "started the next poll")?

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,263 @@
...
+		shouldEnable = true
+	}
+	if switched := c.enable.CompareAndSwap(!shouldEnable, shouldEnable); switched {
+		if shouldEnable {
+			c.logEvent(autoScalerEventEnable)
+		} else {
+			c.resetConcurrency()
+			c.logEvent(autoScalerEventDisable)
+		}
+	}
```
Groxx, line 168: tbh I think I'd prefer just using a mutex.

right now there's nothing preventing:
- `updatePollerPermit` begins, checks enable...
- disable
- reset concurrency
- ... `updatePollerPermit` sets quota

and leaving things in an abnormal state with no ability to notice it.

with the possible exception of metrics/log buffer flushing and whatnot, there's no I/O in here and no expected lock contention - should be no issue at all with using a mutex.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,263 @@
...
+	c.wg.Wait()
+	c.logEvent(autoScalerEventStop)
+}
+
+// ProcessPollerHint reads the poller response hint and take actions
+// 1. update poller wait time
+// 2. enable/disable auto scaler
+func (c *ConcurrencyAutoScaler) ProcessPollerHint(hint *shared.AutoConfigHint) {
+	if hint == nil {
+		c.log.Warn("auto config hint is nil, this results in no action")
```
Groxx, line 146: maybe just info?  otherwise this'll make a flood of warns in user services when rolling out or disabling, but it's not really an unexpected / worrying thing in that state.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/uber-go/tally"
+	"go.uber.org/zap"
+
+	"go.uber.org/cadence/.gen/go/shared"
+)
+
+const (
+	defaultAutoScalerUpdateTick = time.Second
+	// concurrencyAutoScalerObservabilityTick = time.Millisecond * 500
+	targetPollerWaitTimeInMsLog2  = 4 // 16 ms
```
Groxx, line 39: took a while to figure out what exactly the goal here was, I think because the log2/exp2 steps are so separated...

might be easier to follow if you make this a normal ms value, and do a `current*log2(actual/target)` instead?  feels more like a smoothing operation at that point / a "react to the magnitude rather than the value", where currently it's kinda hidden as a `current*log2(actual)/log2(target)` that's being done in three separate locations.

re `TimeInMs` vs a `time.Duration`: seems fine to me.  it's a ton of value-casting noise otherwise, sadly.  and/or maybe we should consider a "duration math" util somewhere, as we do a moderate amount of duration-related floaty calculations in both client and server, and they're super verbose.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,263 @@
...
+func (c *ConcurrencyAutoScaler) updatePollerPermit() {
+	if !c.enable.Load() { // skip update if auto scaler is disabled
+		c.logEvent(autoScalerEventPollerSkipUpdateNotEnabled)
+		return
+	}
+	updateTime := c.clock.Now()
+	if updateTime.Before(c.pollerPermitLastUpdate.Add(c.cooldown)) { // before cooldown
+		c.logEvent(autoScalerEventPollerSkipUpdateCooldown)
+		return
+	}
```
Groxx, line 203: since this func (and therefore `updateTime`'s last value) is only ever called in a loop driven by a timer, it seems like this only has two modes of operation:
- cooldown < tick: every tick causes an update
- cooldown > tick: every N ticks cause an update

neither of which seems particularly useful.

Groxx:
to stick it in here too: overall looks pretty good.  simpler and the overall goal is clearer too.  seems like just minor tweaks (many optional) and it's probably good to go

internal/worker/concurrency_auto_scaler.go
shijiesheng: It's actually updated when poller permit is updated. It serves as a sustain time to avoid updating multiple times in a small time window.

internal/worker/concurrency_auto_scaler.go
shijiesheng: make sense. Maybe just poller quota is needed here

internal/worker/concurrency_auto_scaler.go
shijiesheng: makes sense. I'll mark it as info level

internal/worker/concurrency_auto_scaler.go
shijiesheng: The only concern is currently I'm averaging the wait time. `average(log2(wait_time))`
If I adopt this approach, large wait time will play a bigger role in the average, which is not desired.

internal/worker/concurrency_auto_scaler.go
shijiesheng: make sense

internal/internal_worker_base.go
```diff
@@ -301,7 +308,7 @@ func (bw *baseWorker) pollTask() {
 	var err error
 	var task interface{}
 
-	if bw.pollerAutoScaler != nil {
+	if bw.concurrencyAutoScaler != nil {
 		if pErr := bw.concurrency.PollerPermit.Acquire(bw.limiterContext); pErr == nil {
```
3vilhamster, line 312: nit: this looks like a leaking abstraction. This should be handled inside concurrencyAutoScaler.
I suggest moving all 
concurrencyAutoScaler != nil checks inside methods where it is required.
This code should be simpler. Just calling methods on autoscaler. If it is nil, do nothing.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,274 @@
...
+		defer c.wg.Done()
+		ticker := c.clock.NewTicker(c.updateTick)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-c.shutdownChan:
+				return
+			case <-ticker.Chan():
+				c.logEvent(autoScalerEventMetrics)
+				c.lock.Lock()
```
3vilhamster, line 134: nit: push lock/unlock to updatePollerPermit, then you can use defer inside the function and ensure that unlock happens if anything will cause panic.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+}
+
+func (c *ConcurrencyAutoScaler) Start() {
+	c.logEvent(autoScalerEventStart)
+
+	c.wg.Add(1)
+
+	go func() {
+		defer c.wg.Done()
```
3vilhamster, line 125: nit: any calls to go should have a panic handler.
If a bug exists, it will crash the worker process, significantly impacting customer service.
This is an optional functionality that should be safe to break. Worst case, it won't update concurrency.

3vilhamster (commented):
Overall looks good, but I left some nits

internal/internal_worker_base.go
shijiesheng: make sense

internal/worker/concurrency_auto_scaler.go
shijiesheng: good catch

internal/worker/concurrency_auto_scaler.go
shijiesheng: right, it's simpler

internal/internal_worker_base.go
davidporter-id-au: minor nonblocking suggestion: not to be too OO pilled or wahtever, If this is a property of the task type, maybe make it a method on the task interface? 

ie, instead consider 
```
type autoConfigAwareTask interface {
  getAutoConfigHint() *shared.AutoConfigHint 
}

// for each task, make them responsible for their internal getters and internal state
func (t *workflowTask) getAutoConfigHint() {
  return t.shared.AutoConfigHint 
}
```

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,305 @@
...
+	if hint.PollerWaitTimeInMs != nil {
+		waitTimeInMs := *hint.PollerWaitTimeInMs
+		c.pollerWaitTime.Add(time.Millisecond * time.Duration(waitTimeInMs))
+	}
+
+	var shouldEnable bool
+	if hint.EnableAutoConfig != nil && *hint.EnableAutoConfig {
+		shouldEnable = true
+	}
+	if shouldEnable != c.enabled { // flag switched
```
davidporter-id-au, line 168: nit: consider an early abort?

```
if hint.EnableAutoConfig == nil || !*hint.EnableAutoConfig {
   return 
}
``

File: internal/worker/concurrency_auto_scaler.go [hunk 13]
```diff
@@ -0,0 +1,305 @@
...
+	}
+}
+
+func (r *rollingAverage[T]) Average() T {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	if r.count == 0 {
+		return 0
+	}
+	return r.sum / T(r.count)
```
davidporter-id-au, line 304: are there any conditions under which sum might become zero?

shijiesheng: Yes, if Average() is called before any `Add` method is called, the returned value will be zero. For usages of this rollingAverage, it needs to handle zero value case. 
In auto scaler, I handled it through a `SmoothFunc`

internal/worker/concurrency_auto_scaler.go
shijiesheng: Actually, we need a resetConcurrency() action here if the pollers change from enabled to disabled state to avoid locking them in a unhealthy state. 

For example, when users found an issue and wish to turn off auto scaler, it's best to reset concurrency limit to original value.

internal/internal_worker_base.go
shijiesheng: > what happens if this is wrong? does it just consider [something] as having no cost and... do what?

For new tasks or unrelated tasks, getAutoConfigHint would return nil and thus skip the auto scaler logic. So these tasks won't enter into autoscaler's calculation and just get ignored. 

> minor nonblocking suggestion: not to be too OO pilled or wahtever, If this is a property of the task type, maybe make it a method on the task interface?

sure. added

internal/internal_task_handlers.go
```diff
@@ -153,6 +165,20 @@ type (
 	}
 )
 
+func (t *workflowTask) getAutoConfigHint() *s.AutoConfigHint {
+	if t.task != nil {
+		return t.task.AutoConfigHint
```
taylanisikdemir, line 173: do we need to check whether `t.task.AutoConfigHint` is nil?

internal/internal_worker_base.go
taylanisikdemir: we are guarding calls againts `bw.concurrency` based on nilness of `bw.concurrencyAutoScaler` which indicates that these two should be abstracted behind a single interface to avoid additional complexity in this file

internal/internal_worker_base.go
```diff
@@ -421,3 +424,20 @@ func (bw *baseWorker) Stop() {
...
+		}
+		return t.autoConfigHint
+	case *activityTask:
+		if t.task != nil {
+			return t.task.AutoConfigHint
+		}
+		return t.autoConfigHint
+	default:
+		return nil
+	}
```
taylanisikdemir: instead of this switch case (which is not future proof), we can cast the task to `autoConfigHintAwareTask` interface and get the auto config hint

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,305 @@
...
+	"go.uber.org/zap"
+
+	"go.uber.org/cadence/.gen/go/shared"
+	"go.uber.org/cadence/internal/common/metrics"
+)
+
+const (
+	defaultAutoScalerUpdateTick   = time.Second
+	lowerPollerWaitTime           = 16 * time.Millisecond
+	upperPollerWaitTime           = 256 * time.Millisecond
```
taylanisikdemir: it looks like we would want to iterate on these to adjust sensitivity. consider exposing these to worker config

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,338 @@
...
+			[]eventLog{
+				{autoScalerEventStart, false, 100, "00:00:00"},
+				{autoScalerEventEnable, true, 100, "00:00:00"},
+				{autoScalerEventPollerSkipUpdateCooldown, true, 100, "00:00:01"},
+				{autoScalerEventPollerUpdate, true, 200, "00:00:02"},
+				{autoScalerEventStop, true, 200, "00:00:02"},
+			},
+		},
+		{
+			"idl pollers waiting for tasks",
```
taylanisikdemir: nit: typo idl**e**. same in other cases below

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,338 @@
...
+		enabled     bool
+		pollerQuota int64
+		time        string
+	}
+
+	for _, tt := range []struct {
+		name               string
+		pollAutoConfigHint []*shared.AutoConfigHint
+		expectedEvents     []eventLog
+	}{
```
taylanisikdemir, line 76: would be nice to add a case where it scales up and down a few times

internal/internal_worker_base.go
shijiesheng: i've removed this check in all places. Regarding the comment `hese two should be abstracted behind a single interface to avoid additional complexity in this file`, I still think this is two separate entities. Client still needs concurrency whether autoscaler is enabled or not.

internal/worker/concurrency_auto_scaler.go
shijiesheng: I later found a race condition. I actually need to lock on both `logEvent` and `updatePollerPermit`.  The way I avoid deadlocks is to only lock/unlock on exported methods. Locking on helper methods would easily lead to deadlock for me.

internal/internal_task_handlers.go
shijiesheng: good catch. I changed the order so it makes more sense.

internal/internal_worker_base.go
shijiesheng: I've removed this to use autoConfigHintAwareTask

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: fixed

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: added

internal/worker/concurrency_auto_scaler.go
shijiesheng: The poller wait time is an invariant. User doesn't need to tune it. The sensitivity control (time-to-react) is actually controlled by the Cooldown which is already in the parameter

shijiesheng:
coverage failed due to deprecation changes

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+	autoScalerEventPollerSkipUpdateCooldown   autoScalerEvent = "poller-limit-skip-update-cooldown"
+	autoScalerEventPollerSkipUpdateNoChange   autoScalerEvent = "poller-limit-skip-update-no-change"
+	autoScalerEventPollerSkipUpdateNotEnabled autoScalerEvent = "poller-limit-skip-update-not-enabled"
+	autoScalerEventEmitMetrics                autoScalerEvent = "auto-scaler-emit-metrics"
+	autoScalerEventEnable                     autoScalerEvent = "auto-scaler-enable"
+	autoScalerEventDisable                    autoScalerEvent = "auto-scaler-disable"
+	autoScalerEventStart                      autoScalerEvent = "auto-scaler-start"
+	autoScalerEventStop                       autoScalerEvent = "auto-scaler-stop"
+	autoScalerEventLogMsg                     string          = "concurrency auto scaler event"
+	testTimeFormat                            string          = "15:04:05"
```
Groxx:
```suggestion
```

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,390 @@
...
+				{autoScalerEventStop, true, 80, "00:00:03"},
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			defer goleak.VerifyNone(t)
+			core, obs := observer.New(zap.DebugLevel)
+			logger := zap.New(core)
+			clock := clockwork.NewFakeClockAt(time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC))
+			scaler := createTestConcurrencyAutoScaler(t, logger, clock)
```
Groxx, line 269:
```suggestion
			scaler := createTestConcurrencyAutoScaler(t, logger, clock)
			testTimeFormat := time.TimeOnly
```
since it's only used here / is meaningful here anyway

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+func NewConcurrencyAutoScaler(input ConcurrencyAutoScalerInput) *ConcurrencyAutoScaler {
+	tick := defaultAutoScalerUpdateTick
+	if input.Tick != 0 {
+		tick = input.Tick
+	}
+	return &ConcurrencyAutoScaler{
+		shutdownChan:             make(chan struct{}),
+		concurrency:              input.Concurrency,
+		cooldown:                 input.Cooldown,
+		log:                      input.Logger.Named(metrics.ConcurrencyAutoScalerScope),
```
Groxx: I think this might be our first use of `Named` 🤔 

since this isn't a concept in `log/slog` I kinda feel like we might drop it eventually, but for now I think it makes sense 👍

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+
+	c.wg.Add(1)
+
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				c.log.Error("panic in concurrency auto scaler, stopping the auto scaler", zap.Any("error", r))
+			}
+		}()
+		defer c.wg.Done()
```
Groxx:
```suggestion
		defer c.wg.Done()
		defer func() {
			if r := recover(); r != nil {
				c.log.Error("panic in concurrency auto scaler, stopping the auto scaler", zap.Any("error", r))
			}
		}()
```
since otherwise the log happens after we think it's shut down, and that can cause log-races for testing purposes.

(minor, but probably worth it)

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+	}()
+}
+
+func (c *ConcurrencyAutoScaler) Stop() {
+	if c == nil {
+		return // no-op if auto scaler is not set
+	}
+	close(c.shutdownChan)
+	c.wg.Wait()
+	c.logEvent(autoScalerEventStop)
```
Groxx:
```suggestion
	c.lock.Lock()
	c.logEvent(autoScalerEventStop)
	c.lock.Unlock()
	close(c.shutdownChan)
	c.wg.Wait()
```
maybe?  so you can see when it was requested, rather than when complete.  otherwise e.g. a stuck shutdown might never log, and that can lead to questions of "did this get called at all".

also mutex because otherwise there's a race between `ProcessPollerHint` and `logEvent`.

the log in `Start` is probably fine since "autoscaler must be started before polls start, because pollers depend on autoscaler" seems like a reasonable expectation.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+func (c *ConcurrencyAutoScaler) ProcessPollerHint(hint *shared.AutoConfigHint) {
+	if c == nil {
+		return // no-op if auto scaler is not set
+	}
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if hint == nil {
+		return
+	}
```
Groxx:
```suggestion
	if hint == nil {
		return
	}

	c.lock.Lock()
	defer c.lock.Unlock()
```

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+
+// Add always add positive numbers
+func (r *rollingAverage[T]) Add(value T) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	// no op on zero rolling window
+	if len(r.window) == 0 {
+		return
+	}
```
Groxx:
```suggestion
	// no op on zero rolling window
	if len(r.window) == 0 {
		return
	}

	r.mu.Lock()
	defer r.mu.Unlock()
```
minor/optional but since the slice header is never changed, this is safe to do without locking.

this won't be a major contention point or anything though, so I think it's entirely fine to be more defensive / lock everything if you have any preference for it.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+		c.pollerPermitLastUpdate = updateTime
+		c.logEvent(autoScalerEventPollerScaleUp)
+	} else if pollerWaitTime > c.pollerWaitTimeUpperBound { // pollers are idle
+		newQuota = c.scaleDownPollerPermit(pollerWaitTime)
+		c.concurrency.PollerPermit.SetQuota(newQuota)
+		c.pollerPermitLastUpdate = updateTime
+		c.logEvent(autoScalerEventPollerScaleDown)
+	} else {
+		c.logEvent(autoScalerEventPollerSkipUpdateNoChange)
+		return
```
Groxx:
```suggestion
```
sorta odd that this is the only explicit return branch otherwise

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+func (c *ConcurrencyAutoScaler) logEvent(event autoScalerEvent) {
+	if c.enabled {
+		c.scope.Counter(metricsEnabled).Inc(1)
+	} else {
+		c.scope.Counter(metricsDisabled).Inc(1)
+	}
+	c.scope.Histogram(metricsPollerQuota, metricsPollerQuotaBuckets).RecordValue(float64(c.concurrency.PollerPermit.Quota()))
+	c.scope.Histogram(metricsPollerWaitTime, metricsPollerWaitTimeBuckets).RecordDuration(c.pollerWaitTime.Average())
+	c.log.Debug(autoScalerEventLogMsg,
+		zap.Time("time", c.clock.Now()),
```
Groxx: not sure how I feel about this one tbh... 

zap has [an EncoderConfig field](https://pkg.go.dev/go.uber.org/zap@v1.27.0/zapcore#EncoderConfig) for this, and a [way to set a different Clock source](https://pkg.go.dev/go.uber.org/zap#WithClock) (TIL), probably better to use those.  and that should cover it for test purposes too.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+// smoothingFunc is a log2 function with offset to smooth the scaling and address 0 values
+func smoothingFunc(x time.Duration) float64 {
+	return math.Log2(2 + float64(x/time.Millisecond))
+}
+
+type number interface {
+	int64 | float64 | time.Duration
+}
+
+type rollingAverage[T number] struct {
```
Groxx, line 291: maybe worth making it non-generic, since it's only used with `time.Duration`?  or do we expect other uses, since it is pretty general?

I think "number-like" is plenty defensible and this is quite simple so no complaints either way.

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// inverse scaling with edge case of 0 wait time
+	// use logrithm to smooth the scaling to avoid drastic change
+	newQuota := math.Round(
+		float64(currentQuota) * smoothingFunc(c.pollerWaitTimeLowerBound) / smoothingFunc(pollerWaitTime))
+	newQuota = math.Max(
+		float64(c.pollerMinCount),
+		math.Min(float64(c.pollerMaxCount), newQuota),
+	)
+	return int(newQuota)
```
Groxx, line 265: (not specifically in this location, but just to choose one...)

there is a general assumption woven through ~all of this code that we never encounter negative numbers, since that can lead to NaNs in a couple places.

maybe worth enforcing that somewhere?  e.g. `rollingAverage.Add` seems like an easy target since that'd protect everything getting `Average`.
it'd be clear misbehavior from the server to send that negative hint back, and _correct_ Go usage should guarantee non-negative durations are measured everywhere client-side, but it's probably worth being a bit paranoid since it'll be really hard to figure out if it does occur.

---

e.g. concretely, if the average becomes like -3ms, `math.Log2` will produce `NaN` -> `math.Max(Min(...))` becomes `NaN` -> `int(NaN)` becomes `-9223372036854775808`.  yay floats!

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+		ticker := c.clock.NewTicker(c.updateTick)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-c.shutdownChan:
+				return
+			case <-ticker.Chan():
+				c.lock.Lock()
+				c.logEvent(autoScalerEventEmitMetrics)
+				c.updatePollerPermit()
```
Groxx, line 153: I kinda get the intent here (log "emit metrics" in debug mode, so the cause is clear), but it's sorta odd that outside debug mode this just means we rapidly double-emit since `updatePollerPermit` always emits metrics and a log too.

probably worth dropping the log here since the only calls to `updatePollerPermit` are due to this ticker?

internal/worker/concurrency_auto_scaler.go
```diff
@@ -0,0 +1,332 @@
...
+		pollerMaxCount           int
+		pollerMinCount           int
+		pollerWaitTime           *rollingAverage[time.Duration]
+		pollerPermitLastUpdate   time.Time
+		pollerWaitTimeUpperBound time.Duration
+		pollerWaitTimeLowerBound time.Duration
+	}
+
+	ConcurrencyAutoScalerInput struct {
+		Concurrency              *ConcurrencyLimit
```
Groxx, line 90: do we need the full `ConcurrencyLimit` here, or would it make more sense to pass just the poller's `Permit`?

I can see task-based autoscaling info being useful eventually, so it's probably fine if that's planned, it's just not currently used.

internal/internal_worker.go
```diff
@@ -1316,6 +1289,25 @@ func AugmentWorkerOptions(options WorkerOptions) WorkerOptions {
 	return options
 }
 
+func augmentAutoScalerOptions(options AutoScalerOptions) AutoScalerOptions {
+	if options.PollerMinCount <= 1 {
+		options.PollerMinCount = defaultMinPollerSize
+	}
```
Groxx, line 1300:
```suggestion
	// need at least 2 for sticky task polling to work
	if options.PollerMinCount <= 1 {
		options.PollerMinCount = defaultMinPollerSize
	}
```
I think that's the reason at least?  or is there another?

internal/internal_worker_base.go
```diff
@@ -167,34 +168,42 @@ func createPollRetryPolicy() backoff.RetryPolicy {
 
 func newBaseWorker(options baseWorkerOptions, logger *zap.Logger, metricsScope tally.Scope, sessionTokenBucket *sessionTokenBucket) *baseWorker {
 	ctx, cancel := context.WithCancel(context.Background())
+	logger = logger.With(zapcore.Field{Key: tagWorkerType, Type: zapcore.StringType, String: options.workerType})
```
Groxx:
```suggestion
	logger = logger.With(zap.String(tagWorkerType, options.workerType))
```
odd that it was like this before, but eh.  might as well clean up a bit.

internal/internal_worker_test.go
```diff
@@ -1498,3 +1507,55 @@ func TestTestValidateFnFormat_Workflow(t *testing.T) {
...
+			"empty activity task",
+			&activityTask{
+				autoConfigHint: &hint},
+			&hint,
+		},
+		{
+			"localactivity task",
+			&localActivityTask{},
+			nil,
+		},
```
Groxx, line 1555: maybe just replace the one nil-test with a
```go
_, ok := &localActivityTask{}.(autoConfigHintAwareTask)
assert.False(t, ok)
```
since I that's what is actually being checked in that case, and then the others all just become an obvious "must equal tt.want"

internal/worker.go
```diff
@@ -88,6 +88,7 @@ type (
 		// rate at which the worker is able to consume tasks from a task list,
 		// unless FeatureFlags.PollerAutoScalerEnabled is set to true.
 		// Default value is 1
+		// Deprecated: Use AutoScalerOptions instead.
```
Groxx: probably worth calling out that these are also completely ignored and do nothing, not just the ones below.

and maybe [something] on the max value, since it's now just "the poller count" if not using the autoscaler?

internal/worker.go
```diff
@@ -110,23 +111,30 @@ type (
...
+		// Deprecated: not used any more
 		PollerAutoScalerTargetUtilization float64
 
 		// optional: Sets whether to start dry run mode of autoscaler.
 		// Default value is false
+		// Deprecated: not used any more
 		PollerAutoScalerDryRun bool
 
+		// optional: Sets the options for poller autoscaler
+		AutoScalerOptions AutoScalerOptions
```
Groxx, line 250: I'm kinda curious if we'll want to split this into decision/activity options eventually..... but seems probably fine for now anyway.  and simpler to config if they're shared.

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+			var actualEvents []eventLog
+			for _, event := range obs.FilterMessage(autoScalerEventLogMsg).All() {
+				if event.ContextMap()["event"] != string(autoScalerEventEmitMetrics) {
+					t.Log("event: ", event.ContextMap())
+					actualEvents = append(actualEvents, eventLog{
+						eventType:   autoScalerEvent(event.ContextMap()["event"].(string)),
+						enabled:     event.ContextMap()["enabled"].(bool),
+						pollerQuota: event.ContextMap()["poller_quota"].(int64),
+						time:        event.ContextMap()["time"].(time.Time).Format(testTimeFormat),
```
Groxx:
```suggestion
						time:        event.Time.Format(testTimeFormat),
```
with a `logger.WithClock`, I think this handles the "logs are hard to identify uniquely" thing that seems to be the intent here?

Groxx (commented):
dropping notes for now, while reading tests carefully 👍 

overall looks pretty good I think - fairly easy to follow, behavior looks good (e.g. up to 4x growth when "instant", 0.5x shrink when slow, one scale change every 10 seconds sounds reasonable), everything's pretty close.  
so just a small pile of minor stuff, some nits some not.

internal/worker/concurrency_auto_scaler.go
shijiesheng: Just curious what should be the alternative to do it. I'll change once we want to remove it.

internal/worker.go
shijiesheng: My thought was to simplify things. I don't think it needs tuning in the first place so a single config should be good enough

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: Yes, I just didn't find a good way to assert event logs and find this is easier (or quicker) with a test entity.

internal/internal_worker.go
shijiesheng: Right. That's correct

internal/internal_worker_test.go
shijiesheng: fixed

internal/internal_worker_base.go
shijiesheng: fixed

internal/worker/concurrency_auto_scaler.go
shijiesheng: good catch. Thank you

internal/worker/concurrency_auto_scaler.go
shijiesheng: good catch

internal/worker.go
shijiesheng: For max, I've changed the logic and updated the comment. AutoscalerOptions will override Top level setting instead.

internal/worker/concurrency_auto_scaler.go
Groxx: the _main_ alternative is probably either:
- `logger.WithGroup`: https://pkg.go.dev/log/slog#Logger.WithGroup
  - i.e. pushing everything from "this logger" into a sub-field
- or `logger.With("logger", "concurrency-auto-scaler")`
  - setting a top-level field ("logger") and leaving everything else flat / possibly conflicting in meaning.

which is not a clear win in either direction.  flatter is much easier to query for shared fields, structured is much easier to be unambiguous and is often more efficient to index, etc.

I bring it up mostly because I think a move to `log/slog` is inevitable, and we'll have to decide [something] at that point.  It'll probably just be `.With("logger", "concurrency-auto-scaler")` tho, since I don't think we'll care much about dot.separated.names at that point (and none exist now).

internal/worker/concurrency_auto_scaler_test.go
Groxx: yea, the "make a simplified struct for comparison" is a good choice I think, this was just for the time-context-map-field.

Groxx: Huh.  `zap.WithClock` is unusable, given its definition 🤔 

still seems like an odd addition here tbh

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+				{common.PtrOf(true), common.PtrOf(int64(0))}, // <- tick, in cool down
+				{common.PtrOf(true), common.PtrOf(int64(0))}, // <- tick, scale down to minimum
+			},
+			[]eventLog{
+				{autoScalerEventStart, false, 100, "00:00:00"},
+				{autoScalerEventEnable, true, 100, "00:00:00"},
+				{autoScalerEventPollerSkipUpdateCooldown, true, 100, "00:00:01"},
+				{autoScalerEventPollerScaleUp, true, 200, "00:00:02"},
+				{autoScalerEventStop, true, 200, "00:00:02"},
+			},
```
Groxx: might be easier to follow the actual behavior of this one with a less-than-1/2-maximum set of values, e.g. start with 10 rather than 100 -> it won't scale to maximum, it'll scale to 42.

kinda similar for others below, e.g. `pollers, scale up and down multiple times` becomes:
```
{autoScalerEventStart, false, 10, "00:00:00"},
{autoScalerEventEnable, true, 10, "00:00:00"},
{autoScalerEventPollerSkipUpdateCooldown, true, 10, "00:00:01"},
{autoScalerEventPollerScaleUp, true, 42, "00:00:02"},
{autoScalerEventPollerSkipUpdateCooldown, true, 42, "00:00:03"},
{autoScalerEventPollerScaleDown, true, 25, "00:00:04"},
{autoScalerEventPollerSkipUpdateCooldown, true, 25, "00:00:05"},
{autoScalerEventPollerScaleUp, true, 104, "00:00:06"},
{autoScalerEventPollerSkipUpdateCooldown, true, 104, "00:00:07"},
{autoScalerEventPollerScaleDown, true, 63, "00:00:08"},
{autoScalerEventStop, true, 63, "00:00:08"},
```
which seems a bit more informative than "to max, down, back to max, back to same down value"

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+				{autoScalerEventPollerSkipUpdateCooldown, true, 100, "00:00:01"},
+				{autoScalerEventPollerScaleUp, true, 116, "00:00:02"},
+				{autoScalerEventStop, true, 116, "00:00:02"},
+			},
+		},
+		{
+			"busy pollers, scale up to maximum",
+			[]*shared.AutoConfigHint{
+				{common.PtrOf(true), common.PtrOf(int64(0))}, // <- tick, in cool down
+				{common.PtrOf(true), common.PtrOf(int64(0))}, // <- tick, scale down to minimum
```
Groxx:
```suggestion
				{common.PtrOf(true), common.PtrOf(int64(0))}, // <- tick, scale up significantly
```

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+				{autoScalerEventPollerSkipUpdateCooldown, true, 200, "00:00:07"},
+				{autoScalerEventPollerScaleDown, true, 121, "00:00:08"},
+				{autoScalerEventStop, true, 121, "00:00:08"},
+			},
+		},
+		{
+			"idle pollers waiting for tasks",
+			[]*shared.AutoConfigHint{
+				{common.PtrOf(true), common.PtrOf(int64(1000))}, // <- tick, in cool down
+				{common.PtrOf(true), common.PtrOf(int64(1000))}, // <- tick, scale up
```
Groxx:
```suggestion
				{common.PtrOf(true), common.PtrOf(int64(1000))}, // <- tick, scale down
```

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+				{autoScalerEventPollerSkipUpdateCooldown, true, 100, "00:00:01"},
+				{autoScalerEventPollerScaleDown, true, 80, "00:00:02"},
+				{autoScalerEventStop, true, 80, "00:00:02"},
+			},
+		},
+		{
+			"idle pollers, scale down to minimum",
+			[]*shared.AutoConfigHint{
+				{common.PtrOf(true), common.PtrOf(int64(60000))}, // <- tick, in cool down
+				{common.PtrOf(true), common.PtrOf(int64(60000))}, // <- tick, scale up
```
Groxx:
```suggestion
				{common.PtrOf(true), common.PtrOf(int64(60000))}, // <- tick, scale down
```

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+				{autoScalerEventPollerSkipUpdateCooldown, true, 100, "00:00:01"},
+				{autoScalerEventPollerScaleDown, true, 50, "00:00:02"},
+				{autoScalerEventStop, true, 50, "00:00:02"},
+			},
+		},
+		{
+			"idle pollers but disabled scaling",
+			[]*shared.AutoConfigHint{
+				{common.PtrOf(false), common.PtrOf(int64(100))}, // <- tick, in cool down
+				{common.PtrOf(false), common.PtrOf(int64(100))}, // <- tick, scale up
```
Groxx:
```suggestion
				{common.PtrOf(false), common.PtrOf(int64(100))}, // <- tick, no update
```

also this one isn't really "idle", that'd be ~60k / something larger than the scale-down value, yea?

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+				{autoScalerEventPollerSkipUpdateNotEnabled, false, 100, "00:00:01"},
+				{autoScalerEventPollerSkipUpdateNotEnabled, false, 100, "00:00:02"},
+				{autoScalerEventStop, false, 100, "00:00:02"},
+			},
+		},
+		{
+			"idle pollers but disabled scaling at a later time",
+			[]*shared.AutoConfigHint{
+				{common.PtrOf(true), common.PtrOf(int64(1000))},  // <- tick, in cool down
+				{common.PtrOf(true), common.PtrOf(int64(1000))},  // <- tick, scale up
```
Groxx:
```suggestion
				{common.PtrOf(true), common.PtrOf(int64(1000))},  // <- tick, scale down
```

internal/worker/concurrency_auto_scaler_test.go
Groxx: a possible option that appears to work: sort the logs by `event.Time`, and just make sure the other values occur in the same order.  you can get rid of the time field entirely from `eventLog` then.

(afaict they are always in order in these tests, but an explicit sort is probably a good idea)

there is technically room for some out-of-order-ness to occur by doing that, but I don't think these tests really run that risk.  and ensuring unique values in all logs would take care of it too.

internal/worker/concurrency_auto_scaler_test.go
```diff
@@ -0,0 +1,392 @@
...
+func createTestConcurrencyAutoScaler(t *testing.T, logger *zap.Logger, clock clockwork.Clock) *ConcurrencyAutoScaler {
+	return NewConcurrencyAutoScaler(ConcurrencyAutoScalerInput{
+		Concurrency: &ConcurrencyLimit{
+			PollerPermit: NewResizablePermit(100),
+			TaskPermit:   NewResizablePermit(1000),
+		},
+		Cooldown:                 2 * testTickTime,
+		Tick:                     testTickTime,
+		PollerMaxCount:           200,
+		PollerMinCount:           50,
```
Groxx, line 55: somewhat odd that `PollerPermit` starts at a different value than `PollerMinCount`, since I don't believe that'll ever be the case in practice?

it does seem harmless though, and kinda simplifies the "scale down" tests... just not sure that's worth breaking the normal pattern to achieve.

internal/worker/concurrency_auto_scaler.go
Groxx: correction: zap.WithClock needs an interface that provides a real `*time.Ticker` and that makes it incompatible with pretty much every possible mock clock (because a real Ticker cannot be controlled).

so lets skip that part.

Groxx (approved):
Just minor stuff as optional cleanups, I think - looks good to go to me 👍

internal/worker/concurrency_auto_scaler.go
shijiesheng: Yes, we need them all. This should also scale other Permit in the future.

internal/worker/concurrency_auto_scaler.go
shijiesheng: I ended up with using `With`

internal/worker/concurrency_auto_scaler.go
shijiesheng: Yes it's already in Debug level

internal/worker/concurrency_auto_scaler.go
shijiesheng: done

internal/worker/concurrency_auto_scaler.go
shijiesheng: good catch

internal/worker/concurrency_auto_scaler.go
shijiesheng: good catch. I've added the check on `ProcessHint()`

internal/worker/concurrency_auto_scaler.go
shijiesheng: make sense. Fixed

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: In practice, we observe the `MaxConcurrentPollers` set in the top level and use that as the initial poller count.

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: I see. I'll remove the time field then.

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: I'll keep the maximum case but will change the test cases below

internal/worker/concurrency_auto_scaler_test.go
shijiesheng: right, test case was wrong. I've fixed it

PR #1414, opened 2024-12-26
Fix retry policy attempts to match server behavior

Groxx:
# What this changes

Apparently we've had an off-by-one error in the test suite for years now, but few noticed.

Due to this mistake, retry policy config like this:
```go
&workflow.RetryPolicy{
  // ...
  MaximumAttempts: 3,
}
```
would result in up to _four_ calls in tests, but only _three_ in production.

The server has always considered "MaximumAttempts" to be "total number of calls" (going back to at least 2018 iirc, when I checked), so the test suite is simply wrong.
https://github.com/cadence-workflow/cadence/blob/830974dd9e49365bd974d288b21e949c48089216/service/history/execution/retry.go#L47-L51

# What this means for users whose tests are now failing

Unfortunately, this means your tests have been misleading you.
Production is unchanged, your tests are just failing because they've been wrong all along, and you'll need to update them.

Sorry about that 🤦

This is especially true when `MaximumAttempts: 1` is configured: **this actually means _zero retries_**, not _one retry_.
Ideally we would disallow `MaximumAttempts: 1` because it's so misleading... but this would affect production, so we've been holding off on it.  It may change in a later version with more communication though.

---

making the CI job happy:

**Detailed Description**
bugfix, covered above

**Impact Analysis**
- only affects tests
- no cross-version compatibility at all, it's intentionally "breaking" (at runtime)

**Testing Plan**
- **Unit Tests**: all existing tests
- **Persistence Tests**: n/a
- **Integration Tests**: n/a
- **Compatibility Tests**: n/a

**Rollout Plan**
- autobots, which are self-rolling-out.


dkrotx (approved):

pbacsko:
Hey folks, we started to hit this issue in our environment - luckily, just tests are affected. Any plans to merge this fix?

cc @Groxx @dkrotx

PR #1416, opened 2025-01-16
Populate tasks in internal workflowTask and activityTask entities for empty polls

shijiesheng:
**What changed?**
ensure task field is populated in workflowTask and activityTask for empty polls.

**Why?**

Task response even with empty task contains information like AutoConfigHint, which is needed for poller auto scaler

**How did you test it?**

Unit Test

**Potential risks**

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -744,7 +744,7 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 	workflowTask *workflowTask,
 	heartbeatFunc decisionHeartbeatFunc,
 ) (completeRequest interface{}, errRet error) {
-	if workflowTask == nil || isEmptyDecisionTask(workflowTask.task) {
+	if workflowTask == nil || workflowTask.task == nil {
 		return nil, errors.New("nil workflow task provided")
 	}
 	task := workflowTask.task
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -744,7 +744,7 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 	workflowTask *workflowTask,
 	heartbeatFunc decisionHeartbeatFunc,
 ) (completeRequest interface{}, errRet error) {
-	if workflowTask == nil || workflowTask.task == nil {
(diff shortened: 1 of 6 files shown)


internal/internal_task_handlers.go
```diff
@@ -75,15 +75,17 @@ type (
 	// workflowTask wraps a decision task.
 	workflowTask struct {
 		task            *s.PollForDecisionTaskResponse
+		autoConfigHint  *s.AutoConfigHint
```
neil-xie, line 78: why can not we reuse the autoConfigHint inside task(PollForDecisionTaskResponse) here? I'm assuming task != nil will mess up the logic somewhere else

shijiesheng: I haven't tried. But the field is widely used in many different places. Breaking the Nilness assumption might mess it up.

taylanisikdemir: I don't think we need the new field. It's a redundant information and requires each instantiation of workflowTask/activityTask to fill it based on the response. 
Instead expose a helper function such as:
```
func (wt *workflowTask) GetAutoConfigHint() *s.AutoConfigHint {
   if wt.task == nil {
     return nil
  }
  return wt.task.GetAutoConfigHint()
```

Less code to write and maintain.

shijiesheng: yeah, I checked carefully. We can populate this task field for empty polls. Thus I've changed the title as well.

internal/internal_task_pollers_test.go
```diff
@@ -61,6 +61,104 @@ func Test_newWorkflowTaskPoller(t *testing.T) {
...
+			},
+		}
+		emptyTask := &s.PollForDecisionTaskResponse{
+			TaskToken: nil,
+			AutoConfigHint: &s.AutoConfigHint{
+				common.PtrOf(true),
+				common.PtrOf(int64(1000)),
+			},
+		}
+		for _, tt := range []struct {
```
taylanisikdemir, line 80: let's also add nil cases

taylanisikdemir (approved):

internal/internal_task_pollers_test.go
shijiesheng: It's actually not possible for client to return `nil, nil` for both response and error so I'll skip the test

PR #1418, opened 2025-01-21
Add possibility to remove the RetryPolicy from the workflow context without overriding the other existing ActivityOptions in it

kittinanasi:
The WithRetryPolicy function accepts a pointer receiver RetryPolicy parameter, so it can be used to remove the RetryPolicy from the ActivityOptions.

I set the ActivityOptions including a non-nil RetryOptions configuration at the beginning of my workflow, these ActivityOptions are used by all activities in my workflow by default. I would like to override the RetryPolicy to nil for just one activity in that workflow without overriding the other ActivityOptions configurations.

I added a unit test.

When users set a retryPolicy with the WithRetryPolicy function, it could be broken, so the workflow might not be retried even if the RetryPolicy is set.

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1839,9 +1839,9 @@ func WithWaitForCancellation(ctx Context, wait bool) Context {
 }
 
 // WithRetryPolicy adds retry policy to the copy of the context
-func WithRetryPolicy(ctx Context, retryPolicy RetryPolicy) Context {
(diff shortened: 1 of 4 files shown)


shijiesheng:
This is a breaking change. It's better to add `WithoutRetryPolicy` to actually remove it. Also passing a nil pointer as function input is bad practice.

PR #1421, opened 2025-05-05
Upgrade internal/tools go module for x/tools and mockery

Groxx:
Mockery is failing when run with go 1.24 due to https://github.com/vektra/mockery/issues/914 , so let's upgrade it.

Basic steps to upgrade are:
```
# add go toolchain to go.mod to enforce a higher minimum
cd internal/tools
go get golang.org/x/tools@latest github.com/vektra/mockery/v2@latest
cd ../..
make all
```
and some careful reading to verify the results.

And then some changes to address our CI needs, e.g. since runtime and tool-time versions have diverged (currently) we now automatically pull and build with the old version.
Most of these version values can be found by doing a grep like
```
❯ rg 'go(lang)?.?1.?2\d'
go.mod
3:go 1.21                # intentionally different

docker/buildkite/Dockerfile
1:FROM golang:1.24

Makefile
36:EXPECTED_GO_VERSION := go1.24

CHANGELOG.md
28:- Bump x/tools for tools, to support go 1.22 (#1336)
143:- Fix TestActivityWorkerStop: it times out with go 1.20 by @dkrotx in #1223

internal/tools/go.mod
3:go 1.23.0
5:toolchain go1.24.2
```

I had _expected_ `go 1.21` in go.mod to take care of this kind of check, but apparently not: https://github.com/golang/go/issues/73603
so it's now done by hand.

---

There are *some* changes which I think stand a very small chance of causing problems:
- Our RPC client's mocks now check for `.Get(0).(func(args) (any, error)` signatures where before they did not
  - I believe anyone using this would've hit a `.Error` further down after these changes, so hopefully this just fixes buggy behavior and does not cause any existing tests to fail
- Many funcs gained a `if len(ret) == 0 { panic(..) }` check
  - Since there's a `ret.Get(0)` immediately after which also panics, I suspect this just gives better error messages.
- Our mocks now have a `DoAndReturn` method
  - Since there are no interfaces for all this, this should be fine, ignoring reflection (which we must ignore or else we will go mad).

Otherwise seems like nothing exciting, and this does not affect any user-facing libraries or Go versions.
Since this needs a newer CI Go version to build, I've updated those references too.

---

Separately, this was the source of some rather disturbing TILs:
```bash
❯ bash -ec '
[[[ "$a" = "foo"]] && echo first
echo second
'
bash: line 2: [[[: command not found
second

❯ echo $?
0
```

Apparently syntax errors are (sometimes) not script failures, even with `set -e`.

And `sh` does not support `[[`, so it was _printing an error_ and continuing without erroring.  CI uses `sh`, locally we generally get bash or zsh, so a locally-working `[[` may be a remote-silently-failing command.

Madness.  Absolute madness.  This is going to haunt me for a while, unless I can find a way to prevent it (without shellcheck).


shijiesheng (approved):

taylanisikdemir (approved):

Makefile
```diff
@@ -278,6 +278,14 @@ build: $(BUILD)/fmt ## ensure all packages build
 	go build ./...
 	$Q # caution: some errors are reported on stdout for some reason
 	go test -exec true ./... >/dev/null
+	$Q # compare the current go version with what is in go.mod, and get `{gomod_version}.1` or similar and ensure it builds.
+	$Q GOVERSION="$(shell go env GOVERSION | cut -d. -f1-2)" \
+		MODVERSION="go$(shell grep '^go 1' go.mod | cut -d' ' -f2)"; \
+		if [ "$$GOVERSION" != "$$MODVERSION" ]; then \
```
Groxx, line 284: so this was the source of some rather disturbing TILs:

```bash
❯ bash -ec '
[[[ "$a" = "foo"]] && echo first
echo second
'
bash: line 2: [[[: command not found
second

❯ echo $?
0
```

Apparently syntax errors are (sometimes) not script failures, even with `set -e`.

And `sh` does not support `[[`, so it was _printing an error_ and continuing without erroring.  CI uses `sh`, locally we generally get bash or zsh.

Madness.  Absolute madness.  This is going to haunt me for a while, unless I can find a way to prevent it (without shellcheck).

internal/internal_workflow_task_handler_mock.go
```diff
@@ -41,7 +41,15 @@ func (_m *MockWorkflowTaskHandler) EXPECT() *MockWorkflowTaskHandler_Expecter {
...
 
+	if len(ret) == 0 {
+		panic("no return value specified for ProcessWorkflowTask")
+	}
+
 	var r0 interface{}
+	var r1 error
+	if rf, ok := ret.Get(0).(func(*workflowTask, decisionHeartbeatFunc) (interface{}, error)); ok {
+		return rf(task, f)
+	}
```
Groxx, line 52: Below there is this:
```go
if ret.Get(0) != nil {
	r0 = ret.Get(0).(interface{})
}
```
which should have panicked if the new signature was used, so I believe this is a safe addition / nobody could have been passing this value in previously.

internal/internal_workflow_task_handler_mock.go
```diff
@@ -84,13 +91,17 @@ func (_c *MockWorkflowTaskHandler_ProcessWorkflowTask_Call) Return(response inte
 	return _c
 }
 
-type mockConstructorTestingTNewMockWorkflowTaskHandler interface {
-	mock.TestingT
-	Cleanup(func())
+func (_c *MockWorkflowTaskHandler_ProcessWorkflowTask_Call) RunAndReturn(run func(*workflowTask, decisionHeartbeatFunc) (interface{}, error)) *MockWorkflowTaskHandler_ProcessWorkflowTask_Call {
```
Groxx, line 94: new mock method, should be fine as it's not on an interface

PR #1423, opened 2025-05-21
Add workflow.GetSpanContext and workflow.WithSpanContext API

shijiesheng:
**Detailed Description**

* GetSpanContext API is used for fetch span context from workflow if it exists
* WithSpanContext() API is used for overriding the existing span context in the workflow so it could be passed to activities and child workflows

**Impact Analysis**
- **Backward Compatibility**: Yes, they are new APIs
- **Forward Compatibility**: Yes, WithSpanContext would write the existing span context field and it's readable for old client versions

**Testing Plan**
- **Unit Tests**: No
- **Persistence Tests**: No
- **Integration Tests**: Yes
- **Compatibility Tests**: No

**Rollout Plan**
- What is the rollout plan? new release
- Does the order of deployment matter? no
- Is it safe to rollback? Does the order of rollback matter? Safe
- Is there a kill switch to mitigate the impact immediately? Rollback to previous version and remove usages of the new APIs

**Why?**

Users can now access baggage items in the workflow code.

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -494,6 +494,6 @@ func GetSpanContext(ctx Context) opentracing.SpanContext {
 	return nil
 }
 
-func contextWithSpan(ctx Context, spanContext opentracing.SpanContext) Context {
+func WithSpanContext(ctx Context, spanContext opentracing.SpanContext) Context {
 	return WithValue(ctx, activeSpanContextKey, spanContext)
 }
--- a/internal/context.go
+++ b/internal/context.go
@@ -486,7 +486,7 @@ func (c *valueCtx) Value(key interface{}) interface{} {
 	return c.Context.Value(key)
 }
 
-func spanFromContext(ctx Context) opentracing.SpanContext {
+func GetSpanContext(ctx Context) opentracing.SpanContext {
 	val := ctx.Value(activeSpanContextKey)
 	if sp, ok := val.(opentracing.SpanContext); ok {
 		return sp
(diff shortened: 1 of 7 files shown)


internal/context.go
```diff
@@ -486,7 +486,7 @@ func (c *valueCtx) Value(key interface{}) interface{} {
 	return c.Context.Value(key)
 }
 
-func spanFromContext(ctx Context) opentracing.SpanContext {
+func GetSpanContext(ctx Context) opentracing.SpanContext {
```
taylanisikdemir, line 489: should we also expose `contextWithSpan` or is there another way for users to add their existing context?

timl3136 (approved):

internal/context.go
shijiesheng: Now I'm a bit concerned about the determinism if users start a new span inside the workflow.

taylanisikdemir: I guess we don't have a clear story around replay traces. It wouldn't cause non-determinism though. Just weird/duplicate traces potentially. 
Are you suggesting to expose the span via `GetSpanContext` but don't let user set/override the span in the context?

shijiesheng: You are right. It's still deterministics. Replay is going to cause issues. We will see multiple child spans if we allow users to set them.

shijiesheng: Using Sideeffect would work. But is span serializable or recoverable?

shijiesheng: SpanContext is serializable and should be passed. With that said, we'll require users to start a short-lived new span just to modify spancontext. Please refer the examples in the comments

shijiesheng: I've added an integration test which works.

PR #1424, opened 2025-06-02
Revert "Populate tasks in internal workflowTask and activityTask entities"

shijiesheng:
**What changed?**

A pure revert of https://github.com/cadence-workflow/cadence-go-client/commit/d9dcddc97c7a87e93dc5f9ca1ec252a8ccf620a3#diff-caac3c5868c1fd134487cacc53e459b7375722e23c6aba3e9407521feba00007 

**Why?**

Nilness check is no longer enough after the [commit](https://github.com/cadence-workflow/cadence-go-client/commit/d9dcddc97c7a87e93dc5f9ca1ec252a8ccf620a3 )

One alternative is to add additional emptiness check. This makes the client code error prone. Thus, a pure revert is desired. 

**How did you test it?**

Unit Test

**Potential risks**


internal/internal_task_pollers.go
```diff
@@ -1148,7 +1148,7 @@ func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
 	}
 
 	activityTask := task.(*activityTask)
-	if activityTask.task == nil {
+	if activityTask.task == nil || len(activityTask.task.TaskToken) == 0 {
```
Groxx: do we need a similar change on workflow tasks?  since that path was changed too.

shijiesheng: WorkflowTask doesn't need change since it's already handling emptiness [here](https://github.com/shijiesheng/cadence-client/blob/c06b7bb191cd4bdc9b57307607e3f7b40855bc1d/internal/internal_task_handlers.go#L751-L756)

Groxx: 👍 thanks, I'm convinced :)  same kinds of paths as before then, should be safe.

taylanisikdemir (approved):

Groxx (approved):

PR #1425, opened 2025-06-05
Introduce CronOverlapPolicy in StartWorkflowOptions from cadence-idl

timl3136:
**Detailed Description**
Introduce CronOverlapPolicy in StartWorkflowOptions from cadence-idl.
Also create empty methods for DeleteDomain operation matching update from cadence-idl.

**Impact Analysis**
- Backward compatible since this is a new field added
- Forward compatible as this field will be default to not affect any current behavior.

**Testing Plan**
- Yes
- Persistence tests will be implemented in server repo
- will be implemented in server repo
- No yet

**Rollout Plan**
- What is the rollout plan?
- Rollout client change after server one landed
- Not really since server will drop the field if not recognized and default value won't affect current behavior
- Yes, order does not matter
- Rollback if needed

### DIFF
--- a/internal/compatibility/thrift/enum.go
+++ b/internal/compatibility/thrift/enum.go
@@ -391,5 +391,6 @@ func CronOverlapPolicy(t apiv1.CronOverlapPolicy) *shared.CronOverlapPolicy {
 	case apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_BUFFER_ONE:
 		return shared.CronOverlapPolicyBufferone.Ptr()
 	}
-	return shared.CronOverlapPolicySkipped.Ptr()
+	// we treat any unknown value as invalid
+	return nil
 }
--- a/internal/compatibility/enum_test.go
+++ b/internal/compatibility/enum_test.go
@@ -344,5 +344,5 @@ func TestCronOverlapPolicy(t *testing.T) {
 	} {
 		assert.Equal(t, v, proto.CronOverlapPolicy(thrift.CronOverlapPolicy(v)))
 	}
-	assert.Panics(t, func() { proto.CronOverlapPolicy(thrift.CronOverlapPolicy(999)) })
+	assert.Equal(t, apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_INVALID, proto.CronOverlapPolicy(thrift.CronOverlapPolicy(999)))
 }


.gen/go/shared/shared.go
taylanisikdemir, line 1: we need to update the corresponding types/mapper functions in common/types/shared.go.

timl3136: Thanks, I plan to land the corresponding change in server first before releasing this.

internal/compatibility/proto/enum.go
```diff
@@ -393,3 +393,16 @@ func ChildWorkflowExecutionFailedCause(t *shared.ChildWorkflowExecutionFailedCau
...
+	if t == nil {
+		return apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_INVALID
+	}
+	switch *t {
+	case shared.CronOverlapPolicySkipped:
+		return apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_SKIPPED
+	case shared.CronOverlapPolicyBufferone:
+		return apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_BUFFER_ONE
+	}
+	panic("unexpected enum value")
```
Shaddoll: This is a bad pattern and preventing us from introducing new enum values.

timl3136: Sorry but I am a bit confused here. I think we can also add a new enum values from the idl and sync that here.
Do you think we should return invalid in this case?

internal/client.go
```diff
@@ -471,6 +471,13 @@ type (
 		// This will only be used and override DelayStart and JitterStart if provided in the first run
 		// Optional: defaulted to Unix epoch time
 		FirstRunAt time.Time
+
+		// CronOverlapPolicy - Policy for handling cron workflow overlaps.
+		// Currently supported values are:
+		// - CronOverlapPolicySkip: skip the new execution if the previous one is still running
+		// - CronOverlapPolicyBufferOne: buffer one execution if the previous one is still running
```
taylanisikdemir: let's be more descriptive here. give an example scenario and also mention that jitter is still respected

internal/common/metrics/service_wrapper.go
```diff
@@ -444,3 +445,11 @@ func (w *workflowServiceMetricsWrapper) RestartWorkflowExecution(ctx context.Con
 	scope.handleError(err)
 	return resp, err
 }
+
+func (w *workflowServiceMetricsWrapper) DeleteDomain(ctx context.Context, request *shared.DeleteDomainRequest, opts ...yarpc.CallOption) error {
+	//TODO implement me
```
taylanisikdemir: what else is left to implement here?

internal/compatibility/adapter.go
```diff
@@ -268,6 +268,11 @@ func (a thrift2protoAdapter) RestartWorkflowExecution(ctx context.Context, reque
 	return thrift.RestartWorkflowExecutionResponse(response), proto.Error(err)
 }
 
+func (a thrift2protoAdapter) DeleteDomain(ctx context.Context, DeleteRequest *shared.DeleteDomainRequest, opts ...yarpc.CallOption) error {
+	//TODO implement me
+	return nil
```
taylanisikdemir: better to return an unimplemented error

internal/compatibility/proto/enum.go
taylanisikdemir: +1 to the bad pattern comment. These unknown enum panics make it impossible to add new enum values until all the clients are updated. Better to return invalid enum here which will be interpreted as skip (default), right?

internal/compatibility/thrift/enum.go
```diff
@@ -381,3 +381,15 @@ func ChildWorkflowExecutionFailedCause(t apiv1.ChildWorkflowExecutionFailedCause
...
+func CronOverlapPolicy(t apiv1.CronOverlapPolicy) *shared.CronOverlapPolicy {
+	switch t {
+	case apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_INVALID:
+		return nil
+	case apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_SKIPPED:
+		return shared.CronOverlapPolicySkipped.Ptr()
+	case apiv1.CronOverlapPolicy_CRON_OVERLAP_POLICY_BUFFER_ONE:
+		return shared.CronOverlapPolicyBufferone.Ptr()
+	}
+	panic("unexpected enum value")
```
taylanisikdemir: same here

taylanisikdemir (commented):
Did we test this e2e with and without setting the policy?

internal/common/metrics/service_wrapper.go
timl3136: Thank you, I don't think there are more needed to be implement here. I will remove the TODO here.

internal/compatibility/adapter.go
timl3136: Thank you, I have implemented this adapter method instead of return nil.

internal/compatibility/proto/enum.go
timl3136: Thanks, I will change that to return SKIP if it's unexpected enum

internal/compatibility/thrift/enum.go
timl3136: Thanks, I will change that to return SKIP if it's unexpected enum

internal/client.go
timl3136: Thanks, I have added an example scenario.

timl3136:
> Did we test this e2e with and without setting the policy?

I have tested this locally, I plan to add an integration test later.

PR #1426, opened 2025-06-11
add NewBatchFuture API

shijiesheng:
**Why?**

Fanout pattern is common in user workflows. However, it's only possible to fanout <100 futures concurrently. Anything more than that would cause history service to stuck processing. 

This BatchFuture caps the number of concurrent futures and thus promotes the correct usage. 

**How did you test it?**

Unit Test in test environment

**Detailed Description**
* Added `NewBatchFuture` API for batch use cases
* Added `BatchFuture` interface

**Impact Analysis**
- **Backward Compatibility**: yes
- **Forward Compatibility**: yes

**Testing Plan**
- **Unit Tests**: Yes
- **Persistence Tests**: Not related
- **Integration Tests**: No
- **Compatibility Tests**: No due to new API change only

**Rollout Plan**
- What is the rollout plan? deploy new version 


internal/executor/executor.go
```diff
@@ -0,0 +1,91 @@
...
+
+type Executor interface {
+	AddFuture(factory func(ctx internal.Context) internal.Future, valuePtr interface{}) error
+	Execute(ctx internal.Context) error
+}
+
+type batchExecutor struct {
+	isRunning bool
+	factories []func(ctx internal.Context) internal.Future
+	valuePtrs []interface{}
```
Groxx: to stick it in here and not just chat:
tbh I think I'd prefer to avoid trying to write out results.
trying to write to an out-var internally risks decoding errors that you can't retry, and you have to decide what to do with the error.

I'm not quite sure what the api _should_ be, but I think we might be better off just exposing a list of futures / giving access to futures.  then they can `.Get` on them like normal, and we could even keep both successful and error cases in the same list, so correlating indexes to success-or-error is easier.

internal/executor/executor.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+func (e *batchExecutor) Execute(ctx internal.Context) error {
+	if e.isRunning {
+		return errors.New("executor is already running")
+	}
+	e.isRunning = true
+
+	futuresToProcess := internal.NewNamedChannel(ctx, "futures")
```
Groxx: maybe add an optional(?) name arg, and use that here?  "futures" is a bit vague, and while I hope someone doesn't make a dozen of these, it might be nice to be able to tell two of them apart.

we could also use the name in `internal.GoNamed` below, to be a bit clearer about library-internal goroutines when someone collects a stacktrace.

internal/executor/executor.go
```diff
@@ -0,0 +1,91 @@
...
+	})
+
+	// submit all futures
+	for i := range e.factories {
+		buffered.Send(ctx, nil)
+		futuresToProcess.Send(ctx, futureWithResult{
+			future:   e.factories[i](ctx),
+			valuePtr: e.valuePtrs[i],
+		})
+	}
```
Groxx: since I'm not 100% sure from a quick read through, it's probably worth checking the behavior of cancellation in the middle of this loop very carefully.

I _think_ what it'll do now is just continue the loop, with a canceled context going to all future closures that are executed, so e.g. if they do the ExecuteActivity in the test it'll just be a silent noop that doesn't get submitted to the server.  

That seems.... probably fine?  But we should make sure it works, and document it.

internal/executor/executor.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"errors"
+
+	"go.uber.org/multierr"
+
+	"go.uber.org/cadence/internal"
+)
+
+type Executor interface {
+	AddFuture(factory func(ctx internal.Context) internal.Future, valuePtr interface{}) error
```
Groxx: as a general Q: a func that doesn't allow returning an error puts a pretty tight restriction on what this is allowed to do (safely).  it sorta forces things to only be the example, with "just return an ExecuteActivity future".

.... which might be fine, and it does fit most cases + lead to a simpler API.  But do you see any reasonable way to expand this to something that allows returning errors, and [does something] internally if that happens?  
I can see it probably being quite a bit more complex so maybe it's not worth it, but it seems worth an attempt at least.

---

though I guess people could just use a `f, s := NewFuture(ctx); s.SetError(err)` and return that.

internal/executor/executor.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type futureWithResult struct {
+	future   internal.Future
+	valuePtr interface{}
+}
+
+func (e *batchExecutor) AddFuture(factory func(ctx internal.Context) internal.Future, valuePtr interface{}) error {
+	if e.isRunning {
+		return errors.New("executor is already running")
```
Groxx: probably worth making this a specific error type, to differentiate it from closure-returned errors, since it implies pretty special behavior ("no changes").

internal/executor/executor.go
```diff
@@ -0,0 +1,91 @@
...
+	wg.Wait(ctx)
+
+	return errs
+}
+
+func NewBatchExecutor(ctx internal.Context, batchSize int) Executor {
+	return &batchExecutor{
+		batchSize: batchSize,
+	}
+}
```
Groxx: 🤔 if we keep the current API largely untouched, but change this to
```go
ExecuteBatch(ctx, size, closures...) (???, error)
```
and only returning the results, we could eliminate the risk of anyone adding-after-executing or executing 2x in a row...

there's some convenience / ergonomics to be had with a mutating wrapper, but I'm never sure how I feel about them for simpler patterns like this :\\
a returned object like it is now would be easier to extend with more features / more options later though.

Groxx (commented):
I think I need to think about it a bit more 🤔 

Broad feelings though are in inline comments, and:
- I think this is a good candidate for (finally) creating a `go.uber.org/cadence/x/` experimental folder, for early-stage things like this.  We can "promote" it to `/workflow` when we're sure we're fully happy with it, and have a couple user feedbacks.
- Overall makes sense and I see the use, we have a lot of problems with people doing stuff like this and it'll almost certainly help.
- The current setup where everything is enqueued synchronously, and then `.Execute(ctx)` runs the closures async, seems probably best 👍 
  - In particular it avoids making closures "maybe sync, maybe async" which is the source of a lot of logic bugs in my experience.
  - We should make sure this is documented very clearly tho.

So mostly I like it.  Might need a little tweaking / a couple discussion rounds with people, but it seems worth building.

internal/executor/executor.go
shijiesheng: On second thought, this won't work. We rely on futures.get() to ensure we can safely enqueue next one (thus `buffered.Receive(ctx, nil)` to release one token). Let users to access future meaning we need to keep track of future status periodically.

taylanisikdemir: This requires users to preallocate an array and they will only have access once the whole batch completes executing. 
No strong opinion but another alternative is to have user pass a callback to process result. Maybe they will just do some kind of map-reduce. No need to require them to allocate upfront and have access at the end. 
e.g. 
```

sum := 0
var errs multierr.Error
for _, row := range myData {
  executor.AddFuture(
    func(ctx internal.Context) internal.Future {
      aCtx := internal.WithActivityOptions(ctx, internal.ActivityOptions{
        ScheduleToStartTimeout: time.Second * 10,
        StartToCloseTimeout:    time.Second * 10,
      })
      return internal.ExecuteActivity(aCtx, batchActivity, row)
    },
    func(result any, err error) { // process result
       if err != nil {
         errs = multierr.Append(errs, err) 
         return
       }
       sum += any.(int)
    }
  )
}
```

internal/executor/executor.go
shijiesheng: Right, futures are actually chainable which allows me to return all futures at the beginning.

internal/executor/executor.go
Groxx: > On second thought, this won't work. We rely on futures.get() to ensure we can safely enqueue next one (thus `buffered.Receive(ctx, nil)` to release one token). Let users to access future meaning we need to keep track of future status periodically.

I don't think it'd change anything, you can `.Get` multiple times on futures - just use a `nil` out-var internally because the value doesn't matter.  They can `.Get(ctx, &usefulOutVar)` any time, totally unaffected by this other `.Get(ctx, nil)` call.

Or do you mean something else?

---

To be clear, I'm definitely not trying to claim "this should definitely return a list of futures", more of a "this might be more flexible, has it been considered".  It'd likely be more complicated to use and that might be a net-worse result.  I'm just a bit reluctant to _hide_ futures without a good reason, since they're a useful primitive in many cases.

internal/batch/batch_future.go
```diff
@@ -0,0 +1,149 @@
...
+	// ensure valuePtr is a slice
+	var sliceValue reflect.Value
+	if valuePtr != nil {
+
+		switch v := reflect.ValueOf(valuePtr); v.Kind() {
+		case reflect.Ptr:
+			if v.Elem().Kind() != reflect.Slice {
+				return fmt.Errorf("valuePtr must be a pointer to a slice, got %v", v)
+			}
+			sliceValue = v.Elem()
```
Groxx: might want to allow pointer-to-nil-slice to allocate a slice of the appropriate size.

also for both of these: you probably want `%T", valuePtr` instead of `%v", v` since the latter won't tell you the type, just the value.

there's also [something] we can/should do in here to make sure values are `.CanSet()`, before waiting - likely worth doing since that'll prevent a panic further down.

internal/batch/batch_future.go
```diff
@@ -0,0 +1,149 @@
...
+		defer b.wg.Done()
+
+		wgForFutures := internal.NewWaitGroup(ctx)
+
+		var idx int
+		for channel.Receive(ctx, &idx) {
+			idx := idx
+
+			wgForFutures.Add(1)
+			internal.GoNamed(ctx, "batch-future-processor-one-future", func(ctx internal.Context) {
```
Groxx: maybe worth `"batch-future-processor-future-"+strconv.Itoa(idx)`?

internal/batch/batch_future.go
```diff
@@ -0,0 +1,149 @@
...
+			errs = multierr.Append(errs, b.futures[i].Get(ctx, nil))
+		} else {
+			value := sliceValue.Index(i)
+			if value.Kind() != reflect.Ptr {
+				value = value.Addr()
+			}
+			// if value is nil, initialize it
+			if value.IsNil() {
+				value.Set(reflect.New(value.Type().Elem()))
+			}
```
Groxx: I _think_ you can skip this, since `.Get` does it internally.

and/or probably _should_ skip it, since if `.Get` doesn't, it'd mean this sets a non-nil value while a normal future would leave it nil.

shijiesheng: make sense. I've changed the behavior. Now it only accepts pointers to slice (similar to Get that only accept pointers). The behavior is much simpler now.
I'll simply resize the slice to match required length. (Also this is why pointer to slice is important here) This is also documented in the `Get` method comment.

internal/batch/batch_future.go
shijiesheng: fixed

internal/batch/batch_future_test.go
```diff
@@ -0,0 +1,334 @@
...
+	batchFuture, err := NewBatchFuture(ctx, concurrency, factories)
+	if err != nil {
+		return nil, err
+	}
+
+	var valuePtr []int
+	if err := batchFuture.Get(ctx, nil); err != nil {
+		return nil, err
+	}
+	return valuePtr, nil
```
Groxx, line 294: _probably_ better to return an explicit `nil` here, just to avoid implying typed nils behave in a certain way (they should not, unless they have a custom marshaler)

x/doc.go
```diff
@@ -0,0 +1,5 @@
+package x
+
+/*
+Package x is an experimental package for early-stage features. The API here is not stable and may change.
+*/
```
Groxx:
```suggestion
// Package x is an experimental package for early-stage features. The API here is not stable and may change.
package x
```
no actual preference for `//` vs `/*` (`/*` is kinda nice for big blobs of text tho), but it needs to be above the package definition to be a package-doc / to appear in godoc displays.

internal/workflow_testsuite.go
```diff
@@ -601,7 +601,12 @@ func (t *TestWorkflowEnvironment) GetWorkflowResult(valuePtr interface{}) error
 	if !t.impl.isTestCompleted {
 		panic("workflow is not completed")
 	}
-	if t.impl.testError != nil || t.impl.testResult == nil || t.impl.testResult.HasValue() == false || valuePtr == nil {
+	// fast lane for no result assignment
+	if t.impl.testResult == nil || t.impl.testResult.HasValue() == false || valuePtr == nil {
+		return t.impl.testError
+	}
+	if t.impl.testError != nil {
+		t.impl.testResult.Get(valuePtr)
```
Groxx: errors need to not return a result, since that's how we record it in prod
```suggestion
```
I think splitting these blocks of logic apart is fine, it's arguably doing too much at once, but that's a substantial change

internal/batch/batch_future.go
```diff
@@ -0,0 +1,142 @@
...
+
+	"go.uber.org/multierr"
+
+	"go.uber.org/cadence/internal"
+)
+
+type BatchFuture interface {
+	internal.Future
+	GetFutures() []internal.Future
+}
```
Groxx: it's probably fine either way, but it's a bit odd to have this both internal and public, and the internal one is essentially unused.

maybe just make `batchFutureImpl` public, and keep (and document) the public one?  it's in `internal/`, it can't be used or referenced externally, should be functionally the same as keeping it private here.

that said, I don't believe there's any _downside_ to this interface-duplication.  so I think it's fine to leave it like it is.

shijiesheng: Should I use alias then? It still doesn't feel right to expose batchFutureImpl directly.

internal/workflow_testsuite.go
shijiesheng: I'll revert this change.

x/batch.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"go.uber.org/cadence/internal/batch"
+	"go.uber.org/cadence/workflow"
+)
+
+// BatchFuture is a interface that extends workflow.Future and adds a method to get the futures created
+type BatchFuture interface {
+	workflow.Future
+	GetFutures() []workflow.Future
+}
```
Groxx:
```suggestion
// BatchFuture wraps a collection of futures, and provides some convenience methods for dealing with them in bulk.
type BatchFuture interface {
	// IsReady returns true when all wrapped futures return true from their IsReady
	IsReady() bool
	// Get acts like workflow.Future.Get, but it reads out all wrapped futures into the provided slice pointer.
	// You MUST provide a pointer to a slice as the value-pointer here, but the slice itself can be nil - it will be allocated and/or resized to fit if needed.
	//
	// This call will wait for all futures to resolve, and will then write all results to the output slice in the same order as the input.
	// 
	// Any errors encountered are merged with go.uber.org/multierr, so single errors are
	// exposed normally, but multiple ones are bundled in the same way as errors.Join.
	// For consistency when checking individual errors, consider using `multierr.Errors(err)` in all cases.
	Get(ctx internal.Context, valuePtr interface{}) error
	// GetFutures returns a slice of all the wrapped futures.
	// This slice MUST NOT be modified, but the individual futures can be used normally.
	GetFutures() []workflow.Future
}

var _ workflow.Future = (BatchFuture)(nil) // to ensure it's compatible
```
^ this is the main place we have to document stuff for users, so let's do that.

also, though embedding is convenient, I think it's worth expanding it so the differences from a normal future can be described.

internal/batch/batch_future_test.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	return result, err
+}
+
+func batchActivity(ctx context.Context, taskID int) (int, error) {
+	select {
+	case <-ctx.Done():
+		return taskID, fmt.Errorf("batch activity %d failed: %w", taskID, ctx.Err())
+	case <-time.After(time.Duration(rand.Int63n(100))*time.Millisecond + 900*time.Millisecond):
```
Groxx, line 82: probably worth making this a lot faster tbh, currently the tests take ~22s which seems a bit excessive.

the rest of the slowness in here is probably _mostly_ due to trying to measure wall-clock time instead of simulated time, but I think simulated time might be more relevant - there's `env.Now()` for measuring simulated time.  switching to that and e.g. mocking activities with `.After(rand*time.Second)` would make these essentially instant, but would still give you the ability to ensure "runs concurrently within batch bounds"

internal/batch/batch_future_test.go
```diff
@@ -0,0 +1,297 @@
...
+	go func() {
+		defer wg.Done()
+		env.ExecuteWorkflow(batchWorkflow, batchWorkflowInput{
+			Concurrency: concurrency,
+			TotalSize:   totalSize,
+		})
+	}()
+
+	time.Sleep(totalExpectedTime / 2)
+	env.CancelWorkflow()
```
Groxx: I'm a little surprised this works, but neat.

if you want to try fully simulating it, this is basically
```go
env.RegisterDelayedCallback(func() {
	env.CancelWorkflow()
}, totalExpectedTime / 2)
env.ExecuteWorkflow(...)
```

x/batch.go
shijiesheng: right. Also I added the comment for Nil use cases `Get(ctx, nil)`.

Groxx (commented):
some small bits of polish but otherwise I think it looks good.

tests... are kinda nice that it's "real" and demonstrates things working the way they should + cancellation timing + etc, but 22s is quite a lot to pay for that.  might be worth rewriting with fully simulated time, it should be pretty easy to achieve.

(and maybe keeping some small ones non-mocked to show that they both work, because it's a fairly good stress test for the test suite, and we don't have many of those.  but seems fine if that's too big for some reason)

---

So some minor polish and docs, and otherwise I think this is good to go.  Tests would be nice / interesting / a useful exercise to redo with simulated time, but 22s isn't fatally slow or anything.  They seem to do a good job showing that the behavior is correct, and that's by far the most important part, so we can do that later or something if you want.

internal/batch/batch_future.go
Groxx: (also IRL but) no, I'm just thinking:
- delete this interface, nothing uses it _as an interface_ (no other impls)
- make `batchFutureImpl` public so the linter doesn't complain / it is legitimately annoying to return a private type, and this is still in an internal package so it's still hidden from users
- keep the public interface, it's useful / matches our other future APIs even if we should perhaps prefer returning a `*BatchFuture` struct for API-stability reasons.  but I doubt this'll need to change, the interface is extremely straightforward.

shijiesheng: Sure. Sorry I pushed wrong commit ealier that overrode my change

x/batch.go
```diff
@@ -0,0 +1,33 @@
...
+	// Get acts like workflow.Future.Get, but it reads out all wrapped futures into the provided slice pointer.
+	// You MUST either
+	//	1. provide a pointer to a slice as the value-pointer here, but the slice itself can be nil - it will be allocated and/or resized to fit if needed.
+	//	2. provide a nil to indicate that you don't want to collect the results.
+	//
+	// This call will wait for all futures to resolve, and will then write all results to the output slice in the same order as the input.
+	//
+	// Any errors encountered are merged with go.uber.org/multierr, so single errors are
+	// exposed normally, but multiple ones are bundled in the same way as errors.Join.
+	// For consistency when checking individual errors, consider using `multierr.Errors(err)` in all cases.
```
Groxx:
```suggestion
	// For consistency when checking individual errors, consider using `multierr.Errors(err)` in all cases,
	// or `GetFutures()[i].Get(ctx, nil)` to get the original errors at each index.
```
^ seems worth adding?

internal/batch/batch_future_test.go
shijiesheng: added a TODO

internal/batch/batch_future_test.go
shijiesheng: My approach does have race condition. This is better and I've changed to this one.

x/batch.go
shijiesheng: sure

x/batch.go
```diff
@@ -0,0 +1,40 @@
...
+	GetFutures() []workflow.Future
+}
+
+// NewBatchFuture creates a bounded-concurrency helper for doing bulk work in your workflow.
+// It does not reduce the amount of history your workflow stores, so any event-count
+// or history-size limits are unaffected - you must still be cautious about the total
+// amount of work you do in any workflow.
+//
+// When NewBatchFuture is called, futures created by the factories will be started sequentially until the concurrency limit (batchSize) is reached.
+// The remaining futures will be queued and started as previous futures complete, maintaining the specified concurrency level.
```
Groxx:
```suggestion
// When NewBatchFuture is called, futures created by the factories will be started concurrently until the concurrency limit (batchSize) is reached.
// The remaining factories will be queued and started as previous futures complete, maintaining the specified concurrency level.
```
to try to blend all the discussion together, since "started concurrently" is I think important to be accurate about.

we could change it to sequential, but I think concurrent is probably strictly more useful, since it'll more-reasonably let you do blocking things in the factory.

Groxx (approved):
small comment on the NewBatchFuture doc, but otherwise LGTM and ready to go 👍

PR #1427, opened 2025-06-17
Introduce ExecuteWithVersion and ExecuteWithMinVersion to GetVersion

arzonus:
**What changed?**

- Added support for new version control options in workflow execution:
  - Introduced `GetVersionOption` interface and related options
  - Added `ExecuteWithVersion` option to use specific version instead of max supported version
  - Added `ExecuteWithMinVersion` option to use specific version instead of min supported version
- Enhanced `GetVersion` function to support `GetVersionOptions` optional parameters
- Added test coverage for new version control features
- Updated documentation with detailed examples of version control usage
- Made changes making SearchAttributes are updated and DecisionMarker is created only if version is not equal to `DefaultVersion`
- `GetVersion` of `testWorkflowEnvironmentImpl` got support of new options `ExecuteWithVersion` and `ExecuteWithMinVersion` 

**Why?**

This change improves the workflow versioning system by providing more granular control over version execution. The new options allow for safer rollouts of workflow changes by separating code deployment from feature activation

Testing framework must emulate Cadence worker to enable customer writing unit tests. Without the change, customers will not be able to start using new functions in their unit tests. 

**How did you test it?**

- Added additional test cases  in GetVersion 
- Manual reproducing of the steps mentioned in description of `ExecuteWithVersion` function to reproduce a safe rollout. 
- Integration tests reproducing of the steps mentioned in description of `ExecuteWithVersion` function to reproduce a safe rollout. 

**Potential risks**
- Existing workflows using `GetVersion` without options will continue to work as before
- If maxSupported version was used to be DefaultVersion, it created SearchAttributes and DecisionMarker, but the changes don't do anymore.

**Detailed Description**
This PR introduces a new version control system for workflows by adding `GetVersionOption `interface and two new options (`ExecuteWithVersion`and `ExecuteWithMinVersion`) to the existing `GetVersion` function, allowing for more granular control over workflow version execution and safer deployment strategies. The changes maintain backward compatibility while enabling a three-step deployment process that separates code deployment from feature activation, ensuring safe rollback capabilities and better version management.

##### 1. WorkflowInterceptor, WorkflowEnvironment, TestWorkflowEnvironment Interfaces
```go
// Before
GetVersion(ctx Context, changeID string, minSupported, maxSupported Version) Version

// After
GetVersion(ctx Context, changeID string, minSupported, maxSupported Version, opts ...GetVersionOption) Version
```
- Added variadic parameter `opts ...GetVersionOption` to support version control options

###### GetVersionOption Type
```go
type GetVersionOption interface {
	apply(*getVersionConfig)
}
```
- New function type for configuring GetVersion behavior

###### ExecuteWithVersion
```go
// ExecuteWithVersion returns a GetVersionOption that forces a specific version to be returned
// when executed for the first time, instead of returning maxSupported version.
func ExecuteWithVersion(version Version) GetVersionOption
```

###### ExecuteWithMinVersion
```go
// ExecuteWithMinVersion returns a GetVersionOption that makes GetVersion return minSupported version
// when executed for the first time, instead of returning maxSupported version.
func ExecuteWithMinVersion() GetVersionOption
```

###### New Implementation
```go
	switch {

	// GetVersion for changeID is called first time in replay mode, use DefaultVersion
	case wc.isReplay:
		version = DefaultVersion

	// If ExecuteWithVersion option is used, use the custom version provided
	case options.customVersion != nil:
		version = *options.customVersion

	// If ExecuteWithMinVersion option is set, use the minimum supported version
	case options.useMinVersion:
		version = minSupported

	// Otherwise, use the maximum supported version
	default:
		version = maxSupported
	}
```
**Impact Analysis**
- **Backward Compatibility**: 
  - New changes are fully backward-compatible
  - If there is a usage of the new functions, there is no forward compatibility, as they have been introduced, so the client code will not be compiled. 

**Testing Plan**
  - Unit tests covering the new functions have been added
  - The changes are not related to a data type which is persisted
  - Yes, there are tests covering the full cycle of version upgrade. They uses the replayer to emulate replaying on workers supporting previous and newer versions of a workflow
  - The updated GetVersion uses variadic argument `GetVersionOption` that makes the changes compatible with the previous signature. At the same time, `testify.Mock` supports accepting `GetVersion` without a variadic argument, so there is no need to change the mock. 

**Rollout Plan**
- What is the rollout plan?
  - The new functions will be available for usage immediately after the release of the library
  - No
  - If customers start to use, they will have to remove the usage of the functions, otherwise their code will not be compiled
  - No
---


internal/workflow.go
```diff
@@ -1565,11 +1565,97 @@ const DefaultVersion Version = -1
...
+
+// GetVersionOptions contains options for GetVersion
+type GetVersionOptions struct {
+	// CustomVersion is used to force GetVersion to return a specific version
+	// instead of maxSupported version. Set up via ExecuteWithVersion option.
+	CustomVersion *Version
+
+	// UseMinVersion is used to force GetVersion to return minSupported version
+	// instead of maxSupported version. Set up via ExecuteWithMinVersion option.
+	UseMinVersion bool
```
3vilhamster, line 1583: Nice, it will allow avoiding very wordy:
```
GetVersion(ctx, workflow.DefaultVerion, 1, workflow.WithVersion(workflow.DefaultVerion))
```

3vilhamster (commented):
I like the idea, and the implementation is rather small.

But I remember that we wanted to introduce x/ folder for experiments, and this API could be experimental for now until we are fully committed on the final design and merge it to the main.

internal/internal_event_handlers.go
```diff
@@ -602,29 +602,60 @@ func validateVersion(changeID string, version, minSupported, maxSupported Versio
...
 		version = maxSupported
+	}
+
+	// Validate the version against the min and max supported versions
+	// ensuring it is within the acceptable range
+	validateVersion(changeID, version, minSupported, maxSupported)
+
+	// If the version is not the DefaultVersion, and it's not a replay, record it and update search attributes
+	// Keeping the DefaultVersion as a special case where no version marker is recorded
+	if !wc.isReplay && version != DefaultVersion {
```
taylanisikdemir, line 646: the `version != DefaultVersion` condition is confusing to me. Don't we need version marker if DefaultVersion is specified by user?

internal/workflow.go
```diff
@@ -1565,11 +1565,97 @@ const DefaultVersion Version = -1
 // CadenceChangeVersion is used as search attributes key to find workflows with specific change version.
 const CadenceChangeVersion = "CadenceChangeVersion"
 
+// GetVersionOption configures GetVersion behavior
+type GetVersionOption func(*GetVersionOptions)
+
+// GetVersionOptions contains options for GetVersion
+type GetVersionOptions struct {
```
taylanisikdemir: I understand this is not supposed to be used directly and instead helper functions below are recommended. However this is still public and open for misuse. Let's clarify the precedence of the fields in the comments

internal/workflow.go
```diff
@@ -1565,11 +1565,97 @@ const DefaultVersion Version = -1
...
+// both versions of the workflow code are compatible with each other.
+func ExecuteWithVersion(version Version) GetVersionOption {
+	return func(o *GetVersionOptions) {
+		o.CustomVersion = &version
+	}
+}
+
+// ExecuteWithMinVersion returns a GetVersionOption that makes GetVersion return minSupported version
+// when executed for the first time, instead of returning maxSupported version.
+// To see how this option can be used, see the ExecuteWithVersion option
```
taylanisikdemir: `ExecuteWithMinVersion` is not mentioned above. I see some comments about it in the tests but would be nice to provide some example here

workflow/workflow.go
```diff
@@ -353,6 +356,77 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+//	GetVersion(ctx, "fooChange", 1, 1)
+//	err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//
+// ExecuteWithVersion option is useful when you want to ensure that your changes can be safely rolled back if needed, as
+// both versions of the workflow code are compatible with each other.
+func ExecuteWithVersion(version Version) GetVersionOption {
+	return func(o *internal.GetVersionOptions) {
+		o.CustomVersion = &version
+	}
+}
```
Groxx: I've had nothing but pain as a user of function-closure functional-options :|  Building wrappers and tests around them is a huge mess because they're not comparable.

It's more annoying for us, but we should definitely just change these to be an interface-impl of some type, e.g. a `type versionOption int`.  The type can be opaque (e.g. if we want to go the extra mile, shove it in a `type versionOption struct{int}`, but I don't know OTOH if that gives us any more protection), but "comparable" is extremely useful.

test/replaytests/versioned_workflow.go
```diff
@@ -0,0 +1,213 @@
...
+
+// BarActivity returns "bar" as a result of the activity execution.
+func BarActivity(ctx context.Context, _ string) (string, error) {
+	return "bar", nil
+}
+
+// BazActivity returns "baz" as a result of the activity execution.
+func BazActivity(ctx context.Context, _ string) (string, error) {
+	return "baz", nil
+}
```
Groxx: you can simplify the setup a fair bit by just using one activity func and giving it different names, no need for them to actually be different.

and same as for workflows - drop the arg/return, it's irrelevant and just adds noise.

test/replaytests/versioned_workflow.go
```diff
@@ -0,0 +1,213 @@
...
+	ctx = workflow.WithActivityOptions(ctx, activityOptions)
+
+	var result string
+	err := workflow.ExecuteActivity(ctx, FooActivityName, "data").Get(ctx, &result)
+	if err != nil {
+		return "", err
+	}
+
+	return result, nil
+}
```
Groxx: for all of these: I'd just drop the args and return values.  they're completely unused.

that'll also get rid of the need to do anything for error checking, just return the value of `.Get(ctx, nil)`.

test/replaytests/replay_test.go
```diff
@@ -191,3 +191,137 @@ func TestContinueAsNew(t *testing.T) {
...
+		SetupWorkerForVersionedWorkflowV1(replayer)
+
+		t.Run("successfully replayed with VersionedWorkflowFoo", func(t *testing.T) {
+			err := replayer.ReplayWorkflowHistoryFromJSONFile(zaptest.NewLogger(t), versionedWorkflowFooHistoryFile)
+			require.NoError(t, err, "Failed to replay VersionedWorkflowFoo history")
+		})
+
+		t.Run("fail to replay with VersionedWorkflowBar", func(t *testing.T) {
+			err := replayer.ReplayWorkflowHistoryFromJSONFile(zaptest.NewLogger(t), versionedWorkflowBarHistoryFile)
+			require.Error(t, err, "Expected to fail replaying VersionedWorkflowBar history")
```
Groxx: definitely going to need some kind of check for the error contents, as there are LOTS of ways they can fail.  `require.ErrorContains` is probably good enough tho.

test/replaytests/replay_test.go
```diff
@@ -191,3 +191,137 @@ func TestContinueAsNew(t *testing.T) {
...
+// safely across different versions without causing non-deterministic errors.
+// There are 2 workflow executions:
+//
+// * VersionedWorkflowFoo - which is the first version of the workflow which version of change id is DefaultVersion
+//   - This workflow is supposed to execute FooActivity
+//
+// * VersionedWorkflowBar - which is the second version of the workflow which version of change id is 1
+//   - This workflow is supposed to execute BarActivity
+//
+// There are 4 versions of the workflow:
```
Groxx: four?

test/replaytests/versioned_workflow.go
```diff
@@ -0,0 +1,213 @@
...
+
+// VersionedWorkflowV3 is the third version of the workflow. It supports DefaultVersion and Version 1 as well.
+// However, all workflows started by this version will have the change ID set to Version 1.
+// It supports workflow executions started by VersionedWorkflowV1 and VersionedWorkflowV2,
+// as all of them will have the change ID set to DefaultVersion, and it supports them.
+// It also supports workflow executions started by VersionedWorkflowV3 and VersionedWorkflowV4,
+// because the code supports execution of Version 1 of the workflow.
```
Groxx: quite a lot of strange phrasing in here, e.g.

> It supports DefaultVersion and Version 1 as well.

but there's nothing to compare to and say "as well".

> It supports workflow executions started by ... and it supports them.

pretty self explanatory.

> It also supports workflow executions started by VersionedWorkflowV3

I should hope so, this is version 3!

> It also supports workflow executions started by ... VersionedWorkflowV4

and version 5.

---

this is somewhat true of other doc-comments too, but this kinda the most egregious.

workflow/workflow.go
```diff
@@ -353,6 +356,77 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+// When the previous version of the code is no longer running, there is no need to start new workflow executions
+// with DefaultVersion anymore, and you can the maxSupported version to activate the new code. To achieve that you can
+// remove the usage of ExecuteWithVersion option. When GetVersion is executed for the first time, it will return maxSupported version:
+//
+//	v := GetVersion(ctx, "fooChange", DefaultVersion, 1)
+//	if v == DefaultVersion {
+//	    err = workflow.ExecuteActivity(ctx, foo).Get(ctx, nil)
+//	} else {
+//	    err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//	}
```
Groxx: hmmm.  not sure this one makes much sense tbh.  partly because it's sorta ambiguous about what "the code" is referring to, e.g. is it any default-version of this workflow?  it's in GetVersion-related features, so that seems like it might be reasonable.

(I know it's "there are no workers running pre-GetVersion code", but someone new to versioning and this feature likely won't jump to that conclusion)

also:

> and you can the maxSupported version to activate the new code.

has accidentally a word, and `maxSupported` didn't change compared to the previous, so it's a bit misleading about what changed / what effect it had.

internal/workflow.go
Groxx: it's in an internal folder, so this isn't visible outside the client's library (unless we alias it out, which isn't happening below)

internal/internal_event_handlers.go
Groxx: OTOH I _think_ it can be skipped in that case, because during replay it'll be treated as "default" if it doesn't exist.

I will definitely need to think about that harder to be sure it's safe in all possible uses though 🤔.  I think there's a pretty good chance it is, but cross-version logic is full of possibilities and it's worth trying to be exhaustive.

test/integration_test.go
```diff
@@ -31,6 +31,8 @@ import (
 	"testing"
 	"time"
 
+	"go.uber.org/cadence/test/replaytests"
+
```
Groxx: huh.  I wonder why this file isn't being auto-goimports'd...  CI allowed it though so there's apparently a gap

test/integration_test.go
```diff
@@ -551,11 +554,119 @@ func (ts *IntegrationTestSuite) TestNonDeterministicWorkflowQuery() {
...
+	case "TestVersionedWorkflowV3":
+		replaytests.SetupWorkerForVersionedWorkflowV3(w)
+	case "TestVersionedWorkflowV4":
+		replaytests.SetupWorkerForVersionedWorkflowV4(w)
+	case "TestVersionedWorkflowV5":
+		replaytests.SetupWorkerForVersionedWorkflowV5(w)
+	case "TestVersionedWorkflowV6":
+		replaytests.SetupWorkerForVersionedWorkflowV6(w)
+	}
+}
```
Groxx: just put these inline, only a single test cares about each individual value.  no need to add a shared switch statement that every test runs.

test/integration_test.go
```diff
@@ -551,11 +554,119 @@ func (ts *IntegrationTestSuite) TestNonDeterministicWorkflowQuery() {
...
+func (ts *IntegrationTestSuite) TestVersionedWorkflowV1() {
+	execution, err := ts.executeWorkflow("test-versioned-workflow-v1", replaytests.VersionedWorkflowName, nil, "arg")
+	ts.NoError(err)
+
+	ts.NoError(ts.replayVersionedWorkflow(replaytests.SetupWorkerForVersionedWorkflowV2, execution), "Failed to replay on the replayer with VersionedWorkflowV2")
+	ts.NoError(ts.replayVersionedWorkflow(replaytests.SetupWorkerForVersionedWorkflowV3, execution), "Failed to replay on the replayer with VersionedWorkflowV3")
+	ts.Error(ts.replayVersionedWorkflow(replaytests.SetupWorkerForVersionedWorkflowV4, execution), "Expected to fail replaying the replayer with VersionedWorkflowV4")
+	ts.Error(ts.replayVersionedWorkflow(replaytests.SetupWorkerForVersionedWorkflowV5, execution), "Expected to fail replaying the replayer with VersionedWorkflowV5")
+	ts.Error(ts.replayVersionedWorkflow(replaytests.SetupWorkerForVersionedWorkflowV6, execution), "Expected to fail replaying the replayer with VersionedWorkflowV6")
+}
```
Groxx: hmm.  tbh these are a lot easier to follow than the replay_test.go versions, and they're also a lot more comprehensive.  and they're also replays, the replay_test.go versions are essentially just a subset of these + a `json.Unmarshal(data, &execution)`, which doesn't really need to be tested.

maybe just get rid of the recorded histories + replay_test additions, and move the definitions into this package?

---

also personally I think I'll vote to add "execute v1 and assert no error replaying with v1" because both:
 1. it should be true (so it's kinda a sanity test for these tests), and
 2. it'll make the assertions structurally identical across all of these, which makes it a bit easier to glance and see the range of support (you don't need to pay attention to where one is missing).

(arguably you could turn this into a helper that just takes `true, true, true, false, false, false` or similar, since that's the only thing that is actually changing between each test)

and... hopefully all these fail with the same type of error?  probably worth checking though, so other causes can't sneak in.

test/replaytests/versioned_workflow.go
```diff
@@ -0,0 +1,213 @@
...
+// All workflows started by this version will have the change ID set to Version 1.
+// It supports workflow executions started by VersionedWorkflowV3, VersionedWorkflowV4,
+// VersionedWorkflowV5, VersionedWorkflowV6
+func VersionedWorkflowV5(ctx workflow.Context, _ string) (string, error) {
+	ctx = workflow.WithActivityOptions(ctx, activityOptions)
+
+	var result string
+	var err error
+
+	version := workflow.GetVersion(ctx, TestChangeID, 1, 2, workflow.ExecuteWithVersion(1))
```
Groxx: no JSONs exist with version 2, which makes v5 / v6 a bit fuzzy on replay_test.go... but the integration tests are probably good enough there

test/integration_test.go
```diff
@@ -617,7 +726,7 @@ func (ts *IntegrationTestSuite) executeWorkflowWithOption(
 			logger.Info(event.String())
 		}
 	}
-	return err
+	return execution, err
```
Groxx, line 765: surprisingly simple to add this return.  👍 yea, seems like a good strategy.

Groxx (commented):
Some small stuff, some not-so-small stuff, and I think I need to spend some more time to think through the implementation's details... but I think it's on the right path, and it's a low-level semantic that users can't really build on their own so it seems worth having.  though it is kinda fiddly (but I think that's unavoidable).

And, generally speaking, docs need a fair bit of work I think.  They refer to things that don't exist, leave gaps in some places, and tbh I kinda doubt that anyone new to versioning will know what to do from just reading the docs :\  I do think we need to get that into a relatively good state before merging, since users will roughly immediately be exposed to it.

---

this is kinda making me wonder if we should make an entirely new API for this, to just switch to functional options everywhere.  like maybe
```go
v := GetVersionV2(
  ctx, "name",                  // unchanged
  3,                            // desired version (equal to ExecuteWithVersion)
  workflow.MaxVersion(4),       // default..4 range
  workflow.VersionRange(-1, 4), // or an explicit min/max
                                // if omitted, same as `GetVersion(ctx, 3, 3)`
)
```
since I think we can't avoid an explicit "desired" version.

but we can also do that later.  adding varargs like you've got here is _probably_ the least-effort way to add features "in place" with no real migration needed - technically a breaking change, but not in a way most code will notice.

internal/internal_event_handlers.go
arzonus: Technically, customers could specify `DefaultVersion` as the max version and a new version marked would be added to the history. However, I am not fully sure how it will be replayed on the worker where no call of `GetVersion` that can handle it, so I think the execution probably will fail. 

So, to ensure that DefaultVersion will be handled correctly on the worker without `GetVersion` I skipped the adding of the marker, because with new functions a new version marker will definitely be added.

workflow/workflow.go
arzonus: Good point about loosing comparability, I have changed it to just an slice of `internal.GetVersionOptions`. 
I believe the customers should not have direct access to this structure, because they should use only `ExecuteWithVersion` and `ExecuteWithMinVersion` functions to specify the GetVersion behavior, so decided to not expose them.

internal/workflow.go
```diff
@@ -1565,11 +1565,94 @@ const DefaultVersion Version = -1
 // CadenceChangeVersion is used as search attributes key to find workflows with specific change version.
 const CadenceChangeVersion = "CadenceChangeVersion"
 
+// GetVersionOptions contains options for GetVersion
+type GetVersionOptions struct {
```
shijiesheng: This naming is super confusing. `GetVersion` actually means "get version or set if empty". Here this option only change how we set version. Thus may I suggest name change to be `SetVersionOptions`?

test/integration_test.go
arzonus: Didn't know that the worker can register workflow after its start, so removed it

test/integration_test.go
arzonus: fixed

workflow/workflow.go
arzonus: Rewritten

test/replaytests/replay_test.go
arzonus: fixed

test/replaytests/replay_test.go
arzonus: Added them, unfortunately for cases where `GetVersion` panics, the returned error has encoded bytes, so can't fetch a particular reason like `version > maxSupported`, so used only `WORKFLOW_WORKER_UNHANDLED_FAILURE`

test/replaytests/versioned_workflow.go
arzonus: Simplified

test/replaytests/versioned_workflow.go
arzonus: Tried to simplify them to one activity with different arguments and errors stopped occurring for incompatible workers. Decided to keep it them in this way.

internal/workflow.go
arzonus: Rewritten

internal/workflow.go
arzonus: Decided to keep it `GetVersionOptions`, as for option structures for functions we use a naming convention - `<FuncName>Options`. Anyway, it's a internal type, so we will always have an option to rename it.

test/integration_test.go
arzonus: > maybe just get rid of the recorded histories + replay_test additions, and move the definitions into this package?

Agree, there are overlaps, so decided to simplify and move them to integration tests. 

> also personally I think I'll vote to add "execute v1 and assert no error replaying with v1"

Good point, added it

> (arguably you could turn this into a helper that just takes true, true, true, false, false, false or similar, since that's the only thing that is actually changing between each test)

Add `testVersionedWorkflowTestCase` with compatible and inCompatible versions fields to have it more explicit instead of just boolean values. 

>  with the same type of error?

It's tricky, because the replayer returns `fmt.Error` with encoded in-bytes real error.

test/replaytests/versioned_workflow.go
arzonus: Removed replay_test, no need to have json

test/versioned_workflow.go
```diff
@@ -0,0 +1,234 @@
+package test
```
Groxx: this is all exposed publicly, "test" is not a special package name or anything.  probably just needs to be moved to a _test.go file.

looks like we have some other unnecessarily-public stuff in this package too 🤔  we can clean that up some other time though.  or just shove it all in `internal/test` to prevent accidents since it's really not something anyone else should be using.

internal/workflow_testsuite.go
```diff
@@ -365,8 +365,8 @@ func (t *TestWorkflowEnvironment) OnRequestCancelExternalWorkflow(domainName, wo
 }
 
 // OnGetVersion setup a mock for workflow.GetVersion() call. By default, if mock is not setup, the GetVersion call from
-// workflow code will always return the maxSupported version. Make it not possible to test old version branch. With this
-// mock support, it is possible to test code branch for different versions.
+// workflow code will always return the maxSupported version or version specified by GetVersionOption.
+// Make it not possible to test old version branch. With this mock support, it is possible to test code branch for different versions.
```
Groxx:
```suggestion
```
I kinda think this whole line is confusing and not worth keeping tbh

internal/workflow.go
```diff
@@ -1617,28 +1693,28 @@ const CadenceChangeVersion = "CadenceChangeVersion"
 //	err = workflow.ExecuteActivity(ctx, baz).Get(ctx, nil)
 //
 // The reason to keep it is: 1) it ensures that if there is older version execution still running, it will fail here
-// and not proceed; 2) if you ever need to make more changes for “fooChange”, for example change activity from baz to qux,
+// and not proceed; 2) if you ever need to make more changes for "fooChange", for example change activity from baz to qux,
```
Groxx, line 1665: always 👍 for getting rid of smart quotes

internal/workflow.go
```diff
@@ -1567,11 +1567,87 @@ const DefaultVersion Version = -1
...
+	// CustomVersion is used to force GetVersion to return a specific version
+	// instead of maxSupported version. Set up via ExecuteWithVersion option.
+	CustomVersion *Version
+
+	// UseMinVersion is used to force GetVersion to return minSupported version
+	// instead of maxSupported version. Set up via ExecuteWithMinVersion option.
+	UseMinVersion bool
+}
+
+// ExecuteWithVersion forces a specific version to be returned when GetVersion is executed for the first time,
```
Groxx, line 1594: tbh I think I'll vote to just do this on the public version, and leave a "see the public version" note here - we're not too likely to keep them in sync, and nobody uses the internal version.

internal/internal_event_handlers.go
```diff
@@ -602,28 +602,58 @@ func validateVersion(changeID string, version, minSupported, maxSupported Versio
...
 	if version, ok := wc.changeVersions[changeID]; ok {
 		validateVersion(changeID, version, minSupported, maxSupported)
 		return version
 	}
 
+	// Use the first option if they're present
+	var options GetVersionOptions
+	if len(opts) > 0 {
+		options = opts[0]
+	}
```
Groxx: tbh I think if we're going to do this, we should make a wholly new API that requires it / doesn't have varargs.  hidden noops is quite likely to be a surprise to anyone who tries to use it.
though I do see the general "it only makes sense for a single version arg to ever be passed" API-design issue.  that's kinda fundamental for functional options though, and we've already got an established "irrational args panic" pattern here so we can do whatever validation we need.

---

anyway:

implementing as a shared struct internally: seems fine, whatever's easy to use, we can change it later if we want.  it's instantiated internally so we don't need to worry about irrational combinations of values.

breaking functional-opts' "last value wins" or "irrational args panic" patterns: not worth it imo, probably switch this to "last value wins" since that's how most of our APIs work?  e.g. yarpc opts and context-opt-wrapping both do that, and AFAICT that's the most common from what I've encountered.  for now that means choosing the _last_ index, not the first, if you don't feel like iterating through the list (which seems fine, it's unlikely to be missed when changing opts).

internal/internal_event_handlers.go
```diff
@@ -602,28 +602,58 @@ func validateVersion(changeID string, version, minSupported, maxSupported Versio
...
 		wc.decisionsHelper.recordVersionMarker(changeID, version, wc.GetDataConverter())
 		err := wc.UpsertSearchAttributes(createSearchAttributesForChangeVersion(changeID, version, wc.changeVersions))
 		if err != nil {
 			wc.logger.Warn("UpsertSearchAttributes failed", zap.Error(err))
 		}
 	}
 
-	validateVersion(changeID, version, minSupported, maxSupported)
+	// Store the version in the changeVersions
+	// ensuring that it can be retrieved later
```
Groxx: not sure how I feel about this one tbh 🤔 
it's technically true, but mostly in a way that's true for any map, so it's not really pointing out anything useful / it's arguably implying it's "just" a map and has no special semantics.

the important part here is that storing it in this map ensures that the version is stable for the duration of the workflow, due to the check earlier in the func.  it's why it's also populated when reading history, whether GetVersion is ever called or not.

workflow/workflow.go
```diff
@@ -353,6 +353,66 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+//
+// 3. Remove a support of foo activity:
+//
+//	GetVersion(ctx, "fooChange", 1, 1)
+//	err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//
+// When there are no workflow executions running DefaultVersion the support of foo activity can be removed.
+//
+// ExecuteWithVersion option is useful when you want to ensure that your changes can be safely rolled back if needed.
+func ExecuteWithVersion(version Version) internal.GetVersionOptions {
```
Groxx: can't return an internal type like this - it'll be just as unusable / un-reference-able as returning a private type.

(tbh I'm kinda surprised goland isn't complaining, though this is fine for entirely-internal usage)

test/integration_test.go
Groxx: > > with the same type of error?
> 
> It's tricky, because the replayer returns `fmt.Error` with encoded in-bytes real error.

aah, yea, replayer's worker errors complicate this a fair bit don't they :\

eh, seems fine.  reading through all the failures by hand looks correct.

test/replaytests/versioned_workflow.go
Groxx: seems to work for me fwiw, e.g. this plus dropping args/nil-returns everywhere:
```diff
-func VersionedWorkflowV1(ctx workflow.Context, _ string) (string, error) {
+func VersionedWorkflowV1(ctx workflow.Context) error {
        ctx = workflow.WithActivityOptions(ctx, activityOptions)
 
-       var result string
-       err := workflow.ExecuteActivity(ctx, FooActivityName, "data").Get(ctx, &result)
-       if err != nil {
-               return "", err
-       }
-
-       return result, nil
+       return workflow.ExecuteActivity(ctx, FooActivityName).Get(ctx, nil)
 }
```
and everything runs identically.  which is definitely how it should behave, lots of things would have to break to make that not-work. 

that said, everything runs identically.  and they do a pretty good job at showing the compatibility behavior, which is what we really want to achieve.  and it's not enough visual-noise to hide what's actually happening because they're nice and small.  
so I think they're fine to leave 👍

test/integration_test.go
```diff
@@ -551,11 +550,159 @@ func (ts *IntegrationTestSuite) TestNonDeterministicWorkflowQuery() {
...
+
+	ts.Require().Equalf(len(c.compatibleVersions)+len(c.inCompatibleVersions), int(MaxVersionedWorkflowVersion),
+		"Test case should cover all versions, but got %d compatible (one of them the testing version itself) and %d incompatible versions, that not equal to %d",
+		len(c.compatibleVersions),
+		len(c.inCompatibleVersions),
+		MaxVersionedWorkflowVersion)
+
+	for _, replayedVersion := range c.compatibleVersions {
+		err := ts.replayVersionedWorkflow(replayedVersion, execution)
+		ts.NoErrorf(err, "Failed to replay on the replayer with VersionedWorkflowV%d", replayedVersion)
```
Groxx, line 691:
```
... with VersionedWorkflowV%d", replayedVersion)
```
always glad to see "assertions in loops can be uniquely identified", thanks 👍

internal/internal_workflow_testsuite_test.go
```diff
@@ -1274,6 +1274,228 @@ func (s *WorkflowTestSuiteUnitTest) Test_GetVersion() {
...
+		// test searchable change version
+		wfInfo := GetWorkflowInfo(ctx)
+		s.NotNil(wfInfo.SearchAttributes)
+		changeVersionsBytes, ok := wfInfo.SearchAttributes.IndexedFields[CadenceChangeVersion]
+		s.True(ok)
+		var changeVersions []string
+		err = json.Unmarshal(changeVersionsBytes, &changeVersions)
+		s.NoError(err)
+		s.Equal(1, len(changeVersions))
+		s.Equal("test_change_id-1", changeVersions[0])
```
Groxx: I really wish this thrift type wasn't leaked in here 😭
but 👍 yep, that's the correct way to do it.  sadly.

also, bleh.  OnGetVersion and OnUpsertSearchAttributes and some others don't support matchers?  wat.  otherwise that'd probably be an easier way to make this assertion :(

internal/internal_workflow_testsuite_test.go
```diff
@@ -1274,6 +1274,228 @@ func (s *WorkflowTestSuiteUnitTest) Test_GetVersion() {
...
+
+	env := s.NewTestWorkflowEnvironment()
+	env.RegisterWorkflow(workflowFn)
+	env.RegisterActivity(oldActivity)
+	env.RegisterActivity(newActivity)
+
+	env.OnGetVersion("change_1", DefaultVersion, 2).Return(func(string, Version, Version) Version {
+		return DefaultVersion
+	})
+	env.OnGetVersion(mock.Anything, DefaultVersion, 2).Return(Version(2))
```
Groxx:
```suggestion
	env.OnGetVersion("change_1", DefaultVersion, 2).Return(DefaultVersion)
	env.OnGetVersion("change_2", DefaultVersion, 2).Return(Version(2))
```
since it's weirdly inconsistent and loose otherwise (also below)

internal/internal_workflow_testsuite_test.go
```diff
@@ -1274,6 +1274,228 @@ func (s *WorkflowTestSuiteUnitTest) Test_GetVersion() {
...
+		} else {
+			f = ExecuteActivity(ctx, newActivity, "new1")
+		}
+		var ret1 string
+		err := f.Get(ctx, &ret1) // wait for result
+		if err != nil {
+			return "", err
+		}
+
+		v2 := GetVersion(ctx, "change_2", DefaultVersion, 2)
```
Groxx, line 1384: since the only difference between this test and the next is
```
v1 := GetVersion(ctx, "change_1", DefaultVersion, 2, ExecuteWithVersion(1))
v1 := GetVersion(ctx, "change_1", DefaultVersion, 2, 2)

v2 := GetVersion(ctx, "change_2", DefaultVersion, 2)
v2 := GetVersion(ctx, "change_2", DefaultVersion, 2, ExecuteWithMinVersion())
```
... I kinda think this isn't testing much related to the new code.  some things about it, but mostly just "if mock-versioned, it still uses that version".

probably deserves a "when not mocking GetVersion, the test suite behaves like prod, and respects the WithVersion/MinVersion value" test too.  which can probably just be one, checking that both activities are not using the max values.

internal/internal_workflow_testsuite_test.go
```diff
@@ -1274,6 +1274,228 @@ func (s *WorkflowTestSuiteUnitTest) Test_GetVersion() {
...
+	}
+	workflowFn := func(ctx Context) error {
+		ctx = WithActivityOptions(ctx, s.activityOptions)
+		var f Future
+		v := GetVersion(ctx, "test_change_id", DefaultVersion, 2, ExecuteWithVersion(1))
+		if v == DefaultVersion {
+			f = ExecuteActivity(ctx, oldActivity, "ols_msg")
+		} else {
+			f = ExecuteActivity(ctx, newActivity, "new_msg")
+		}
```
Groxx, line 1331: (also below)
probably deserves a change to show "does not use v1 even though it's `ExecuteWithVersion(1)` since as written it's a seemingly unsafe change.  a `return errors.New("no v1 call expected")` is probably good enough.

workflow/workflow.go
```diff
@@ -353,6 +353,66 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+//	err = workflow.ExecuteActivity(ctx, foo).Get(ctx, nil)
+//
+// It should be updated to:
+//
+//	err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//
+// Following the steps below, your changes will be forward and backward compatible, keeping possible a safe rollback
+// to the previous version of the workflow code:
+//
+// 1. Keep execution of foo activity, add a support of bar activity
```
Groxx:
```suggestion
// 1. Keep execution of foo activity, add support for bar activity
```

workflow/workflow.go
```diff
@@ -353,6 +353,66 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+//	v := GetVersion(ctx, "fooChange", DefaultVersion, 1,  ExecuteWithVersion(DefaultVersion))
+//	if v == DefaultVersion {
+//		err = workflow.ExecuteActivity(ctx, foo).Get(ctx, nil)
+//	} else {
+//		err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//	}
+//
+// The code above supports replaying of workflow execution with both versions DefaultVersion and 1.
+// All new workflow executions will execute foo activity, because GetVersion with ExecuteWithMinVersion option
+// returns DefaultVersion, that will be recorded into the workflow history.
```
Groxx:
```suggestion
// All new workflow executions will execute foo activity, because ExecuteWithVersion(DefaultVersion)
// causes GetVersion to return DefaultVersion rather than 1.
// In this example, this is also exactly the same as using ExecuteWithMinVersion().
```
I'm kinda divided on keeping/removing "will be recorded into the workflow history" here... for non-default versions it's true, but for this specific example it won't record the version at all because it's the default.  though the behavior will otherwise act as if it was recorded.

workflow/workflow.go
```diff
@@ -353,6 +353,66 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+//	v := GetVersion(ctx, "fooChange", DefaultVersion, 1)
+//	if v == DefaultVersion {
+//	    err = workflow.ExecuteActivity(ctx, foo).Get(ctx, nil)
+//	} else {
+//	    err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//	}
+//
+// The code above supports replaying of workflow execution with both versions DefaultVersion and 1.
+// All new workflow executions will execute bar activity, because
+// GetVersion returns the maximum supported version - 1, that will be recorded into the workflow history.
```
Groxx:
```suggestion
// GetVersion returns the maximum supported version (1), which will be recorded into the workflow history.
```

workflow/workflow.go
```diff
@@ -353,6 +353,66 @@ func MutableSideEffect(ctx Context, id string, f func(ctx Context) interface{},
...
+//
+// 3. Remove a support of foo activity:
+//
+//	GetVersion(ctx, "fooChange", 1, 1)
+//	err = workflow.ExecuteActivity(ctx, bar).Get(ctx, nil)
+//
+// When there are no workflow executions running DefaultVersion the support of foo activity can be removed.
+//
+// ExecuteWithVersion option is useful when you want to ensure that your changes can be safely rolled back if needed.
```
Groxx:
```suggestion
```
kinda just restating the beginning of the doc

workflow/workflow.go
```diff
@@ -418,8 +478,8 @@ const DefaultVersion Version = internal.DefaultVersion
 //	} else {
 //	  err = workflow.ExecuteActivity(ctx, qux, data).Get(ctx, nil)
 //	}
-func GetVersion(ctx Context, changeID string, minSupported, maxSupported Version) Version {
-	return internal.GetVersion(ctx, changeID, minSupported, maxSupported)
+func GetVersion(ctx Context, changeID string, minSupported, maxSupported Version, opts ...internal.GetVersionOptions) Version {
```
Groxx: probably worth mentioning ExecuteWithVersion somewhere in these docs so it's more discoverable, but I have no concrete suggestion at the moment

Groxx (commented):
Yea, I think the behavior's correct.  Let's do this 👍 

I think comments summarize as:
- some minor cleanups
- can't expose `internal.GetVersionOptions` like that, probably just make an empty interface with a private method and alias it.
  - technically you can also cast to a same-shaped struct and re-implement or something.  which is what should have been done for ~all of the things in these public packages.  but that ship has sailed, more aliases is fine IMO since we can't remove the existing ones.
- like one more test-suite test, to show that it behaves like the integration tests

otherwise looks good :)

test/versioned_workflow.go
arzonus: Good point, fixed it

workflow/workflow.go
arzonus: Mentioned it in the GetVersion documentation. Probably, we should merge GetVersion and ExecuteWithVersion doc to one chapter Versioning and point customers there.

workflow/workflow.go
arzonus: good point, changed to option interface

internal/workflow.go
arzonus: agreed, removed it and left a reference to the public version

internal/internal_event_handlers.go
arzonus: Rewritten

internal/internal_event_handlers.go
arzonus: Good point, changed it to the GetVersionOption interface approach

test/replaytests/versioned_workflow.go
arzonus: good point, simplified workflows

internal/internal_workflow_testsuite_test.go
arzonus: Yeah, the test ensures that despite on the options, `OnGetVersion` mocks behavior. 

> probably deserves a "when not mocking GetVersion, the test suite behaves like prod, and respects the WithVersion/MinVersion value" test too. 

There are already tests: `Test_GetVersion_ExecuteWithVersion` and `Test_GetVersion_ExecuteWithMinVersion` there are no mocking `GetVersion`

internal/internal_workflow_testsuite_test.go
arzonus: Not sure about the comment. The test is `Test_GetVersion_ExecuteWithVersion`, so there is no mocked `GetVersion`, so only v1 is returned. Perhaps, did you want to point another place?

internal/internal_workflow_testsuite_test.go
arzonus: good point, replaced it

test/replaytests/versioned_workflow.go
arzonus: Simplified

internal/internal_workflow_testsuite_test.go
Groxx: ah, yea, true - those check the output (the activity that was run), but not the version strings, so I wasn't thinking of them earlier.

that's good enough evidence 👍  I don't think we need much for the version strings, even a single one is probably good enough for this change since its version-search-attribute behavior is so straightforward (and we probably have other pre-existing ones too?)

internal/workflow.go
```diff
@@ -1567,13 +1567,58 @@ const DefaultVersion Version = -1
...
+	UseMinVersion bool
+}
+
+// customVersionOption forces GetVersion to return a specific version
+type customVersionOption Version
+
+func (c customVersionOption) apply(config *getVersionConfig) {
+	version := Version(c)
+	config.CustomVersion = &version
+}
```
Groxx, line 1592: yea.  verbose and kinda annoying, but I really don't think we have much of an alternative ¯\\\_(ツ)\_/¯

internal/workflow.go
```diff
@@ -1567,13 +1567,58 @@ const DefaultVersion Version = -1
...
+
 // GetVersion is used to safely perform backwards incompatible changes to workflow definitions.
 // It is not allowed to update workflow code while there are workflows running as it is going to break
 // determinism. The solution is to have both old code that is used to replay existing workflows
 // as well as the new one that is used when it is executed for the first time.
 // GetVersion returns maxSupported version when is executed for the first time. This version is recorded into the
 // workflow history as a marker event. Even if maxSupported version is changed the version that was recorded is
 // returned on replay. DefaultVersion constant contains version of code that wasn't versioned before.
+// Check documentation for ExecuteWithVersion and ExecuteWithMinVersion to make your changes forward and backward compatible.
```
Groxx, line 1620: yea, good enough for now anyway :)  we can consider a bigger rewrite some other time.

internal/internal_workflow_testsuite_test.go
Groxx: nope, I was just forgetting those tests / what they covered.  seems fine :)

Groxx (approved):
Looks good to go, thank you for the idea and all the effort!  It's a good low-level-primitive to have, this was definitely a gap in GetVersion.

we'll need to add some details to the changelog / release eventually, but that'll be another PR.

PR #1429, opened 2025-06-24
fix and enforce lint

shijiesheng:
**What changed?**

* Fix lint
* Makefile build rule will depend on lint rather than fmt

**Why?**

introduce best practice

**How did you test it?**

unit test

**Potential risks**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,7 @@ $Q +$(MAKE) --no-print-directory $(addprefix $(BUILD)/,$(1))
 endef
 
 .PHONY: build
-build: $(BUILD)/fmt ## ensure all packages build
+build: $(BUILD)/lint ## ensure all packages build
 	go build ./...
 	$Q # caution: some errors are reported on stdout for some reason
 	go test -exec true ./... >/dev/null
--- a/internal/internal_worker_test.go
+++ b/internal/internal_worker_test.go
@@ -217,10 +217,13 @@ func (s *internalWorkerTestSuite) TestCreateWorker_WithDataConverter() {
 
 func (s *internalWorkerTestSuite) TestCreateShadowWorker() {
 	worker := createShadowWorker(s.T(), s.service, &ShadowOptions{})
-	s.Nil(worker.workflowWorker)
-	s.Nil(worker.activityWorker)
-	s.Nil(worker.locallyDispatchedActivityWorker)
-	s.Nil(worker.sessionWorker)
+	w, ok := worker.(*aggregatedWorker)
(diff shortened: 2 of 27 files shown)


Makefile
```diff
@@ -252,7 +252,7 @@ $Q +$(MAKE) --no-print-directory $(addprefix $(BUILD)/,$(1))
 endef
 
 .PHONY: build
-build: $(BUILD)/fmt ## ensure all packages build
+build: $(BUILD)/lint ## ensure all packages build
```
taylanisikdemir, line 255: can we run both?

shijiesheng: it's already running both. lint depends on fmt.

internal/worker.go
```diff
@@ -38,6 +38,118 @@ import (
 )
 
 type (
+	// Worker hosts workflow and activity implementations.
```
davidporter-id-au, line 41: not following, where did this change come from? Are they intended to be part of the same lint PR?

taylanisikdemir (approved):

internal/internal_worker_test.go
```diff
@@ -261,9 +264,8 @@ func (s *internalWorkerTestSuite) TestCreateWorkerRun() {
 func (s *internalWorkerTestSuite) TestNoActivitiesOrWorkflows() {
 	t := s.T()
 	w := createWorker(s.T(), s.service)
-	w.registry = newRegistry()
-	assert.Empty(t, w.registry.getRegisteredActivities())
-	assert.Empty(t, w.registry.GetRegisteredWorkflowTypes())
+	// assert.Empty(t, w.GetRegisteredActivities())
```
davidporter-id-au: should these be asserted or cleaned up?

internal/worker.go
davidporter-id-au: ah, I missed that in the diff description earlier. What's the intent behind the copy?

taylanisikdemir: these were defined in worker/worker.go. it seems they are moved here but not aliased in the previous place

internal/internal_worker_test.go
shijiesheng: should be asserted. this is a miss

internal/worker.go
shijiesheng: Yes, it's copied from `worker/worker.go`. Basically internal package is missing this interface. 
Alternatively we can use reference which has caused issues in go tooling (like mock) before. 
```
Worker = internal.Worker
```

PR #1438, opened 2025-07-08
Add active cluster selection policy to workflow start options

taylanisikdemir:
**Detailed Description**

Following changes are made to support specifying custom policies for active-active domains:
- `ActiveClusterSelectionPolicy` is added to workflow start options
- Corresponding compatibility layers, mappers and tests updated

Misc change:
- Added `CronOverlapPolicy` to compatibility layer which was missed in #1425 

**Impact Analysis**
New field. ok to be nil.
Not rollback compatible once users start taking dependency on this.

**Testing Plan**
- **Unit Tests**: [Do we have unit test covering the change?] Yes
Already handled in server repo.
Already handled in server repo.
Will be done in dev environment

**Rollout Plan**
- What is the rollout plan?
Create a new prerelease and try it in canary services.
Server changes must be deployed first before clients start using this.
Yes before users start taking dependency on new field. 
Rolling back client is safe and should mitigate unforeseen issues. Server side changes are already in.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -490,6 +490,11 @@ type (
 		// Note: JitterStart remains used when determining the actual start time of executions.
 		// Optional: defaulted to CronOverlapPolicySkip
 		CronOverlapPolicy s.CronOverlapPolicy
+
+		// ActiveClusterSelectionPolicy - Policy for selecting the active cluster to start the workflow execution on for active-active domains.
+		// Note: This doesn't apply to local or active-passive domains.
+		// Optional: defaulted to region sticky strategy with receiver cluster as the active cluster.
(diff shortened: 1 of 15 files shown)


internal/client.go
```diff
@@ -490,6 +490,11 @@ type (
 		// Note: JitterStart remains used when determining the actual start time of executions.
 		// Optional: defaulted to CronOverlapPolicySkip
 		CronOverlapPolicy s.CronOverlapPolicy
```
shijiesheng, line 492: should be `*CronOverlapPolicy`

taylanisikdemir: This is already released so cannot change. Setting it to 0 is same as nil case. Default behavior is SKIP.

PR #1444, opened 2025-07-29
Remove all buildkite references

mandyschen:
**What changed?**
Remove all deprecated references to buildkite.

**Why?**
Because we are migrating to only use Github Actions.

**How did you test it?**
I am testing using the Github Actions pipeline.

**Potential risks**
Removing it could make it so that a push/PR passes when it shouldn't (because the Github Actions failed to catch it).


shijiesheng (approved):
Thank you for the change!

PR #1448, opened 2025-07-31
add always and test

mandyschen:
**What changed?**
Added always uploading artifacts

**Why?**
So even in failure, can look at the artifacts

**How did you test it?**
Will test using Github Actions

**Potential risks**


shijiesheng (approved):
Thank you!

PR #1449, opened 2025-08-07
Add support for Ephemeral TaskLists

natemort:
Update Sessions to use Ephemeral TaskLists behind a feature flag. This ensures that the per-host TaskList is automatically removed once it is no longer used.

This should only be enabled once the server fully supports Ephemeral TaskLists as it will otherwise return errors for the unknown TaskListKind.

**What changed?**
- Added support for dispatching session activities to ephemeral TaskLists and polling Ephemeral TaskLists

**Why?**
- To support automatically removing session and integration test TaskLists when they're no longer used.

**How did you test it?**
- Unit/integration test. Will run through canary tests before enabling.

**Potential risks**
- Errors in task polling would significantly impact users, so this is a high risk change.

**Detailed Description**
See above.

**Impact Analysis**
- **Backward Compatibility**: 

**Testing Plan**
- **Unit Tests**: Unit tests
- **Persistence Tests**: Covered in the server. See https://github.com/cadence-workflow/cadence/pull/7157
- **Integration Tests**: Added here, additionally present in the server.

**Rollout Plan**
- What is the rollout plan? Enable where fully rolled out on the server
- Does the order of deployment matter? Server must support this feature before it is enabled clientside, otherwise poll requests will return an error.
- Is it safe to rollback? Does the order of rollback matter? Safe to roll back
- Is there a kill switch to mitigate the impact immediately? Disabling the flag.


internal/client.go
```diff
@@ -656,6 +656,7 @@ func getFeatureFlags(options *ClientOptions) FeatureFlags {
 		return FeatureFlags{
 			WorkflowExecutionAlreadyCompletedErrorEnabled: options.FeatureFlags.WorkflowExecutionAlreadyCompletedErrorEnabled,
 			PollerAutoScalerEnabled:                       options.FeatureFlags.PollerAutoScalerEnabled,
```
shijiesheng, line 658: just a comment: this flag is no longer used and I'll delete it in the next PR

internal/internal_activity.go
```diff
@@ -59,6 +59,7 @@ type (
 	activityOptions struct {
 		ActivityID                    *string // Users can choose IDs but our framework makes it optional to decrease the crust.
 		TaskListName                  string
+		TaskListKind                  shared.TaskListKind
 		ScheduleToCloseTimeoutSeconds int32
 		ScheduleToStartTimeoutSeconds int32
 		StartToCloseTimeoutSeconds    int32
```
shijiesheng, line 65:
```suggestion
	activityOptions struct {
		ActivityID                    *string // Users can choose IDs but our framework makes it optional to decrease the crust.
		TaskListName                  string
		ScheduleToCloseTimeoutSeconds int32
		ScheduleToStartTimeoutSeconds int32
		StartToCloseTimeoutSeconds    int32

        // private fields that is used internally and not directly settable from the user
        taskListKind                  shared.TaskListKind
```

test/integration_test.go
```diff
@@ -583,6 +589,16 @@ func (ts *IntegrationTestSuite) TestOverrideSpanContext() {
...
 }
 
+func (ts *IntegrationTestSuite) TestSession() {
+	_, err := ts.executeWorkflow("test-session", ts.workflows.Session, nil)
+	ts.NoError(err)
+}
+
+func (ts *IntegrationTestSuite) TestSession_Ephemeral() {
+	// Ephemeral TaskList is enabled by the test name
+	ts.TestSession()
```
shijiesheng: I wonder if we can describe tasklist to ensure the tasklist is gone.

shijiesheng: somehow this failed
```suggestion
	ts.TestSession()
	response, err := ts.libClient.DescribeTaskList(ts.T().Context(), "test-session", shared.TaskListTypeActivity)
	ts.NoError(err)
	ts.Equal(shared.TaskListKindEphemeral, *response.TaskList.Kind)
```

test/workflow_test.go
```diff
@@ -659,6 +659,29 @@ func (w *Workflows) OverrideSpanContext(ctx workflow.Context) (map[string]string
...
+		CreationTimeout:  time.Second * 10,
+		ExecutionTimeout: time.Second * 10,
+	}
+	sessionCtx, err := workflow.CreateSession(ctx, so)
+	if err != nil {
+		return err
+	}
+	defer workflow.CompleteSession(sessionCtx)
+	var expected string
+	if err = workflow.ExecuteActivity(ctx, "Prefix_ToUpper", "hello").Get(ctx, &expected); err != nil {
```
shijiesheng: should be sessionCtx

test/integration_test.go
```diff
@@ -583,6 +589,16 @@ func (ts *IntegrationTestSuite) TestOverrideSpanContext() {
...
+}
+
```
shijiesheng: this could do verification

```suggestion
func (ts *IntegrationTestSuite) TestSession_Ephemeral() {
	// Ephemeral TaskList is enabled by the test name
	execution, err := ts.executeWorkflow("test-session", ts.workflows.Session, nil)
	ts.NoError(err)

	taskList := ts.getActivityTaskList(execution)
	ts.Equal(shared.TaskListKindEphemeral, taskList.GetKind())

	taskList = ts.describeTaskList(taskList.GetName())
	ts.Equal(shared.TaskListKindEphemeral, taskList.GetKind())
}

func (ts *IntegrationTestSuite) getActivityTaskList(execution *workflow.Execution) *shared.TaskList {
	iter := ts.libClient.GetWorkflowHistory(context.Background(), execution.ID, execution.RunID, false, shared.HistoryEventFilterTypeAllEvent)
	for iter.HasNext() {
		event, err := iter.Next()
		ts.Require().NoError(err)
		if event.GetActivityTaskScheduledEventAttributes() != nil {
			return event.GetActivityTaskScheduledEventAttributes().GetTaskList()
		}
	}
	return nil
}

func (ts *IntegrationTestSuite) describeTaskList(taskListName string) *shared.TaskList {
	descResp, err := ts.libClient.DescribeTaskList(context.Background(), taskListName, shared.TaskListTypeActivity)
	ts.Require().NoError(err)
	return descResp.GetTaskList()
}

```

natemort: Discussed offline and added. The workflow now returns the TaskList the activity ran on, which we can check.

test/workflow_test.go
natemort: Good catch, thanks

test/integration_test.go
natemort: The TTL is rather high unfortunately. We can't really validate that behavior without inspecting the database directly.

PR #1451, opened 2025-08-08
Add attempt number to activity logger

evansandoval:
**Detailed Description**
Adds attempt number to structured logs within `WithActivityTask` in activity.go and `executeLocalActivityTask` in internal_task_pollers.go.

**Impact Analysis**
- **Backward Compatibility**: No changes to core Cadence functionality or public API's / data structures.
- **Forward Compatibility**: Uses existing fields and doesn't introduce any new dependencies.

**Testing Plan**
- **Unit Tests**: Existing unit tests continue to pass.
- **Persistence Tests**: No changes to persisted data structures.
- **Integration Tests**: Existing tests continue to pass (activity execution behaviors are unchanged)
- **Compatibility Tests**: Existing tests continue to pass (activity execution behaviors are unchanged)

**Rollout Plan**
- What is the rollout plan?
    - Standard rollout / deployment process
    - No specific order required.
    - Completely safe to rollback, no specific order required either.
    - Not needed as no business logic changes. Can simply rollback the commit.

### DIFF
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -356,10 +356,8 @@ func WithActivityTask(
 		zapcore.Field{Key: tagWorkflowType, Type: zapcore.StringType, String: *task.WorkflowType.Name},
 		zapcore.Field{Key: tagWorkflowID, Type: zapcore.StringType, String: *task.WorkflowExecution.WorkflowId},
 		zapcore.Field{Key: tagRunID, Type: zapcore.StringType, String: *task.WorkflowExecution.RunId},
+		zapcore.Field{Key: tagAttempt, Type: zapcore.Int64Type, Integer: int64(*task.Attempt)},
 	)
-	if task.Attempt != nil {
-		logger = logger.With(zapcore.Field{Key: tagAttempt, Type: zapcore.Int64Type, Integer: int64(*task.Attempt)})
-	}
 
 	return context.WithValue(ctx, activityEnvContextKey, &activityEnvironment{
 		taskToken:      task.TaskToken,
--- a/internal/activity.go
+++ b/internal/activity.go
@@ -356,7 +356,7 @@ func WithActivityTask(
 		zapcore.Field{Key: tagWorkflowType, Type: zapcore.StringType, String: *task.WorkflowType.Name},
 		zapcore.Field{Key: tagWorkflowID, Type: zapcore.StringType, String: *task.WorkflowExecution.WorkflowId},
 		zapcore.Field{Key: tagRunID, Type: zapcore.StringType, String: *task.WorkflowExecution.RunId},
-		zapcore.Field{Key: tagAttempt, Type: zapcore.Int64Type, Integer: int64(*task.Attempt)},
+		zapcore.Field{Key: tagAttempt, Type: zapcore.Int64Type, Integer: int64(task.GetAttempt())},
 	)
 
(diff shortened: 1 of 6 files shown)


internal/activity.go
```diff
@@ -357,6 +357,9 @@ func WithActivityTask(
 		zapcore.Field{Key: tagWorkflowID, Type: zapcore.StringType, String: *task.WorkflowExecution.WorkflowId},
 		zapcore.Field{Key: tagRunID, Type: zapcore.StringType, String: *task.WorkflowExecution.RunId},
 	)
+	if task.Attempt != nil {
+		logger = logger.With(zapcore.Field{Key: tagAttempt, Type: zapcore.Int64Type, Integer: int64(*task.Attempt)})
+	}
```
shijiesheng: Can you change `localActivityTaskHandler) executeLocalActivityTask(` as well?

internal/activity.go
```diff
@@ -356,6 +356,7 @@ func WithActivityTask(
 		zapcore.Field{Key: tagWorkflowType, Type: zapcore.StringType, String: *task.WorkflowType.Name},
 		zapcore.Field{Key: tagWorkflowID, Type: zapcore.StringType, String: *task.WorkflowExecution.WorkflowId},
 		zapcore.Field{Key: tagRunID, Type: zapcore.StringType, String: *task.WorkflowExecution.RunId},
+		zapcore.Field{Key: tagAttempt, Type: zapcore.Int64Type, Integer: int64(*task.Attempt)},
```
Groxx: field makes sense and I generally assume this is fine, but: how confident are we that this is literally always present?  should we check for nils?

also I'm kinda curious if this matches the "attempt" count in the UI / history, or if it's off by 1, since we'll very likely get questions about it.

evansandoval: I don't have the background context to be able to confidently say that the field is always set, but I think it probably is better to err on the side of checking for nils.

On the note of the UI in the attempt, I think it processes this as retryAttempt, so I believe the count should match up, but I'll double check right now.

Groxx: I think the "actual" value is fine either way / I don't think I'd +1 or -1 it to try to match the UI, to be clear.  but yea, it would be nice to know ahead of time, and maybe comment it in here if it's surprising so it's more easily re-discoverable.

evansandoval: https://github.com/cadence-workflow/cadence-web/blob/master/src/views/workflow-history/helpers/get-history-group-from-events/get-activity-group-from-events.ts#L75

Assuming that pendingStartEvents' attempt field is the same to the attempt field we set in PollForActivityResponse, I think we're good?

Groxx: the web TS stuff is _extremely_ far removed from the actual thrift/proto definitions and has layers of fallbacks added, I honestly would not trust them even the slightest bit.

shijiesheng (approved):

Groxx (approved):
👍 yep, looks good.  `GetAttempt()` is reasonable here regardless of whether it always exists or not, these retry fields are a bit fiddly sometimes and this does seem better safe than sorry.

internal/activity.go
evansandoval: After running the retryactivity sample locally, the attempt number from the log does properly match the attempt retryAttempt number from the web UI

image

image

PR #1453, opened 2025-08-13
remove unnecessary debug level logging for autoscaler event

shijiesheng:
**What changed?**

Emit metrics logging happens every "cooldown time" (10s by default) and thus created a ton of debug logging. These logging are not needed.  

**Why?**

Some user complains about too much debug level logging

**How did you test it?**

unit test

**Potential risks**


taylanisikdemir (approved):

Groxx (approved):
yea, makes sense that'd be noisy 👍

PR #1461, opened 2025-10-02
feat: Introduce FailoverDomain endpoint from cadence-idl

gazi-yestemirova:
**What changed?**
**Detailed Description**
Introducing new FailoverDomain endpoint which will handle failover operations for domains. 

**Why?**
Extracting failover operations for domain updates, as it is a separate logic from UpdateDomain. In addition, UpdateDomain endpoint requires admin permissions to perform the operation, while failing over the domain should be done by anyone who has write permissions over the domain.

**How did you test it?**
Tested locally and unit tests.

**Potential risks**
As it is a new endpoint, no potential risks

**Impact Analysis**
- **Backward Compatibility**: N/A Since introducing a new endpoint
- **Forward Compatibility**: N/A

**Testing Plan**
- **Unit Tests**: We have unit tests for these changes
- **Persistence Tests**: N/A
- **Integration Tests**: tested with integration tests
- **Compatibility Tests**: N/A

**Rollout Plan**
- What is the rollout plan? Rollout wave by wave starting from staging


internal/compatibility/api_test.go
```diff
@@ -2166,6 +2166,7 @@ func TestTaskListStatus(t *testing.T) {
 			ExcludedFields: []string{
 				"NewTasksPerSecond",     // [BUG] NewTasksPerSecond is not mapped
 				"IsolationGroupMetrics", // [BUG] IsolationGroupMetrics is not mapped
+				"Empty",                 // Empty field is not mapped
```
c-warren: I'd recommend adding the reason why it isn't mapped here, otherwise in future it won't be obvious to someone changing the code. e.g 'Empty is only used in matching<>matching communication, and doesn't need to be mapped'

internal/compatibility/api_test.go
```diff
@@ -2454,6 +2455,24 @@ func TestWorkflowExecutionContinuedAsNewEventAttributes(t *testing.T) {
...
+					c.Fuzz(&attr.TaskStartToCloseTimeout)
+					c.Fuzz(&attr.DecisionTaskCompletedEventId)
+					c.Fuzz(&attr.BackoffStartInterval)
+					c.Fuzz(&attr.Initiator)
+					c.Fuzz(&attr.Failure)
+					c.Fuzz(&attr.LastCompletionResult)
+					c.Fuzz(&attr.Header)
+					c.Fuzz(&attr.Memo)
+					c.Fuzz(&attr.SearchAttributes)
+				},
```
c-warren, line 2475: I think it should be possible to just ignore the ActiveClusterSelectionPolicy field (or have a custom func for that) and then leave the rest as defaults. e.g

```golang
func(e apiv1.ActiveClusterSelectionPolicy, c fuzz.Continue) {
  // do nothing/create a default field
},
```

PR #1467, opened 2025-10-24
feat: add ResettableTimer for simplified timer management

Marpond:
**What changed?**
Introduces an abstraction of resetting timers via `ResettableTimer` in `x/resettabletimer`.

**Why?**
Simplifies common timer patterns where timers need to be updated or extended. Previously, we had to manually cancel the existing timer context, handle `CanceledError`, and create a new timer. With `ResettableTimer`, a single `Reset()` call handles this automatically.

**How did you test it?**
Added workflow unit tests to ensure functional correctness and deterministic replay behavior.

**Potential risks**
Low risk. Does not alter current timer behavior, only adds a user-friendly interface built on existing timer primitives.

### DIFF
--- a/x/resettabletimer/readme.md
+++ b/x/resettabletimer/readme.md
@@ -1,0 +1,236 @@
+### Resettable Timers in Cadence Workflows
+
+#### Status
+
+November 27, 2025
+
+This is experimental and the API may change in future releases.
+
+#### Background
+
+In Cadence workflows, timers are a fundamental building block for implementing timeouts and delays. However, standard timers cannot be reset once created - you must cancel the old timer and create a new one, which can lead to complex code patterns.
+
+The resettable timer provides a simple way to implement timeout patterns that need to restart based on external events.
+
+#### Getting Started
+
+Import the package:
+
+```go
+import (
+    "go.uber.org/cadence/workflow"
+    "go.uber.org/cadence/x/resettabletimer"
+)
+```
+
+#### Basic Usage
+
+Create a timer that can be reset:
+
+```go
+func MyWorkflow(ctx workflow.Context) error {
+    // Create a timer that fires after 30 seconds
+    timer := resettabletimer.New(ctx, 30*time.Second)
+    
+    // Wait for the timer
+    err := timer.Future.Get(ctx, nil)
+    if err != nil {
+        return err
+    }
+    
+    // Timer fired - handle timeout
+    workflow.GetLogger(ctx).Info("Timeout occurred")
+    return nil
+}
+```
+
+#### Resetting the Timer
+
+```go
+func MyWorkflow(ctx workflow.Context) error {
+    timer := resettabletimer.New(ctx, 30*time.Second)
+    userActivityChan := workflow.GetSignalChannel(ctx, "user_activity")
+    
+    selector := workflow.NewSelector(ctx)
+    
+    // Add timer to selector
+    selector.AddFuture(timer.Future, func(f workflow.Future) {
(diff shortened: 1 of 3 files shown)


internal/internal_time.go
```diff
@@ -47,4 +47,102 @@ type (
...
+		future      Future
+		settable    Settable
+		duration    time.Duration
+		isReady     bool
+	}
 )
+
+// NewResettableTimer creates a new resettable timer that fires after duration d.
+// The timer can be reset using Reset() to restart the countdown.
+func NewResettableTimer(ctx Context, d time.Duration) ResettableTimer {
```
jakobht: We should return a struct, generally we want to return structs and accept interfaces. This makes backwards compatability easier

internal/internal_workflow_test.go
```diff
@@ -1345,3 +1345,156 @@ func (t *WorkflowOptionTest) TestKnowQueryType_WithHandlers() {
...
+func (s *WorkflowUnitTest) Test_ResettableTimerFiresAfterExpiration() {
+	wf := func(ctx Context) error {
+		timer := NewResettableTimer(ctx, 5*time.Second)
+
+		timerFired := false
+		NewSelector(ctx).AddFuture(timer, func(f Future) {
+			err := f.Get(ctx, nil)
+			s.NoError(err)
+			timerFired = true
+		}).Select(ctx)
```
jakobht: Nit: we could check that 5 seconds actually passed (the cadence workflow test framework mocks time, so it should be fine)

internal/internal_time.go
```diff
@@ -47,4 +47,102 @@ type (
...
 	}
+
+	// ResettableTimer represents a timer that can be reset to restart its countdown.
+	ResettableTimer interface {
+		Future
+
+		// Reset - Cancels the current timer and starts a new one with the given duration.
+		// If duration is not provided, uses the previous duration.
+		// If the timer has already fired, Reset has no effect.
+		Reset(d ...time.Duration)
```
jakobht: We should not take a list if we want 0 or 1, I suggest just always taking the reset interval. 
```suggestion
		Reset(d time.Duration)
```

jakobht (commented):
Looks good. Added a few suggestions.

I think lets put this in the `x/resettabletimer` package since it's on top of the basic client

x/resettabletimer/readme.md
```diff
@@ -0,0 +1,225 @@
...
+    
+    selector.Select(ctx)
+    return nil
+}
+```
+
+#### Example: Inactivity Timeout with Dynamic Duration
+
+```go
+func InactivityTimeoutWorkflow(ctx workflow.Context) error {
```
jakobht, line 79: I like this example! Nice!

x/resettabletimer/readme.md
```diff
@@ -0,0 +1,225 @@
...
+    done := false
+    for !done {
+        selector := workflow.NewSelector(ctx)
+        
+        selector.AddFuture(timer.Future, func(f workflow.Future) {
+            workflow.GetLogger(ctx).Info("User inactive - logging out")
+            done = true
+        })
+        
+        selector.AddReceive(activityChan, func(c workflow.Channel, more bool) {
```
jakobht: Nit: activity already means something in Cadence - can we rename this to userActivity maybe?

jakobht (approved):

x/resettabletimer/readme.md
Marpond: Good idea, I've renamed it to userActivityChan to prevent confusion

PR #1469, opened 2025-10-29
feat: Expose ActiveClusterSelectionPolicy type

Shaddoll:
**Detailed Description**
Expose ActiveClusterSelectionPolicy type which was introduced in https://github.com/cadence-workflow/cadence-go-client/pull/1438

**Impact Analysis**
- **Backward Compatibility**: Yes
- **Forward Compatibility**: Yes

**Testing Plan**
- **Unit Tests**: Yes
- **Persistence Tests**: yes, in server repo
- **Integration Tests**: yes, in server repo

**Rollout Plan**
- What is the rollout plan? Create a new prerelease and try it in canary services.

- Does the order of deployment matter? Server changes must be deployed first before clients start using this.
- Is it safe to rollback? Does the order of rollback matter? Yes before users start taking dependency on new field.
- Is there a kill switch to mitigate the impact immediately? Rolling back client is safe and should mitigate unforeseen issues. Server side changes are already in.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -93,6 +93,15 @@ type (
 	QueryConsistencyLevel = internal.QueryConsistencyLevel
 
 	// ActiveClusterSelectionPolicy - Policy for selecting the active cluster to start the workflow execution on for active-active domains.
+	// An example policy is:
+	//   {
+	//     ClusterAttribute: {
+	//       Scope: "region",
+	//       Name: "us-west",
+	//     }
+	//   }
+	// The workflow will be active in the corresponding active cluster of that cluster attribute specified in the policy
+	// If the policy is empty or nil, the workflow will be active in the domain's active cluster.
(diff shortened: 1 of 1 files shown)


davidporter-id-au (approved):

client/client.go
```diff
@@ -92,6 +92,9 @@ type (
 	// Valid values are QueryConsistencyLevelEventual (served by the receiving cluster), and QueryConsistencyLevelStrong (redirects to the active cluster).
 	QueryConsistencyLevel = internal.QueryConsistencyLevel
 
+	// ActiveClusterSelectionPolicy - Policy for selecting the active cluster to start the workflow execution on for active-active domains.
```
davidporter-id-au, line 95: An example here for the user using some example data would probably be good

davidporter-id-au (approved):

PR #1471, opened 2025-10-29
chore: remove self from CODEOWNERS

Groxx:
Just reducing noise from review-notifications :saluting_face:


davidporter-id-au (approved):

davidporter-id-au:
🫡

PR #1478, opened 2025-11-24
fix: missing fields for FailoverDomain API call 

davidporter-id-au:
**What changed?**
The FailoverDomain endpoint is missing a few fields for cluster attributes. This maps them and adds a few tests to cover this. More problematic, is the response struct which, to be honest, I didn't realise was returning the entire domain data. So this also updates / maps that. 

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/compatibility/api_test.go
+++ b/internal/compatibility/api_test.go
@@ -476,6 +476,137 @@ func TestDeprecateDomainRequest(t *testing.T) {
 		FuzzOptions{},
 	)
 }
+
+func TestFailoverDomainRequest(t *testing.T) {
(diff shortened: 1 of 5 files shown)


internal/compatibility/api_test.go
```diff
@@ -3063,6 +3194,13 @@ func clearFieldsIf(obj interface{}, shouldClear func(fieldName string) bool) {
 						clearFieldsIf(elem.Interface(), shouldClear)
 					}
 				}
+			case reflect.Map:
+				for _, key := range field.MapKeys() {
```
davidporter-id-au, line 3198: needed to add this because it was missing few fields

PR #1482, opened 2025-12-24
fix: update WorkflowInfo ExecutionStartToCloseTimeoutSeconds to match TestWorkflowEnvironment timeout

mkamilli:
**What changed?**

The workflow relies on `ExecutionStartToCloseTimeoutSeconds` ([see usage](https://github.com/cadence-workflow/cadence/blob/66a86e0107bf0ce1a80ddeebd45ea58941be62eb/bench/load/timer/launchWorkflow.go#L65)). 

With this change I am updating test `WorkflowInfo`'s `ExecutionStartToCloseTimeoutSeconds` to match the execution timeout specified to the `TestWorkflowEnvironment`

Relevant issue:
https://github.com/cadence-workflow/cadence/issues/7563

**Why?**

If it is not done then `ExecutionStartToCloseTimeoutSeconds` is [always set to 1](https://github.com/cadence-workflow/cadence-go-client/blob/f0ff1c04a44f4200a5e00c2fc470f38f1cbfe6a0/internal/internal_workflow_testsuite.go#L253).

**How did you test it?**

**Potential risks**

### DIFF
--- a/internal/workflow_testsuite.go
+++ b/internal/workflow_testsuite.go
@@ -478,6 +478,7 @@ func (t *TestWorkflowEnvironment) SetTestTimeout(idleTimeout time.Duration) *Tes
 // This is based on the workflow time (a.k.a workflow.Now() time).
 func (t *TestWorkflowEnvironment) SetWorkflowTimeout(executionTimeout time.Duration) *TestWorkflowEnvironment {
 	t.impl.executionTimeout = executionTimeout
+	t.impl.workflowInfo.ExecutionStartToCloseTimeoutSeconds = int32(executionTimeout.Seconds())
 	return t
 }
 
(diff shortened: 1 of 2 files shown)


natemort:
Do you mind adding a unit test to confirm this functionality? While it's a simple change we want to make sure that other changes in the future don't accidentally cause a regression

mkamilli:
> Do you mind adding a unit test to confirm this functionality? While it's a simple change we want to make sure that other changes in the future don't accidentally cause a regression

@natemort added unit test https://github.com/cadence-workflow/cadence-go-client/commit/10aece77f907a1af4468714c52f6696eb16c8924

natemort (approved):
LGTM, you'll need to change the PR description to match the conventional commits requirements.

PR #1484, opened 2026-01-04
fix: do not terminate worker on non-retryable poll errors

### LINKED ISSUE #7561: Remove Unretryable Error Logic
**Is your feature request related to a problem? Please describe.**
This is specific to the Go client, but posting here to centralize issues more.

If the server returns a NonRetryableError the worker tries to terminate its own process. This is bad.

**Proposed Solution**
Remove [this logic](https://github.com/cadence-workflow/cadence-go-client/blob/f0ff1c04a44f4200a5e00c2fc470f38f1cbfe6a0/internal/internal_worker_base.go#L340) entirely. We should log quite clearly that we're receiving some critical error, but we can just keep trying the poll requests. Invalid requests are not a serious risk to either the server or the client. We should never be terminating the process.

**Additional context**

MPAdler:
**What changed?**
**Related:** cadence-workflow/cadence#7561
Updated baseWorker.pollTask() in internal/internal_worker_base.go addressing request to remove unretryable error logic. Now the worker logs the error and continues polling instead of self-terminating. 
Added a unit test in internal_worker_base_test.go to test error behavior.

**Why?**
Fixes cadence-workflow/cadence#7561

**How did you test it?**
Uses cadence-go-client make unit_test on macOS (exit code 0).

**Potential risks**
The worker no longer exits on non-retryable poll errors, so it may continue running and repeatedly backoff. Backoff is preserved to prevent tight-looping, but log volume could explode if the underlying problem persists.

### DIFF
--- a/internal/internal_worker_base.go
+++ b/internal/internal_worker_base.go
@@ -27,9 +27,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
 	"sync"
-	"syscall"
 	"time"
 
 	"go.uber.org/cadence/internal/common/debug"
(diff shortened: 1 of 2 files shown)


natemort:
I'm open to accepting this without any test coverage, but if you're able to add some then that would be great.

You just need to update the PR title, otherwise LGTM.

MPAdler:
Sounds good — I’ll add a targeted unit test in `internal_worker_base_test.go` to cover the non-retryable PollTask error path.

MPAdler:
Added a unit test as suggested. Happy to adjust if needed, thanks!

neil-xie (approved):

natemort:
Thank you for your contribution!

PR #1489, opened 2026-02-25
feat: Dual emitting timer and histogram metrics

neil-xie:
**What changed?**
Start dual emitting histogram metrics as part of timer -> histogram metrics migration
Added migration doc to example the changes

**Why?**
Time -> histogram migration 
https://github.com/cadence-workflow/cadence/issues/7741

**How did you test it?**
Unit test

**Potential risks**
Backward compatibility, only risk is the metrics storage usage increase


internal/common/metrics/histograms.go
```diff
@@ -0,0 +1,216 @@
...
+	return s.tallyBuckets
+}
+
+// RecordTimer is a convenience helper to record a duration to a histogram.
+// This makes it easier to migrate from Timer to Histogram.
+//
+// Example:
+//
+//	RecordTimer(scope, "my-operation-latency", duration, Default1ms100s)
+func RecordTimer(scope tally.Scope, name string, duration time.Duration, buckets SubsettableHistogram) {
```
timl3136: nit: perhaps rename this to recordTimerAsHistogram since we are not recording a timer here.

internal/common/metrics/histograms.go
```diff
@@ -0,0 +1,216 @@
...
+
+// StartTimer returns a stopwatch for timing operations with a histogram.
+// Call .Stop() on the returned stopwatch to record the duration.
+//
+// Example:
+//
+//	sw := StartTimer(scope, "my-operation-latency", Default1ms100s)
+//	// ... do work ...
+//	sw.Stop()
+func StartTimer(scope tally.Scope, name string, buckets SubsettableHistogram) tally.Stopwatch {
```
timl3136: same here, perhaps rename to startTimerAsHistogram

internal/common/metrics/emit.go
```diff
@@ -0,0 +1,83 @@
...
+// Parameters:
+//   - scope: The tally scope to emit metrics to
+//   - name: The metric name (without suffix)
+//   - latency: The duration to record
+//   - buckets: The histogram bucket configuration to use
+//
+// Example:
+//
+//	EmitLatency(scope, "decision-poll-latency", duration, Default1ms100s)
+func EmitLatency(scope tally.Scope, name string, latency time.Duration, buckets SubsettableHistogram) {
```
timl3136, line 101: Can we introduce a flag to control the dual emit logic instead of needing to actually change the function when we change from dual emit to only histogram so that it's easier to control without needing a PR/version update?

neil-xie: I thought about that, but seems no easy way to use a flag to control since it is in the workers, not on the server side.

shijiesheng: You can do it in FeatureFlags.

neil-xie: Seems they are optional in worker config, we couldn't force the user to enable. 
Oh, maybe we can enable by default and let OSS user know they can turn it off through FeatureFlag, is this what you think?

internal/common/metrics/histograms.go
neil-xie: claude insists recordHistogram

internal/common/metrics/emit.go
neil-xie: Added MetricEmitMode to feature flags, it is dual emitting by default, I will mention in the release note in OSS so they can choose changing it to timer only, or migrating with us

internal/common/metrics/MIGRATION.md
```diff
@@ -0,0 +1,164 @@
+# Timer to Histogram Migration Guide
```
davidporter-id-au, line 1: Having docs is good, but I'd consider moving this to somewhere more easily found, this is deep inside `internal`. I'd put it in the base in a docs folder

internal/common/metrics/MIGRATION.md
```diff
@@ -0,0 +1,164 @@
...
+
+**Cardinality:** +62 histogram metrics (temporary during migration)
+**Performance:** Minimal impact
+**Compatibility:** 100% backward compatible
+
+## Why Migrate?
+
+1. **Better precision control** - Exponential buckets: fine detail at low values, coarse at high
+2. **OTEL compatible** - OpenTelemetry exponential histogram specification
+3. **Cardinality control** - Can reduce resolution with `subsetTo()`
```
davidporter-id-au, line 39: do we have a link to this or a reference with a bit more of an explanation?

davidporter-id-au (approved):
The documentation and migration details are a really nice touch

internal/common/metrics/MIGRATION.md
neil-xie: Yeah, I will test the code and move it to docs folder if it is working for us. We are going to make this optional for OSS users

internal/common/metrics/MIGRATION.md
neil-xie: We have explanation in the histograms.go 
```
// subsetTo reduces the histogram's detail level by reducing the scale.
// This halves the number of buckets for each scale reduction.
//
// Example: If you have a scale=2 histogram with 80 buckets,
// subsetTo(1) will produce a scale=1 histogram with 40 buckets.
```

internal/internal_task_pollers.go
```diff
@@ -706,7 +716,12 @@ func (lath *localActivityTaskHandler) executeLocalActivityTask(task *localActivi
 		defer span.Finish()
```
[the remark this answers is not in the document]
neil-xie: local activities can be long running, but usually short/light work

PR #1490, opened 2026-02-25
feat(auto-forwarding): add feature flag

shijiesheng:
**What changed?**

Allow workers to autoforward to active domain region for test workers

* IDL update 
* add featureflag to enable autoforwarding

server change https://github.com/cadence-workflow/cadence/commit/63d68a238a626213ffe2f73aa4bc1cbba91f5f57 

**Why?**

Ref https://github.com/cadence-workflow/cadence/issues/7707 

**How did you test it?**

Unit Test

Tested in cadence-samples (active in DCA) in production
With autoforwarding, workers that are connected to PHX cluster are able to process workflows and the pollers are visible in the tasklist information. 

**Potential risks**


internal/compatibility/proto/request.go
```diff
@@ -700,6 +700,7 @@ func FailoverDomainRequest(t *shared.FailoverDomainRequest) *apiv1.FailoverDomai
 	}
 	request := &apiv1.FailoverDomainRequest{
 		DomainName: t.GetDomainName(),
+		Reason:     t.GetReason(),
```
fimanishi, line 703: This is not related to the change necessarily, right? But it's something that we should have?

internal/internal_utils.go
```diff
@@ -94,29 +97,22 @@ var (
 	}
 )
 
-func fromInternalFeatureFlags(featureFlags FeatureFlags) s.FeatureFlags {
+func fromInternalFeatureFlags(featureFlags FeatureFlags) apiv1.FeatureFlags {
```
fimanishi, line 100: What's the reason to change from "go.uber.org/cadence/.gen/go/shared" FeatureFlags to "github.com/uber/cadence-idl/go/proto/api/v1"?

fimanishi (approved):

internal/compatibility/proto/request.go
shijiesheng: It's due to the IDL update

internal/internal_utils.go
shijiesheng: FeatureFlags previously were serialized through thrift IDL, which is known to be bad in consistency across languages. This is to use GRPC with gogoprotocol's json encoder/decoder.

PR #1496, opened 2026-04-14
chore(blocks)!: remove experimental x/blocks package

Bueller87:
**What changed?**

- Removed the experimental `x/blocks` package (`blocks.go`, `blocks_test.go`, `readme.md`).

**Why?**

- Cadence Web now treats markdown/Markdoc query responses (`formattedData` + `text/markdown` + `data`) as the supported path for interactive query UIs.
- The blocks JSON format duplicated UI concerns across SDKs and was harder to author and read than markdown. Dropping the Go helper package avoids maintaining a parallel API in the client.

**How did you test it?**

- `go build ./...`

**Potential risks**

- **Breaking change:** Any workflow or library that imported `go.uber.org/cadence/x/blocks` will fail to compile until it returns a serializable value with `cadenceResponseType: "formattedData"`, `format: "text/markdown"`, and `data` set to the markdown string (see cadence-samples `new_samples/query`).

Fixes cadence-workflow/cadence-web#1141


shijiesheng (commented):
I think title should be "chore(blocks)!:" but I've stamped.

shijiesheng (approved):

shijiesheng (approved):

PR #1502, opened 2026-05-13
feat(schedule)!: Add wire protocol infra for 8 schedule RPCs

YaweiZhang-930:
**What changed?**
This is part 1 of 2 for the Go-native ScheduleClient SDK. It wires up the wire-protocol layer — no user-facing types or client interface yet (those are in PR 2).

- IDL submodule updated to include 8 schedule frontend RPCs: CreateSchedule, DescribeSchedule, UpdateSchedule, DeleteSchedule, PauseSchedule, UnpauseSchedule, BackfillSchedule, ListSchedules

- Generated code (.gen/go/shared/, .gen/go/cadence/) regenerated with schedule Thrift structs (ScheduleSpec, ScheduleAction, SchedulePolicies, etc.) and updated workflowserviceclient.Interface
- Thrift to Proto converters added for the gRPC adapter path
 - thrift2protoAdapter extended with 8 schedule methods
 **- NewThrift2ProtoAdapter() signature replaced with a struct parameter AdapterClients — breaking change to public API, but forward-compatible forever (futureservices are new struct fields, not new positional parameters)**
- Auth, metrics, and isolation group service wrappers each get 8 passthrough/interceptor methods to satisfy the updated workflowserviceclient.Interface
- new tests

**Why?**

-  The Cadence server already supports these 8 schedule RPCs. 
- This PR connects the generated Thrift stubs through the gRPC compatibility adapter and all existing cross-cutting service wrappers, so that PR 2 can build a clean, proto-free ScheduleClient on top without any infrastructure changes.

**How did you test it?**

- make build passes — all packages compile, goimports is idempotent, no git drift after build
- go test ./internal/compatibility/... — 6 new adapter round-trip tests for ScheduleSpec, SchedulePauseInfo, ScheduleState, SchedulePolicies, ScheduleStartWorkflowAction, and BackfillInfo all pass
- go test ./... — no regressions in existing tests

**Potential risks**

- NewThrift2ProtoAdapter() signature changed — the four positional YARPC client parameters are replaced by a single AdapterClients struct. **This is a breaking change.**
- 8 new methods on workflowserviceclient.Interface — all existing wrapper implementations (auth, metrics, isolation group, adapter, generated test mock) are updated here. 
- The 8 new wrapper methods are passthrough stubs that delegate directly to the underlying service — they do not alter existing RPC behavior.


internal/compatibility/api_test.go
```diff
@@ -2220,6 +2220,237 @@ func TestSupportedClientVersions(t *testing.T) {
...
+func TestScheduleListEntry(t *testing.T) {
+	for _, item := range []*apiv1.ScheduleListEntry{nil, {}, &testdata.ScheduleListEntry} {
+		assert.Equal(t, item, proto.ScheduleListEntry(thrift.ScheduleListEntry(item)))
+	}
+}
+func TestScheduleAction(t *testing.T) {
+	for _, item := range []*apiv1.ScheduleAction{nil, {}, &testdata.ScheduleAction} {
+		assert.Equal(t, item, proto.ScheduleAction(thrift.ScheduleAction(item)))
+	}
+}
```
abhishekj720, line 2353: Above test all call runFuzzTest but TestScheduleInfo, TestScheduleListEntry, and TestScheduleAction don't. ScheduleInfo has a []*BackfillInfo slice and five timestamp fields; ScheduleListEntry and ScheduleAction have nested structs. These are the exact cases where fuzzing catches nil/empty slice distinction bugs and nested conversion errors. Could you add runFuzzTest for these three?

internal/compatibility/api_test.go
```diff
@@ -2220,6 +2220,237 @@ func TestSupportedClientVersions(t *testing.T) {
...
+	result := proto.UnpauseScheduleRequest(&shared.UnpauseScheduleRequest{
+		Domain:     &domain,
+		ScheduleId: &id,
+		Reason:     &reason,
+	})
+	assert.Equal(t, domain, result.Domain)
+	assert.Equal(t, id, result.ScheduleId)
+	assert.Equal(t, reason, result.Reason)
+}
+func TestBackfillScheduleRequest(t *testing.T) {
```
abhishekj720, line 2444: BackfillScheduleRequest maps 6 fields but the test only asserts 3 of them. StartTimeNano, EndTimeNano, and OverlapPolicy are the non-trivial ones (timestamp conversion + enum mapping), those are exactly the ones worth testing. Could you populate and assert those three fields as well?

internal/compatibility/api_test.go
```diff
@@ -2220,6 +2220,237 @@ func TestSupportedClientVersions(t *testing.T) {
...
+}
+func TestDeleteScheduleRequest(t *testing.T) {
+	assert.Nil(t, proto.DeleteScheduleRequest(nil))
+	assert.NotNil(t, proto.DeleteScheduleRequest(&shared.DeleteScheduleRequest{}))
+	domain, id := "test-domain", "my-schedule"
+	result := proto.DeleteScheduleRequest(&shared.DeleteScheduleRequest{Domain: &domain, ScheduleId: &id})
+	assert.Equal(t, domain, result.Domain)
+	assert.Equal(t, id, result.ScheduleId)
+}
+func TestPauseScheduleRequest(t *testing.T) {
```
abhishekj720, line 2413: PauseScheduleRequest maps an Identity field in proto/schedule.go. Can we assert it here?

internal/compatibility/api_test.go
```diff
@@ -2220,6 +2220,237 @@ func TestSupportedClientVersions(t *testing.T) {
...
+}
+
+// Response converter tests (proto → thrift, one-way)
+
+func TestCreateScheduleResponse(t *testing.T) {
+	assert.Nil(t, thrift.CreateScheduleResponse(nil))
+	result := thrift.CreateScheduleResponse(&apiv1.CreateScheduleResponse{ScheduleId: "my-id"})
+	assert.Equal(t, "my-id", result.GetScheduleId())
+}
+func TestDescribeScheduleResponse(t *testing.T) {
```
abhishekj720, line 2485: It only checks Nil/NotNil. 
DescribeScheduleResponse converts 7 fields (spec, action, policies, state, info, memo, searchAttributes) through the full thrift converter chain, none of those field mappings are exercised by this test. 
Could you populate a DescribeScheduleResponse with testdata fields and assert the key ones round-trip correctly?

internal/compatibility/thrift/schedule.go
```diff
@@ -0,0 +1,250 @@
...
+	return &shared.PauseScheduleResponse{}
+}
+
+func UnpauseScheduleResponse(_ *apiv1.UnpauseScheduleResponse) *shared.UnpauseScheduleResponse {
+	return &shared.UnpauseScheduleResponse{}
+}
+
+func BackfillScheduleResponse(_ *apiv1.BackfillScheduleResponse) *shared.BackfillScheduleResponse {
+	return &shared.BackfillScheduleResponse{}
+}
```
abhishekj720: These converters ignore their argument entirely and always return a non-nil empty struct, even when called with nil. 
The fix is a standard nil guard maybe something like:

```
func UpdateScheduleResponse(t *apiv1.UpdateScheduleResponse) *shared.UpdateScheduleResponse {
    if t == nil {
        return nil
     }
    return &shared.UpdateScheduleResponse{}
 }

```

internal/compatibility/api_test.go
YaweiZhang-930: Done

internal/compatibility/api_test.go
YaweiZhang-930: done

internal/compatibility/api_test.go
YaweiZhang-930: Done

internal/compatibility/api_test.go
YaweiZhang-930: Done

internal/common/auth/service_wrapper.go
```diff
@@ -541,3 +541,75 @@ func (w *workflowServiceAuthWrapper) ListFailoverHistory(ctx context.Context, re
 	result, err := w.service.ListFailoverHistory(ctx, request, opts...)
 	return result, err
 }
+
+func (w *workflowServiceAuthWrapper) BackfillSchedule(ctx context.Context, Request *shared.BackfillScheduleRequest, opts ...yarpc.CallOption) (*shared.BackfillScheduleResponse, error) {
```
shijiesheng, line 545: nit: use lower case. But can be ignored

shijiesheng (approved):

PR #1505, opened 2026-05-18
feat(schedule)!: Add Go-native ScheduleClient SDK

YaweiZhang-930:
**Summary**   
This is part 2 of a two-part implementation:
- Part 1 (#1502 (https://github.com/cadence-workflow/cadence-go-client/pull/1502): wire protocol infrastructure (IDL, generated Thrift/gRPC, adapters)
- Part 2 (this PR): Go-native SDK types, ScheduleClient interface, and thrift↔SDK converters

**Detailed Description**

- The Client interface gains one new method: **ScheduleClient() ScheduleClient**
- Notable design decisions:
1. PauseOnFailure is *bool: nil leaves server's value unchanged on Update; sending false would silently overwrite true.
2. Enum Unspecified → nil on wire: field is omitted rather than sending &Invalid, so the server's configured policy is not disturbed.
3. Unpause takes a flat catchUpPolicy param: Unspecified omits the field; a specific value overrides for this unpause only.
4. Create is not idempotent: duplicate returns BadRequestError; callers should Describe after failure.

**Impact Analysis**

- Backward Compatibility: Breaking — NewScheduleClient() is added to the Client interface. Any custom implementation of Client (e.g. hand-written mocks) must add this method. The generated mocks/Client.go is updated. No existing method signatures change.

- Forward Compatibility: Non-breaking. All ScheduleClient usage is opt-in. The *bool and nil-on-wire patterns ensure unset fields never overwrite server state.

  
**Testing Plan**

- Unit Tests: Yes — internal_schedule_client_test.go covers all 8 methods (validation, happy path, RPC failure). internal_schedule_convert_thrift_test.go covers enum round-trips, nil guards, and struct converters.
- Persistence Tests: Not applicable — schedule state is entirely server-side; the SDK holds no persistent state.
- Integration Tests: YES! Added in this PR
- Compatibility Tests: TestScheduleOverlapPolicyRoundTrip and TestScheduleCatchUpPolicyRoundTrip verify all known values round-trip cleanly through SDK↔Thrift↔SDK.

**Rollout Plan**

- Client-side only change; the server already supports all 8 RPCs. No server deployment needed.
- Deployment order does not matter — SDK upgrade is independent of any server feature flag.
- Rollback is safe and order-independent; reverting the SDK removes NewScheduleClient() with no server-side impact.
- No kill switch needed — all schedule functionality is opt-in via explicit NewScheduleClient() calls.


YaweiZhang-930:
Gitar fix CI

internal/internal_schedule_client.go
```diff
@@ -0,0 +1,226 @@
...
+		workflowService: wc.workflowService,
+		domain:          wc.domain,
+		identity:        wc.identity,
+		dataConverter:   wc.dataConverter,
+		featureFlags:    wc.featureFlags,
+	}
+}
+
+func (sc *scheduleClient) Create(ctx context.Context, request *CreateScheduleRequest) (string, error) {
+	thriftReq, err := scheduleCreateRequestToThrift(sc.domain, request, sc.dataConverter)
```
shijiesheng, line 91: nit: validate requested action payload (start workflow requires some fields like tasklist, domain, and other fields)

internal/internal_schedule_convert_thrift.go
```diff
@@ -0,0 +1,538 @@
...
+	}
+	if r.ScheduleID == "" {
+		return nil, errors.New("Create: ScheduleID is required")
+	}
+	if r.Spec == nil || r.Spec.CronExpression == "" {
+		return nil, errors.New("Create: Spec.CronExpression is required")
+	}
+	if r.Action == nil || r.Action.StartWorkflow == nil {
+		return nil, errors.New("Create: Action.StartWorkflow is required")
+	}
```
shijiesheng, line 280: For create, update and backfill, validate the payload in the client layer rather than in this converter.

internal/schedule.go
```diff
@@ -0,0 +1,236 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package internal
+
+// This file defines all Go-native types for the schedule API. It is intentionally proto-free:
+// no proto imports appear here so that callers never encounter generated types at the API boundary.
+// Conversion to/from proto wire types happens entirely inside internal_schedule_client.go.
```
abhishekj720: Nit: schedule work actually goes through internal_schedule_convert_thrift.go (Thrift), right?

internal/schedule.go
```diff
@@ -0,0 +1,236 @@
...
+	// Memo is metadata attached to the schedule itself (not to each triggered run).
+	// Values are encoded using the DataConverter configured on the ScheduleClient. Optional.
+	Memo map[string]interface{}
+	// SearchAttributes are indexed attributes on the schedule itself. Optional.
+	// Values are JSON-encoded.
+	SearchAttributes map[string]interface{}
+}
+
+// UpdateScheduleRequest is the request to ScheduleClient.Update.
+// Only non-nil fields are applied; nil fields leave the existing value unchanged.
```
abhishekj720: Nit: when Policies is non-nil the server replaces the full policy blob, and BufferLimit/ConcurrencyLimit are plain int32, so a “partial” Go struct can unintentionally reset limits.

internal/internal_schedule_client.go
YaweiZhang-930: Done added some validation

internal/internal_schedule_convert_thrift.go
YaweiZhang-930: These are private helpers only called from the client, so it's already effectively in the client layer. Moving it out means a shared validateStartWorkflowAction since Create and Update both go through the same path, and we'd need to keep the validator and converter in sync when fields change. Happy to change if you feel strongly

internal/schedule.go
YaweiZhang-930: Good catch! updated

internal/schedule.go
YaweiZhang-930: I just update the comments here

internal/internal_schedule_convert_thrift.go
shijiesheng: make sense

shijiesheng (approved):

abhishekj720 (approved):

PR #1509, opened 2026-06-08
feat(schedule): return action Memo and SearchAttributes from DescribeSchedule

YaweiZhang-930:
**Detailed Description**
This change makes DescribeSchedule return the start-workflow action's Memo and SearchAttributes, which were previously dropped on the read path. It does so by splitting the schedule action into separate request (write) and response (read) types, since the two directions need different representations:
 
- New types (read-side):
    - ScheduleStartWorkflowActionDescription — same fields as ScheduleStartWorkflowAction (WorkflowType, TaskList, Input, WorkflowIDPrefix, both timeouts, RetryPolicy) but with Memo map[string][]byte and SearchAttributes map[string][]byte (raw encoded bytes, as stored by the server).
    - ScheduleActionDescription — wraps StartWorkflow *ScheduleStartWorkflowActionDescription.
    - Both re-exported from the client package.
  - Modified data structure: DescribeScheduleResponse.Action changes type from *ScheduleAction to *ScheduleActionDescription.
  - Behavioral change on read: action Memo/SearchAttributes are now populated (previously omitted entirely) and typed as raw map[string][]byte.
  - Unchanged: the request type ScheduleStartWorkflowAction keeps Memo/SearchAttributes as map[string]interface{} (the SDK encodes them on write with the configured DataConverter). No request fields added or removed. No IDL/wire change — thrift and proto messages are identical; this is purely the SDK's Go-type surface and read-side converter.

**Why?**
- The old read path had no DataConverter and the action type used map[string]interface{}, so action-level Memo/SearchAttributes were silently omitted on Describe. 
- **That meant callers couldn't read them back, and any read-modify-write flow (the planned Describe-then-Update for the server's full-replacement UpdateSchedule) would wipe them on every update. This makes Describe faithful to what the server actually returns — opaque pass-through, no decode on read — consistent with the schedule-level Memo/SearchAttributes, which were already returned as raw bytes.**

**Impact Analysis**
- **Backward Compatibility**: 
    1. Wire: fully compatible — the thrift/proto schedule messages are unchanged, so a client built with this change interoperates with any server version (the server already returns these bytes; we are only surfacing them).
    2. Go API (source): breaking — code that references DescribeScheduleResponse.Action as *ScheduleAction, or reads action Memo/SearchAttributes as map[string]interface{}, will no longer compile and must move to *ScheduleActionDescription / map[string][]byte. Mitigated by this being a pre-release feature (the schedule SDK landed in #1505 and is not in a released tag); there are no external consumers.
- **Forward Compatibility**:
   1. The request/response type split also makes the read and write surfaces independently evolvable going forward (e.g., adding read-only fields to the response without touching the request type).

**Testing Plan**
- **Unit Tests**: Yes. internal/internal_schedule_convert_thrift_test.go covers the converter round-trip and explicitly asserts the action's Memo/SearchAttributes are returned as the exact raw bytes the server stored (direct regression for the previously-dropped fields). internal/internal_schedule_client_test.go updated to the new response types.
- **Persistence Tests**: N/A.
- **Integration Tests**: test/schedule_integration_test.go (create → describe lifecycle) updated to the new types and exercises the real read path against a live server.

**Rollout Plan**
- What is the rollout plan? publish a new version 
- Is it safe to rollback? Does the order of rollback matter? Yes, safe. Rollback = revert the commit


abhishekj720 (approved):

YaweiZhang-930:
Gitar fix CI

PR #1519, opened 2026-07-13
feat: Add FailureOptions to CustomError

natemort:
Add support for specifying a FailureCategory and NextRetryDelay on CustomError instances. These allow activity implementations more granular control over retry behavior (whether it retries and the delay until the retry), as well as the ability for polling-style Activities to suppress error messages from expected failures.

Update the idls module to pull in the changes introducing these new fields and update the corresponding mappers. This is necessary due to the fuzzer tests. Fields must be either mapped or exempted.

**What changed?**
- Add support for specifying FailureOptions on CustomError instances returned by activities.
- Update the proto/thrift mappers for newly introduced fields
  - FailureOptions
  - New fields in ScheduleInfo
  - Last heartbeat in RespondActivityFailed

**Why?**
- Allow activities to override the RetryPolicy for specific failures.

**How did you test it?**
- Unit tests

**Potential risks**
- Incorrect behavior could change the retries for activities.
- In general, failures in activity processing would cause activities to time out.

### DIFF
--- a/error.go
+++ b/error.go
@@ -21,6 +21,8 @@
 package cadence
 
 import (
+	"time"
+
 	"go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/internal"
 	"go.uber.org/cadence/workflow"
@@ -29,6 +31,8 @@ import (
 type (
 	// CustomError returned from workflow and activity implementations with reason and optional details.
 	CustomError = internal.CustomError
+
+	CustomErrorOption = internal.CustomErrorOption
 
 	// CanceledError returned when operation was canceled.
 	CanceledError = internal.CanceledError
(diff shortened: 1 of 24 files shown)


internal/activity_task_handler.go
```diff
@@ -205,6 +205,11 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 	return convertActivityResultToRespondRequest(ath.identity, t.TaskToken, output, err, ath.dataConverter), nil
 }
 
+func isPollError(err error) bool {
```
shijiesheng: Could you add a comment here about what a poll error is? It's not a poller error and might be misunderstood.

internal/error.go
```diff
@@ -220,8 +226,34 @@ func NewCustomError(reason string, details ...interface{}) *CustomError {
...
+	return customError
+}
+
+func WithFailureCategory(failureCategory shared.FailureCategory) CustomErrorOption {
+	return func(customError *CustomError) {
+		customError.failureCategory = failureCategory.Ptr()
+	}
+}
+
+func WithNextRetryDelay(delay time.Duration) CustomErrorOption {
```
shijiesheng, line 253: Is it ok to let users to set nextRetryDelay without setting failure category?

Additionally, do you think it's easier to combine the two with one single method `WithFailureOptions()`

error.go
```diff
@@ -46,6 +50,14 @@ func NewCustomError(reason string, details ...interface{}) *CustomError {
 	return internal.NewCustomError(reason, details...)
 }
 
+func WithFailureCategory(failureCategory shared.FailureCategory) CustomErrorOption {
+	return internal.WithFailureCategory(failureCategory)
+}
+
+func WithNextRetryDelay(delay time.Duration) CustomErrorOption {
```
shijiesheng, line 64: Add comments on these new APIs

internal/error.go
natemort: The two fields are independent. I think having individual methods for these options provides the best discoverability and documentation compared to introducing a new FailureOptions type and a single method.

error.go
natemort: Done.

internal/activity_task_handler.go
```diff
@@ -205,6 +205,12 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 	return convertActivityResultToRespondRequest(ath.identity, t.TaskToken, output, err, ath.dataConverter), nil
 }
 
+// errors with a FailureCategory of Poll indicate an expected failure. We don't log them as an activity error
+func isFailureCategoryPoll(err error) bool {
```
natemort, line 215: Renamed to something more explicit

shijiesheng (approved):

PR #1520, opened 2026-07-15
feat: Add Activity lifecycle logging

### LINKED ISSUE #7583: Verbose Activity Logging
**Is your feature request related to a problem? Please describe.**
When debugging issues with Activities it is incredibly helpful to have a concrete record of when a given activity starts and stops. This can differ from the values in the workflow history, especially in two scenarios:
1. If the activity has retries (which it should) then it only keeps a record of the last attempt. This makes it very difficult to debug repeated failures.
2. If the activity runs longer than its timeout then its actual resource usage and behavior on the client is difficult to track. The workflow history is going to show it failing according to the timeout.

If the activity doesn't emit any logs then there's no record of how long it ran for.

**Proposed Solution**
Add a boolean option to the WorkerOptions to enable verbose activity logging. It should be disabled by default.

When enabled, it should log a message on the start and end of each activity attempt. It should have all of the relevant tags for the activity so that it's clear exactly which activity. Ideally the logs should be explicit about the outcome of the activity (success/failure and failure to report).

This can be implemented in both the Java and Go clients.

**Additional context**

joannalauu:
**What changed?**
New `EnableWorkflowLifecycleLogging` option in `WorkerOptions`, defaulted to `false`. When on, the activity worker logs `Activity started` and `Activity completed` for every activity attempt, with an `ActivityOutcome` field of succeeded / failed / canceled / pending / failed_to_report / timeout.

Fixes: https://github.com/cadence-workflow/cadence/issues/7583

**Why?**
- Workflow history keeps only the last attempt of a retried activity, so repeated failures leave no per-attempt record to debug from
- When activity runs longer than its timeout, workflow history only shows that it fails according to timeout, without logs tracking actual resource usage and behavior on the client

**How did you test it?**
- Unit test: `go test ./internal/ -run 'TestActivityOutcome'`

**Potential risks**
- Writes 2 extra `Info` lines to logs per activity attempt

---
**Detailed Description**

New `EnableActivityLifecycleLogging` option in `WorkerOptions`, defaulted to `false`. When on, the activity worker logs `Activity started` and `Activity completed` for every activity attempt, with an `ActivityOutcome` field of succeeded / failed / canceled / pending / failed_to_report / timeout.

**Impact Analysis**
- **Backward Compatibility**: Existing logs are unaffected
- **Forward Compatibility**: Existing logs do not affect the new version of logs

**Testing Plan**
- **Unit Tests**: [Do we have unit test covering the change?] `go test ./internal/ -run 'TestActivityOutcome'`

**Rollout Plan**
- What is the rollout plan?
---

### DIFF
--- a/internal/activity_task_handler.go
+++ b/internal/activity_task_handler.go
@@ -233,6 +233,8 @@ func (ath *activityTaskHandlerImpl) getRegisteredActivityNames() (activityNames 
 // activityOutcome classifies the values Execute returns
 func activityOutcome(result interface{}, err error) string {
 	switch {
+	case err == context.DeadlineExceeded:
+		return "timeout"
 	case err != nil:
 		return "failed_to_report"
 	case result == ErrActivityResultPending:
--- a/internal/activity_task_handler.go
+++ b/internal/activity_task_handler.go
@@ -39,21 +39,22 @@ import (
 type (
 	// activityTaskHandlerImpl is the implementation of ActivityTaskHandler
 	activityTaskHandlerImpl struct {
-		clock              clockwork.Clock
-		taskListName       string
-		identity           string
-		service            workflowserviceclient.Interface
-		metricsScope       *metrics.TaggedScope
-		logger             *zap.Logger
-		userContext        context.Context
-		registry           *registry
-		activityProvider   activityProvider
-		dataConverter      DataConverter
-		workerStopCh       <-chan struct{}
-		contextPropagators []ContextPropagator
-		tracer             opentracing.Tracer
-		featureFlags       FeatureFlags
-		activityTracker    debug.ActivityTracker
+		clock                  clockwork.Clock
+		taskListName           string
+		identity               string
+		service                workflowserviceclient.Interface
+		metricsScope           *metrics.TaggedScope
+		logger                 *zap.Logger
+		userContext            context.Context
+		registry               *registry
+		activityProvider       activityProvider
+		dataConverter          DataConverter
+		workerStopCh           <-chan struct{}
+		contextPropagators     []ContextPropagator
+		tracer                 opentracing.Tracer
+		featureFlags           FeatureFlags
+		activityTracker        debug.ActivityTracker
+		enableLifecycleLogging bool
 	}
 )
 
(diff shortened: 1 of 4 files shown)


c-warren (approved):

shijiesheng:
The new option actually duplicates part of a global developer option through `worker.EnableVerboseLogging()`. But the difference is that this PR tries to add a per-worker option only for activity. For decision tasks in the future, what is the plan here?

internal/worker.go
```diff
@@ -296,6 +296,12 @@ type (
 		// default: false
 		EnableLoggingInReplay bool
 
+		// Optional: Enable logging of activity lifecycle events.
+		// When enabled, the worker logs a message when each activity attempt starts and when it ends,
+		// including the outcome (succeeded / failed / canceled / timeout / pending / failed_to_report).
+		// default: false
+		EnableActivityLifecycleLogging bool
```
natemort: Let's make this option more generic actually so that in a subsequent PR we can reuse this setting to add logs to Workflows as well. I do like `Lifecycle` in the naming quite a bit.

Since all Activities exist within a Workflow I propose the name `EnableWorkflowLifecycleLogging`, let's see what @shijiesheng and @c-warren think.

internal/activity_task_handler.go
```diff
@@ -79,33 +80,50 @@ func newActivityTaskHandlerWithCustomProvider(
...
+		tracer:                 params.Tracer,
+		featureFlags:           params.FeatureFlags,
+		activityTracker:        params.WorkerStats.ActivityTracker,
+		enableLifecycleLogging: params.EnableActivityLifecycleLogging,
 	}
 }
 
 // Execute executes an implementation of the activity.
 func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivityTaskResponse) (result interface{}, err error) {
+	logger := ath.logger.With(
```
natemort, line 104: Standardizing these is a great improvement.

internal/activity_task_handler.go
```diff
@@ -79,33 +80,50 @@ func newActivityTaskHandlerWithCustomProvider(
...
 
+	if ath.enableLifecycleLogging {
+		logger.Info("Activity started.")
+		defer func() {
+			fields := []zap.Field{zap.String(tagActivityOutcome, activityOutcome(result, err))}
+			if err != nil {
+				fields = append(fields, zap.Error(err))
+			}
+			logger.Info("Activity completed.", fields...)
+		}()
```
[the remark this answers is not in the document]
natemort: This is true, maybe "activity returned" might be the more accurate log? Pending activities is pretty rare though, it might be simpler for most users to ignore this nuance and the advanced users would still be able to understand the message. I don't have a strong opinion either way.

natemort (commented):
Awesome work! We discussed this a bit offline, I think this feature makes sense separate from the verbose logging global variable and we may remove that other global variable in the future.

natemort (approved):

PR #1522, opened 2026-08-29
Fail unregistered activities immediately, not at schedule-to-close

### LINKED ISSUE #1492: activity worker not sending respond-activity failures
**Describe the bug**
Go SDK would time out an activity if the activity type is unknown (not registered) to the worker. Activity worker would silently fail the tasks and let the activity time out on the server. 

The behavior is different in Java SDK, which would respond failures immediately. 

Cause is due to:
```
func (atp *activityTaskPoller) ProcessTask(task interface{}) error {
	// Process the activity task.
	request, err := atp.taskHandler.Execute(atp.taskList.GetName(), activityTask.task)
	if err != nil {
		metricsScope.Counter(metrics.ActivityExecutionFailedCounter).Inc(1)
		return err // <- silent failure on SDK with loggings only
	}
```
https://github.com/cadence-workflow/cadence-go-client/blob/70ea3b3f8869a64a76f29e36ae605978733884a1/internal/internal_task_pollers.go#L1166-L1167

**To Reproduce**
Is the issue reproducible? 
- Yes

**Steps to reproduce the behavior:**
Start both Cadence activity worker and workflow worker. Start a simple registered workflow that executes one unregistered activity. Activity would hang in workflow history and eventually timeout. 

**Expected behavior**
Activity should fail instead with error message (activity not registered)

**Screenshots**
N/A

**Additional context**
N/A

1fanwang:
**What changed?**

An activity whose type is not registered on the polling worker now fails immediately. Before this change it stayed running until its schedule-to-close timeout expired, because the worker never told the server anything.

**Why?**

Fixes #1492.

Register `ActivityA` on a worker and have a workflow schedule `ActivityB` on the same task list. The worker logs `unable to find activityType=ActivityB` and sends nothing back, so the workflow blocks for the whole schedule-to-close timeout on a failure the worker already knew about at poll time. With a ten minute timeout that is ten minutes of nothing happening. The Java client responds right away here.

`Execute` returned a plain error, and `ProcessTask` reads a non-nil error as a client-side processing failure: it counts it and returns before the reporting path, so no RPC is made. The panic handler just below already returns a response with a nil error, and the unregistered case now matches it. `ActivityExecutionFailedCounter` moves into the handler so the metric survives the poller no longer seeing an error.

**How did you test it?**

Two new tests: one on the handler's request, one driving `ProcessTask` against a mock service to prove the RPC is sent. Both fail on master, the poller one on `missing call(s) to RespondActivityTaskFailed`, which is the bug itself. Full `go test ./internal/...` is green.

<details>
<summary>Red, with the handler restored from <code>upstream/master</code> (b9c7e58)</summary>

```
$ git checkout upstream/master -- internal/activity_task_handler.go
$ go test ./internal/ -run 'UnregisteredActivity' -count=1
--- FAIL: TestActivityTaskHandler_Execute_UnregisteredActivity (0.00s)
    activity_task_handler_test.go:245:
        	Error:      	Received unexpected error:
        	            	unable to find activityType=notRegistered. Supported types: [activityType]
--- FAIL: TestActivityTaskPoller_ProcessTask_UnregisteredActivity (0.00s)
    internal_task_pollers_test.go:388:
        	Error:      	Received unexpected error:
        	            	unable to find activityType=notRegistered. Supported types: [registered]
    controller.go:99: missing call(s) to *workflowservicetest.MockClient.RespondActivityTaskFailed(is anything, is anything, is anything)
    controller.go:99: aborting test due to missing call(s)
FAIL	go.uber.org/cadence/internal	0.432s
```
</details>

<details>
<summary>Green</summary>

```
$ go test ./internal/ -run 'UnregisteredActivity' -count=1 -v
=== RUN   TestActivityTaskHandler_Execute_UnregisteredActivity
    ERROR	Activity error.	{"ActivityType": "notRegistered", "error": "unable to find activityType=notRegistered. Supported types: [activityType]"}
--- PASS: TestActivityTaskHandler_Execute_UnregisteredActivity (0.00s)
=== RUN   TestActivityTaskPoller_ProcessTask_UnregisteredActivity
    ERROR	Activity error.	{"ActivityType": "notRegistered", "error": "unable to find activityType=notRegistered. Supported types: [registered]"}
--- PASS: TestActivityTaskPoller_ProcessTask_UnregisteredActivity (0.00s)
ok  	go.uber.org/cadence/internal	0.601s

$ go test ./internal/... -count=1
ok  	go.uber.org/cadence/internal	34.900s
ok  	go.uber.org/cadence/internal/batch	24.958s
ok  	go.uber.org/cadence/internal/common	0.639s
... (16 packages, all ok)
```
</details>

**Potential risks**

Low. The error text is unchanged. The behavior change is deliberate: an activity that used to hang until schedule-to-close now fails right away, as a `GenericError` carrying the same message. Retries follow the activity's retry policy, so a fleet mid-deploy with a partial registration retries rather than hangs.

### DIFF
--- a/internal/activity_task_handler.go
+++ b/internal/activity_task_handler.go
@@ -144,9 +144,13 @@ func (ath *activityTaskHandlerImpl) Execute(taskList string, t *s.PollForActivit
 
 	activityImplementation := ath.getActivity(activityType)
 	if activityImplementation == nil {
-		// Couldn't find the activity implementation.
+		// Report the failure to the server instead of returning an error, so the
+		// activity fails now rather than at its schedule-to-close timeout.
 		supported := strings.Join(ath.getRegisteredActivityNames(), ", ")
-		return nil, fmt.Errorf("unable to find activityType=%v. Supported types: [%v]", activityType, supported)
+		notRegistered := fmt.Errorf("unable to find activityType=%v. Supported types: [%v]", activityType, supported)
+		logger.Error("Activity error.", zap.Error(notRegistered))
+		metricsScope.Counter(metrics.ActivityExecutionFailedCounter).Inc(1)
+		return convertActivityResultToRespondRequest(ath.identity, t.TaskToken, nil, notRegistered, ath.dataConverter), nil
 	}
 
 	// panic handler
--- a/internal/activity_task_handler_test.go
+++ b/internal/activity_task_handler_test.go
@@ -235,6 +235,23 @@ func TestActivityTaskHandler_Execute_CustomError(t *testing.T) {
 	require.Equal(t, int32(30), failedRequest.GetFailureOptions().GetNextRetryIntervalSeconds())
 }
 
+func TestActivityTaskHandler_Execute_UnregisteredActivity(t *testing.T) {
+	activityHandler := newHandlerForTest(t, workerExecutionParameters{}, func(ctx context.Context) error { return nil })
+
+	pats := createBasicPollForActivityTaskResponse()
+	pats.ActivityType = &s.ActivityType{Name: common.StringPtr("notRegistered")}
+
+	res, err := activityHandler.Execute(tasklist, pats)
+	require.NoError(t, err)
+
+	failedRequest, ok := res.(*s.RespondActivityTaskFailedRequest)
+	require.True(t, ok, "response is not of type *s.RespondActivityTaskFailedRequest but of type %T", res)
+	require.Equal(t, pats.TaskToken, failedRequest.GetTaskToken())
+	require.Equal(t, errReasonGeneric, failedRequest.GetReason())
+	require.Contains(t, string(failedRequest.GetDetails()), "unable to find activityType=notRegistered")
+	require.Contains(t, string(failedRequest.GetDetails()), "activityType")
+}
+
 func newHandlerForTest(t *testing.T, params workerExecutionParameters, activityFunction any) ActivityTaskHandler {
 	registry := newRegistry()
 	err := registry.registerActivityFunction(activityFunction, RegisterActivityOptions{Name: "activityType"})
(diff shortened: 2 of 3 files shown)


zawadzkidiana (approved):
Nice fix!

A couple of things worth a quick look before merge, not blockers:
- `ActivityExecutionFailedCounter` now increments inside `Execute` for the unregistered case. Is there any other place upstream/downstream (e.g. in the poller) that also increments this counter when `Execute` returns a non-nil error? Since `Execute` no longer returns an error for this case, I'd expect that to be safe, but worth double-checking there's no double-count for other error paths that still return errors from `Execute`.
- The branch shows as "behind" master, might need a quick rebase before it's mergeable, unrelated to the code itself.

shijiesheng (approved):

dianazawadzki:
> Nice fix, and a genuinely useful one — a workflow blocking for the full schedule-to-close timeout on an error the worker already knew about at poll time is a rough failure mode to debug in production, especially with long timeouts.
> 
> The change itself is minimal and well-targeted: matching the panic handler's existing pattern (build a `RespondActivityTaskFailedRequest` via `convertActivityResultToRespondRequest` and return a nil error) rather than inventing a new path is the right call, and it keeps `ProcessTask`'s error handling simple. I also appreciate the PR description showing the failing-then-passing test output — that made this easy to verify without needing to pull the branch.
> 
> Test coverage is solid: one test on `Execute` asserting the request shape, one driving `ProcessTask` end-to-end against a mock to prove the RPC actually fires. That combination is exactly what an intermittent regression here would need to catch.
> 
> A couple of things worth a quick look before merge, not blockers:
> 
> * `ActivityExecutionFailedCounter` now increments inside `Execute` for the unregistered case. Is there any other place upstream/downstream (e.g. in the poller) that also increments this counter when `Execute` returns a non-nil error? Since `Execute` no longer returns an error for this case, I'd expect that to be safe, but worth double-checking there's no double-count for other error paths that still return errors from `Execute`.
> * The branch shows as "behind" master — might need a quick rebase before it's mergeable, unrelated to the code itself.
> 
> Thanks for including the repro and the clear before/after — makes this a fast review. 🙂
> 

Apologies for this team! Glad Shijie took a look, good change.

Issue #1394, opened 2024-11-07
Go client fails to compile on FreeBSD 14: missing function getSendQueueLen


moonetic:
**Describe the bug**
When trying to compile on FreeBSD 14, the compile fails with following key message: missing function getSendQueueLen.
This isuue regards tchannel-go repository.
An issue has been opened there but it hangs for a while:
https://github.com/uber/tchannel-go/issues/914

**To Reproduce**
Is the issue reproducible? 
- Yes

**Steps to reproduce the behavior:**
Try to compile a project with dependency to current repository on FreeBSD 14, if fails because of https://github.com/uber/tchannel-go issue.

**Expected behavior**
Compilation is successfull

moonetic:
@ibarrajo hey, I noticed that the PR on the issue just hangs there, how can we get a review on that?

ibarrajo:
I pinged a maintainer at Uber to review your PR =]

ibarrajo:
@moonetic it's been merged: https://github.com/uber/tchannel-go/pull/925

PR #1523, opened 2026-08-31
feat(workflow): WithCronSchedule for ContinueAsNew and child workflow

shijiesheng:
**What changed?**

By design, cron schedule is not inherited on explicit ContinueAsNew as well as child workflow. However, there are use cases that need it.

* add WithCronSchedule API
* the same field in the workflow options will be used by child workflow as well. 

**Why?**
Ref [#1394 ](https://github.com/cadence-workflow/cadence/issues/1394)

A workflow with cron schedule may receive a signal to change the input for the future runs or even change the cron schedule. This was not possible before because ContinueAsNew error doesn't propagate cron schedule at all. 

To be backward compatible, we need to ask users to explicitly use WithCronSchedule API to inject the cron schedule into the context so they can return a custom ContinueAsNewError to change the cron schedule or the input. 

**How did you test it?**

Unit Test

**Potential risks**

**Detailed Description**

**Impact Analysis**
- **Backward Compatibility**: Yes, new API that doesn't change existing behavior
- **Forward Compatibility**: Yes

**Testing Plan**
- **Unit Tests**: Yes
- **Persistence Tests**: No
- **Integration Tests**: No
- **Compatibility Tests**: No

**Rollout Plan**
- What is the rollout plan? New API usage
- Is it safe to rollback? Does the order of rollback matter? No, new API users will get errors

### DIFF
--- a/internal/workflow.go
+++ b/internal/workflow.go
@@ -1445,6 +1445,24 @@ func WithWorkflowTaskList(ctx Context, name string) Context {
 	return ctx1
 }
 
+// WithCronSchedule sets a cron schedule on the context.
+//
+// The same workflow options are used for ContinueAsNew (via NewContinueAsNewError)
+// and for child workflows (via ExecuteChildWorkflow). Cron is not copied from the
+// current run automatically; pass it explicitly when the next run or child should
+// keep a schedule.
+//
+// WithChildWorkflowOptions overwrites this field with ChildWorkflowOptions.CronSchedule.
+// Apply WithCronSchedule after WithChildWorkflowOptions if both are used.
+//
+// An empty schedule clears a previously set value. Invalid specs are rejected when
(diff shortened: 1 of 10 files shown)


internal/workflow.go
abhishekj720: Should WithChildWorkflowOptions skip the cronSchedule assignment when cwo.CronSchedule == "", and add a separate ClearCronSchedule if explicit removal is needed? 

If the caller doesn't set CronSchedule in ChildWorkflowOptions, the zero value ("") silently clears any cron previously set by WithCronSchedule.

shijiesheng: good catch. Fixing

abhishekj720 (approved):
LGTM. 
Nit: Maybe you should add "" test case.

PR #1524, opened 2026-09-01
fix: reschedule decision task with pending local activity on worker shutdown

abhishekj720:
**What changed?**
on worker shutdown, ProcessWorkflowTask now watches the local-activity tunnel stop channel in the decision wait loop and returns a decisionHeartbeatError, letting the poller reschedule the decision task on a healthy worker.

**Why?**
once the worker starts shutting down, the local-activity pool stops accepting tasks, so a pending local activity (and its retry) can no longer be dispatched and its result never arrives. The loop would otherwise keep decision-heartbeating until the server's decisionHeartbeatTimeout (default 30m), stalling work such as schedule-triggered workflow starts.

**How did you test it?**
Added TestLocalActivity_WorkerShutdownAbandonsDecisionTask: dispatches a pending local activity, closes the worker stop channel, and asserts ProcessWorkflowTask returns a decisionHeartbeatError promptly. Passes under -race; full TaskHandlersTestSuite green; gofmt clean.

**Potential risks**
on graceful shutdown an in-flight local activity is abandoned and re-run after reschedule; safe for idempotent activities (LA results are markers, scheduled starts dedupe via deterministic RequestID). No normal-path impact: the new branch only fires on shutdown, and laTunnel is nil during replay so a nil channel never selects.

### DIFF
--- a/internal/internal_task_handlers.go
+++ b/internal/internal_task_handlers.go
@@ -816,6 +816,18 @@ func (wth *workflowTaskHandlerImpl) ProcessWorkflowTask(
 		workflowContext.Unlock(errRet)
 	}()
 
+	// laStopCh is closed when the worker starts shutting down. At that point the
+	// local activity pool stops accepting tasks, so a pending local activity (and
+	// any retry) can no longer be dispatched and its result will never arrive.
+	// Without this signal the loop below would keep heartbeating the decision task
+	// until the server's decisionHeartbeatTimeout (default 30m). laTunnel is nil
+	// during history replay (no worker); a nil channel never selects, leaving the
+	// replay path unchanged.
+	var laStopCh <-chan struct{}
+	if workflowContext.laTunnel != nil {
+		laStopCh = workflowContext.laTunnel.stopCh
+	}
+
 	var response interface{}
 process_Workflow_Loop:
 	for {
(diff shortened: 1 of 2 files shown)


internal/internal_task_handlers.go
```diff
@@ -827,6 +839,13 @@ process_Workflow_Loop:
 				deadlineToTrigger := time.Duration(float32(ratioToForceCompleteDecisionTaskComplete) * float32(workflowContext.GetDecisionTimeout()))
 				delayDuration := startTime.Add(deadlineToTrigger).Sub(time.Now())
 				select {
```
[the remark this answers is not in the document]
abhishekj720: Adding a non-blocking laResultCh pre-check would just duplicate the result-handling for a rare, benign case, so I'd rather keep the branch minimal.

internal/internal_task_handlers.go
```diff
@@ -827,6 +839,13 @@ process_Workflow_Loop:
...
+				case <-laStopCh:
+					// Worker is shutting down and the pending local activity can no
+					// longer be dispatched. Abandon this decision task so the server
+					// reschedules it on a healthy worker, instead of holding it open
+					// until decisionHeartbeatTimeout.
+					return nil, &decisionHeartbeatError{Message: "worker is shutting down with a pending local activity; abandoning decision task for reschedule"}
```
natemort, line 847: Just to confirm, does returning this result in the decision task timing out? Or does the client return a failed decision task?

abhishekj720: It times out. Returning a decisionHeartbeatError hits the existing branch in the poller that returns the error without calling RespondDecisionTaskCompleted or RespondDecisionTaskFailed, so the worker stops responding and the server times the decision task out and reschedules it on another worker.

abhishekj720: It's the same abandon and reschedule path the real heartbeat send failure case already relies on.

shijiesheng (approved):
Could you add an integration test that can guard against the heartbeat goroutine is no longer running after shutdown? We need a worker with local activities that would always fail with some sleep.

PR #1530, opened 2026-09-22
feat(schedule): expose missedRuns, skippedRuns, bufferedFireCount, runningWorkflowCount in ScheduleInfo

### LINKED ISSUE #1526: ScheduleInfo missing missedRuns, skippedRuns, bufferedFireCount, runningWorkflowCount fields
The IDL (both thrift and proto) exposes four schedule stats fields that the server populates, but the Go SDK silently drops them, they're never added to the `ScheduleInfo` struct and the thrift converter ignores them.

**Missing fields:**
- `missedRuns` — runs that were skipped because they fell outside the catch-up window
- `skippedRuns` — runs skipped due to overlap policy
- `bufferedFireCount` — runs currently waiting in the buffer
- `runningWorkflowCount` — workflows actively running under this schedule

**IDL references:**
- Thrift: `shared.thrift` fields 70–100 on `ScheduleInfo`
- Proto: `schedule.proto` fields 7–10 on `ScheduleInfo`
- The compatibility converters (`thrift/schedule.go`, `proto/schedule.go`) correctly handle them; the gap is only in the top-level Go SDK type

**Impact:** Makes it impossible to diagnose why a schedule underperforms (missed runs), why runs are being dropped (overlap), or how many are queued/running without dropping down below the SDK.

**Fix:** Add the four fields to `ScheduleInfo` in `internal/schedule.go` and populate them in `scheduleInfoFromThrift` / `scheduleInfoFromProto`.

abhishekj720:
## Summary

Fixes #1526

Four fields exist in both the thrift and proto IDLs (`shared.thrift` fields 70–100, `schedule.proto` fields 7–10) and the compatibility layer already handles them correctly but `ScheduleInfo` in the Go SDK had no corresponding fields, and the thrift converter silently dropped them.

- Add `MissedRuns`, `SkippedRuns`, `BufferedFireCount`, `RunningWorkflowCount` to `ScheduleInfo`
- Populate them in `scheduleInfoFromThrift`
- Add assertions to `TestScheduleInfoFromThrift`

## Test plan

- [x] `TestScheduleInfoFromThrift` updated and passing
- [x] All schedule tests passing (`go test ./internal -run Schedule`)


internal/internal_schedule_convert_thrift_test.go
```diff
@@ -251,19 +251,31 @@ func TestScheduleInfoFromThrift(t *testing.T) {
 	startNs := int64(1_700_000_000_000_000_000)
 	endNs := int64(1_800_000_000_000_000_000)
 	total := int64(42)
+	missed := int64(3)
```
c-warren, line 254: You should add some fuzz tests, that will prevent things like this :)

c-warren: https://github.com/cadence-workflow/cadence-go-client/blob/0f1e03eb5f0d8aaab3f3b99870a37d391e2fe978/internal/compatibility/api_test.go#L3471

abhishekj720: Added

internal/schedule.go
```diff
@@ -187,6 +187,14 @@ type ScheduleInfo struct {
 	LastUpdateTime time.Time
 	// OngoingBackfills is not currently populated by the server; it will always be nil.
 	OngoingBackfills []*BackfillInfo
+	// MissedRuns is the number of runs that were skipped because they fell outside the catch-up window.
```
YaweiZhang-930: nit: I checked the server, missedRuns only gets bumped in two places(https://github.com/cadence-workflow/cadence/blob/410e0d157582cdc3407dba9b3689b7dc0b692106/service/worker/scheduler/workflow.go#L579 and https://github.com/cadence-workflow/cadence/blob/410e0d157582cdc3407dba9b3689b7dc0b692106/service/worker/scheduler/workflow.go#L601). neither of them is the catchup window. 

something like this eems more accurate: **MissedRuns is the number of scheduled fires the server failed to start**

abhishekj720: Ahh, you are right.

abhishekj720: changed the comment

internal/internal_schedule_convert_thrift_test.go
```diff
@@ -251,23 +251,70 @@ func TestScheduleInfoFromThrift(t *testing.T) {
...
+	assert.Equal(t, int64(2), got.BufferedFireCount)
+	assert.Equal(t, int64(1), got.RunningWorkflowCount)
 	require.Len(t, got.OngoingBackfills, 1)
 	assert.Equal(t, "bf-1", got.OngoingBackfills[0].BackfillID)
 }
 
+// FuzzScheduleInfoFromThrift verifies that scheduleInfoFromThrift never panics and that
+// the four stat counter fields (MissedRuns, SkippedRuns, BufferedFireCount,
+// RunningWorkflowCount) pass through as-is for any int64 value.
+func FuzzScheduleInfoFromThrift(f *testing.F) {
```
YaweiZhang-930, line 286: nit: The fuzz target only asserts the four fields you just added. If someone adds a fifth field to ScheduleInfo tomorrow and forgets to map it in scheduleInfoFromThrift, this test still passes. I guess random-filling the thrift struct and checking no field on the result came back zero will catch that.

YaweiZhang-930 (approved):
LGTM! One more comment regarding the fuzz test.

internal/internal_schedule_convert_thrift_test.go
abhishekj720: Updated the fuzz test to fill all scalar fields and assert each one round-trips exactly.

PR #1532, opened 2026-09-22
fix(schedule): remove stale always-zero comments on ScheduleInfo timestamps

### LINKED ISSUE #1528: ScheduleInfo.CreateTime, LastUpdateTime and PauseInfo.PausedAt are always zero: server not populating them
Three timestamp fields on schedule describe responses are always zero because the Cadence server does not populate them in responses, even though the IDL wire fields exist.

- `ScheduleInfo.CreateTime` 
- `ScheduleInfo.LastUpdateTime`
- `SchedulePauseInfo.PausedAt`

The Go SDK currently has these fields with comments saying "not currently populated by the server; will always be zero." The client converter code correctly reads them — the gap is entirely on the server side.

**Impact:** 
- Cannot audit schedule lifecycle or detect stale configs
- SLA alerting on "how long has this schedule been paused" always returns zero
- Operational tooling that relies on these timestamps gives incorrect data

**What needs to happen:** The Cadence server needs to persist and return these timestamps. This is tracked here so the Go client can be updated (remove the "always zero" comments and add tests) once the server fix lands.

Temporal handles this by persisting schedule creation/update timestamps as part of the schedule entity in the persistence layer at write time, not lazily on read. Same approach would work here.

abhishekj720:
Closes #1528

## What

The `ScheduleInfo.CreateTime`, `ScheduleInfo.LastUpdateTime`, and `SchedulePauseInfo.PausedAt` fields had doc comments claiming they are always zero / never populated by the server. Those comments were inaccurate.

The Cadence server has set these fields:`CreateTime` and `LastUpdateTime` are written in `initStateOnFirstExecution` / `handlePause`/`handleUnpause`, serialized through `buildScheduleDescription`, and mapped through both the frontend handler and the thrift/proto wire serializers. The only case where these are zero is for schedules created before that server code was deployed.

## Changes

- Remove the three stale "always zero" comments from `internal/schedule.go`

## Testing

```
go test go.uber.org/cadence/internal -run "TestScheduleInfoFromThrift|TestSchedulePauseInfoFromThrift" -v
```
Both pass. `make tidy && make build` clean.


YaweiZhang-930 (approved):
LGTM!

